PR #201, opened 2026-06-11
fix: parse good_till_date and emit correct GTD 126/432 expire encoding (ibx#199)

### LINKED ISSUE #199: TIF=GTD orders always rejected: good_till_date dropped by Order::attrs() + wrong 126/432 ExpireTime/ExpireDate wire encoding
**Environment:** v0.5.1 and current `main` (c6ab5a1), Python bindings, IB paper account, live-session testing.

Every order with `tif = "GTD"` is rejected by the gateway and comes back `Inactive`. Engine log:

```
WARN ibx::engine::hot_loop::ccp] ExecReport REJECTED: clord=... reason='Message must contain field # 432' 103=0
```

**Repro (Python):**

```python
o = Order()
o.order_id = oid; o.action = "BUY"; o.total_quantity = 1.0
o.order_type = "LMT"; o.lmt_price = 1.00
o.tif = "GTD"; o.good_till_date = "20260611-18:00:00"
client.place_order(oid, aapl_stk, o)
# -> order_status 'Inactive', ExecReport rejection above
```

**Two independent bugs:**

1. `Order::attrs()` (src/api/types.rs) hardcodes `good_after: 0, good_till: 0` — the `good_till_date` / `good_after_time` strings are silently dropped, so the engine emits TIF=6 with no expiry fields at all.

2. Even with `good_till` populated, the wire encoding in `SubmitLimitEx` (src/engine/hot_loop/order_builder.rs) doesn't match what the gateway accepts. Behavior verified empirically against the paper gateway, one variant at a time:

| emitted | gateway response |
|---|---|
| 126 = `YYYYMMDD HH:MM:SS` (space, current `unix_to_ib_datetime` format) | `Invalid value in field # 126`, then treated as absent → `Message must contain field # 432` |
| 432 = `YYYYMMDD` + 126 = `YYYYMMDD-HH:MM:SS` | `Invalid value in field # 432` |
| 432 = `YYYYMMDD` alone | **accepted** (date-only GTD) |
| 126 = `YYYYMMDD-HH:MM:SS` (dash) alone | **accepted** (time-precise GTD) |

So the dialect is: time-precise GTD → tag 126 alone, FIX UTCTimestamp with dash separator; date-only GTD → tag 432 alone; never both.

**Validation:** with the fix, a GTD LMT acks (`PreSubmitted`), rests, and the gateway broker-side cancels it at the requested timestamp (verified with a now+75s expiry — `Cancelled` arrives on schedule without any client-side cancel). PR incoming with the fix + round-trip unit tests for the new datetime parser.

positronic-penguin:
Fixes #199.

Two changes, matching the analysis in the issue:

**1. `Order::attrs()` now parses `good_till_date` / `good_after_time`** (src/api/types.rs) instead of hardcoding 0. New `config::ib_datetime_to_unix()` is the inverse of the existing `unix_to_ib_datetime()`: accepts `YYYYMMDD HH:MM:SS`, `YYYYMMDD-HH:MM:SS`, or bare `YYYYMMDD` (treated as UTC; anything after the seconds field, e.g. a tz suffix, is ignored — documented). Includes `ymd_to_days()` (Hinnant days_from_civil, inverse of the existing `days_to_ymd`) and round-trip + garbage-rejection unit tests.

**2. GTD wire encoding** (src/engine/hot_loop/order_builder.rs), per the empirically verified gateway dialect:
- time-precise GTD → tag **126 alone**, `YYYYMMDD-HH:MM:SS` (dash — the space-separated form is rejected as `Invalid value in field # 126`)
- date-only GTD (midnight) → tag **432 alone**, `YYYYMMDD`
- never both (432+126 together → `Invalid value in field # 432`)
- 168 (EffectiveTime / `good_after_time`) switched to the same dash format for consistency

**Validation on an IB paper account:** GTD LMT now acks (`PreSubmitted`), rests, and the gateway broker-side cancels at the requested timestamp — verified with a now+75s expiry, `Cancelled` arriving on schedule with no client-side cancel. DAY/GTC/IOC paths unchanged. `cargo test --lib config` green on this branch.

Co-authored with Claude (Anthropic).

### DIFF
--- a/src/api/types.rs
+++ b/src/api/types.rs
@@ -413,8 +413,8 @@ impl Order {
             min_qty: self.min_qty.max(0) as u32,
             hidden: self.hidden,
             outside_rth: self.outside_rth,
-            good_after: 0,
-            good_till: 0,
+            good_after: crate::config::ib_datetime_to_unix(&self.good_after_time),
+            good_till: crate::config::ib_datetime_to_unix(&self.good_till_date),
             oca_group: self.oca_group.parse().unwrap_or(0),
             oca_group_str: if self.oca_group.parse::<u64>().is_err() && !self.oca_group.is_empty() {
                 self.oca_group.clone()
--- a/src/config.rs
+++ b/src/config.rs
@@ -108,6 +108,47 @@ pub fn unix_to_ib_datetime(secs: i64) -> String {
     )
 }
 
+/// Convert (year, month, day) to days since Unix epoch. Inverse of `days_to_ymd`.
+pub fn ymd_to_days(y: u64, m: u64, d: u64) -> u64 {
+    // Algorithm from Howard Hinnant (days_from_civil)
+    let y = if m <= 2 { y - 1 } else { y };
+    let era = y / 400;
+    let yoe = y - era * 400;
+    let mp = if m > 2 { m - 3 } else { m + 9 };
+    let doy = (153 * mp + 2) / 5 + d - 1;
+    let doe = yoe * 365 + yoe / 4 - yoe / 100 + doy;
+    era * 146097 + doe - 719468
+}
+
+/// Parse an ibapi-style datetime string to unix seconds, treated as UTC.
+/// Accepts "YYYYMMDD HH:MM:SS", "YYYYMMDD-HH:MM:SS", or bare "YYYYMMDD".
+/// Anything after the seconds field (e.g. a timezone suffix) is ignored.
+/// Returns 0 for empty or unparseable input.
+pub fn ib_datetime_to_unix(s: &str) -> i64 {
+    let s = s.trim();
+    if s.len() < 8 || !s.is_char_boundary(8) {
+        return 0;
+    }
+    let (date, rest) = s.split_at(8);
+    let y = match date[0..4].parse::<u64>() { Ok(v) => v, Err(_) => return 0 };
+    let m = match date[4..6].parse::<u64>() { Ok(v) => v, Err(_) => return 0 };
+    let d = match date[6..8].parse::<u64>() { Ok(v) => v, Err(_) => return 0 };
+    if y < 1970 || !(1..=12).contains(&m) || !(1..=31).contains(&d) {
+        return 0;
+    }
+    let mut secs = (ymd_to_days(y, m, d) * 86400) as i64;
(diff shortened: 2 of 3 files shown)


deepentropy (commented):
Reviewed against current `main` (v0.6.1). Thank you for this — the analysis was correct, and the core of it independently matches what shipped: the same date-only-vs-time-precise expiry split, the same mutual-exclusion rule, and the same dash-form timestamp all landed on main through the v0.6.x order-path work. So most of this PR is now already in place, by a slightly different route, and it will not merge as-is. Details:

**Already on main (please don't re-merge):**
- `good_till_date` parsing in `Order::attrs()` — main parses it into an explicit date-only vs instant result rather than inferring date-only from a midnight check, so the `% 86400` heuristic here isn't needed.
- The expiry encoding split (date-only path vs time-precise path, mutually exclusive, dash-form timestamp) is emitted by the shared order encoder.

**Won't merge cleanly:** the v0.6.0 order-path rework (#224) replaced the inline extended-limit block this PR edits with a single shared encoder (`send_order_ex`) that every order type routes through. The hunk in `order_builder.rs` targets code that no longer exists, so this needs a rebase onto current main — and once rebased, that hunk becomes a no-op since the behavior is already there.

**The one genuinely net-new piece:** `good_after_time`. Main still leaves it unset on purpose — I held it back pending a wire capture, since its outbound format wasn't verified against the gateway (the code comment references ibx#199). Your PR body validates the GTD expiry path end to end, but I don't see the same explicit paper validation for the good-after-time path specifically. If you've confirmed a delayed-activation order acks and activates at the requested time on paper, that's the salvageable contribution here — a small rebased PR that only wires up `good_after_time` (plus the `config` parse helper it needs) would be reviewable on its own. If not yet verified, ib-agent is the place to capture it first.

**config.rs helpers:** `ib_datetime_to_unix` / `ymd_to_days` are clean and the round-trip tests are nice, but main already has the parse direction it needs via `parse_ib_expiry`, so only the good-after-time path would actually consume the new helper.

Net: the GTD fix you set out to make is already resolved on main. If you want to carry this forward, the useful follow-up is a rebased, good-after-time-only change with a paper validation of the activation timing. Happy to review that.

Issue #111, opened 2026-03-23
Live account login fails when IB Key (mobile 2FA) is enabled


nbopardi:
## Summary

Accounts with IB Key (mobile 2FA) enabled cannot connect. After SRP authentication completes, the IB server enters a second-factor challenge loop sending type-530 keepalive messages while waiting for IB Key approval on the user's mobile device (notification is never received). ibx has no mechanism to participate in this challenge, so the post-auth phase times out with `"Never received data start after auth"` (`gateway.rs:592`).

## Current State

| Aspect | Status |
|--------|--------|
| SRP auth (username + password) | **Works** — `do_srp` completes, session token obtained |
| Soft token challenge (farm-level `35=S`) | **Works** — `do_soft_token` responds to farm auth challenges (`gateway.rs:188`) |
| IB Key 2FA (login-level) | **Not supported** — server sends NS keepalives (type 530), ibx loops until timeout |
| Paper accounts (no 2FA) | **Works** — full connection including all farms |

## What Happens

1. SRP authentication succeeds, `NS_CONNECT_RESPONSE` (type 523) is received
2. Port type change is sent, server acknowledges
3. Instead of `NS_FIX_START` (type 525), server sends repeated type-530 messages — these are keepalives while IB waits for the user to approve the login on their mobile IB Key app (mobile notification never received)
4. The post-auth loop in `gateway.rs:540–594` doesn't recognize these as a 2FA challenge and logs them as `"Post-auth msg type=530"`
5. After the loop exhausts its iterations, it returns `"Never received data start after auth"`

## What Was Tried

- **Adding `FLAG_SOFT_TOKEN` to connection flags** — server rejects; `SOFT_TOKEN` (`xyz.rs:14`) is for farm-level XYZ auth challenges, not the initial login 2FA gate
- **Proactively sending a `SOFT_TOKEN` init** — server rejects this as well

## What's Needed

IB Key 2FA is a server-side push notification flow. The server waits for the user to tap "approve" in the IB Key mobile app. ibx needs to:

1. **Detect** the 2FA-pending state (type-530 keepalives after `NS_CONNECT_RESPONSE` but before `NS_FIX_START`)
2. **Active** the push notification to the user
3. **Wait** with periodic keepalive responses until the user approves (or a configurable timeout)
4. **Resume** normal post-auth flow once `NS_FIX_START` arrives

deepentropy:
[...]

PR #245, opened 2026-07-28
2FA gate: run the IBKey code provider off the socket thread (ibx#244)

### LINKED ISSUE #176: Implement IBKey Challenge/Response 2FA via code_provider callback (semi-headless path for #174)
## Goal

Add an IBKey **Challenge/Response** 2FA path to ibx so unattended scripts can authenticate without tapping the IBKey app's "Approve" button — by supplying the 8-character response code via a callback instead.

This is the actionable subset of #174. Fully headless TOTP is not achievable (see "Why not Base32 TOTP" below); supplying the code programmatically is.

## Background

The seamless-push flow (#111) is already implemented: client sends `XYZ 775 SWCR_TOKEN state=1`, server replies with `state=2`, the user taps Approve on the phone, server completes with `AUTH_FINISH(771) state=5 PASSED`.

ib-agent#149 has now mapped the **Challenge/Response variant** — the alternative method offered in the same IBKey dialog where the user reads an 8-character code off the phone and types it in. The wire reuses the same msg id (775), with one additional client→server message:

```
CLIENT -> XYZ 775 state=1                        (challenge request, same as push)
SERVER -> XYZ 775 state=2                        (challenge + AVTH_URL, same as push)
                ─── user reads 8-char code from IBKey ───
CLIENT -> XYZ 775 state=3  code="02226534"       (NEW — code submission)
SERVER -> XYZ 775 state=4  result="PASSED"
SERVER -> XYZ 771 AUTH_FINISH state=3 PASSED
```

`state=3` payload — 8-char ASCII code framed at offset 32, length-prefixed, no transformation:

```
00 00 00 14    payload_len = 20
00 00 03 07    msg_type    = 775
00 00 00 01    sub_type    = 1
00 00 00 03    state       = 3
00 00 00 00    username    (empty in submission)
00 00 00 00
00 00 00 08    code length = 8
30 32 32 32 36 35 33 34   ← "02226534" (literal ASCII)
00 00 00 00
```

On `state=4 result="FAILED"`: server skips `AUTH_FINISH` entirely; gateway tears down the socket. **No retry loop exists** — one wrong code is terminal.

Full capture in ib-agent#149.

## Proposed API

Add a `code_provider` field to `GatewayConfig` alongside the existing IBKey settings:

```rust
pub struct GatewayConfig {
    // ... existing ...
    pub ib_key_timeout_secs: u64,
    pub ib_key_token_sub_type: String,

    /// If set, ibx will use the Challenge/Response 2FA path instead of waiting
    /// for a mobile push approval. Called once after the server delivers the
    /// challenge; the returned 8-character code is submitted as XYZ 775 state=3.
    /// The callback receives the AVTH_URL and 6-digit display id from state=2
    /// so the caller can correlate with the IBKey app.
    pub code_provider: Option<Arc<dyn Fn(IbKeyChallenge) -> Result<String, IbKeyError> + Send + Sync>>,
}

pub struct IbKeyChallenge {
    pub display_id: String,    // e.g. "399 830"
    pub avth_url: String,      // https://www.clientam.com/ibkr/ibkey/seamless?S=...
}
```

If `code_provider` is `None`, behavior is unchanged (push flow as today).

## Implementation plan

1. **Extend `XyzMsg::SwcrToken`** decode/encode in `src/protocol/xyz.rs` to round-trip `state=3` with the 8-char code field.
2. **Branch in `do_ib_key_2fa`** (`src/auth/session.rs:609`): after receiving `state=2`, if `config.code_provider.is_some()`, call it, send `state=3`, wait for `state=4`. If `None`, keep current behavior (wait for server-pushed `AUTH_FINISH state=5`).
3. **Map `state=4 result="FAILED"`** to a typed error (`IbKeyError::CodeRejected`) and surface immediately — no retry.
4. **Keep heartbeats running** during the wait (`NS_TEST_REQ(530)` / `NS_HEART_BEAT(531)` every 20 s) — should already be wired but verify against the capture.

## Test plan

Unit:
- Round-trip a `state=3` frame against the byte sequence from ib-agent#149.
- `state=4 FAILED` → `IbKeyError::CodeRejected`.

Integration (live account, manual):
- `examples/ex174_live_2fa_code_provider.rs` — `code_provider` reads the code from stdin; full login succeeds.
- Negative case — wrong code returns `CodeRejected` and the socket closes cleanly.

## Why not Base32 TOTP

The 8-char code IBKey generates is an HMAC over a 20-byte server challenge (from `state=2`) using a seed that lives in the phone's secure enclave. It is **not** RFC 6238 TOTP — there is no time component and no Base32-importable secret. Replacing the callback with a local computation would require extracting the algorithm and seed from the IBKey APK, which is out of scope here.

## Out of scope

- Extracting the IBKey HMAC algorithm
- Legacy SMS / Voice / classic Security Code Card (XYZ 774, class `N` — wire predicted from static analysis only, not captured live)
- Any change to the seamless-push path (#111)

## Related

- #174 — original headless-TOTP request
- #111 — IBKey seamless-push 2FA (already implemented & verified live)
- ib-agent#149 — wire capture this is built from

### LINKED ISSUE #244: 2FA gate: code_provider is called on the socket thread, so keepalive probes go unanswered while the operator enters the code
## Summary

The Challenge/Response path added in #180 calls `code_provider` inline inside `do_ib_key_2fa`'s receive loop (`src/auth/session.rs:759`, v0.7.1). That callback is where a human reads an 8-character code off the IBKey app, so it holds the thread for as long as the operator takes. For that entire window the gate is not reading the socket, and the `NS_TEST_REQUEST` reply that the same loop owes the server (`src/auth/session.rs:820`) never goes out.

The push flow does not have this problem. It sits in `recv_msg` for the whole 18-minute approval window, so probes keep getting answered; only the C/R branch parks the thread that owes the heartbeat. Same shape as #135, one layer up.

What is provably wrong client-side: the gate stops answering probes it otherwise answers, for an operator-controlled duration. What follows from that (the server dropping the connection before the code is submitted) is the expected consequence rather than something confirmed on a live run, since the C/R path has no live validation yet.

This is on the path the #176 test plan uses: the stdin prompt in `examples/ex176_live_2fa_code_provider.rs` blocks exactly this way. #208 would carry the same callback into the Python surface.

## Root cause

`src/auth/session.rs:753-769` (v0.7.1), inside the `state == 2` arm of the receive loop:

```rust
if !code_submitted {
    if let Some(provider) = code_provider {
        let challenge_info = IbKeyChallenge { .. };
        let code = provider(challenge_info)?;      // blocks the receive loop
        let submission = xyz::xyz_build_swcr_token_code_submission(&code);
        stream.write_all(&xyz::xyz_wrap(&submission))?;
```

The heartbeat reply is another arm of the same `match`, so no probe is answered until the callback returns.

Two side effects of the same structure:

- The client deadline at `src/auth/session.rs:703` cannot fire during the callback either, so `ib_key_timeout_secs` does not bound an interactive provider.
- A provider that never returns pins the login open with no way out.

## Proposed fix

Run the provider off the socket thread, keep the loop serving probes, and submit `XYZ 775` state=3 once the callback yields. No wire change: the state=3 builder and the state=4 PASSED/FAILED handling stay as they shipped in #180.

Happy to open a PR with this plus a regression test that fails when the provider is called inline.

userFRM:
## Summary

- `do_ib_key_2fa` called `code_provider` inline in the receive loop, so nothing answered `NS_TEST_REQUEST` while the operator read the code off the IBKey app (`src/auth/session.rs:759`). The provider now runs on its own thread and the loop keeps replying to probes for the whole entry window.
- `XYZ 775` state=3 is submitted on the first inbound message after the callback returns, so it trails entry by at most one probe interval for as long as the server keeps probing.
- No wire change. The state=3 builder and the state=4 PASSED/FAILED handling are exactly as they shipped in #180, and the byte-canonical assertion against the run-A capture still holds.
- The push flow (`code_provider = None`, #111) is untouched.
- Removes `IB_KEY_HEARTBEAT_CADENCE_SECS` and the `let _ = ...` placeholder that kept it referenced.
- Closes #244. Does not close #176: live verification of the C/R path is still outstanding, and this PR does not change that.

Two behaviours the fix does not engineer around, now documented on `CodeProvider` instead: submission is gated on inbound traffic, so a server that goes silent after state=2 never receives the code; and the provider thread is not cancelled if the gate returns early, so a stdin provider holds stdin until it returns on its own.

## Test plan

- [x] `cargo test --lib auth::` — 97 pass
- [x] `ib_key_2fa_cr_submits_code_then_passes_on_auth_finish_state_3` drives a scripted gateway that keeps issuing keepalive probes until the client submits state=3, then serves state=4 PASSED and `AUTH_FINISH(771)` state=3 PASSED. It asserts the byte-canonical run-A frame and that a heartbeat reply precedes it. Restoring the inline call fails it on the heartbeat assertion, so the test discriminates rather than passing by construction.
- [x] `ib_key_2fa_cr_code_rejected_on_state_4_failed` drives the same gateway, which withholds FAILED until the exact framed state=3 bytes appear. Previously the FAILED frame always arrived before the code was submitted, so that test only re-proved the no-provider path.
- [ ] Live verification with a human reading the code off the IBKey app (unchanged from #176)

Unrelated to this branch, present on current `main` and left alone: `config::expiry_tests::instant_round_trips_to_wire` and `named_zone_converts_with_dst` fail, and `tests/ib_paper_compat` does not compile (missing `filters` on `ControlCommand`, `trail_stop_price` and the adjustable-trailing fields on `OrderRequest`). Happy to file those separately if they are not already known.

### DIFF
--- a/src/auth/session.rs
+++ b/src/auth/session.rs
@@ -1983,9 +1983,13 @@ mod tests {
             "no code may reach the wire once the login is lost",
         );
 
-        // Polled path: the provider outlasts the inline grace period, so the
-        // code arrives on a later loop. The guard there needs its own case —
-        // the fast-path test above never reaches it.
+        // A provider that outlasts the inline grace period. This pins that no
+        // code reaches the wire by the polled route once the deadline has
+        // passed — but it does not exercise the guard on that route: the loop
+        // aborts on the deadline before the poll ever returns the code, so
+        // deleting that guard leaves this passing. Reaching it needs the
+        // deadline to expire between the top-of-loop check and the poll within
+        // one iteration, which is a race rather than a sequence.
         let mut stream = ProbingGateway::new(
             frame_xyz(&challenge), submission.clone(), Vec::new(),
         );
--- a/src/auth/session.rs
+++ b/src/auth/session.rs
@@ -645,6 +645,32 @@ pub type CodeProvider = std::sync::Arc<
     dyn Fn(IbKeyChallenge) -> io::Result<String> + Send + Sync,
 >;
 
+/// How long the gate waits inline for a freshly started `code_provider` before
+/// falling back to polling it between inbound messages. Sized to cover a
+/// provider that reads from a vault, an env var or a file, and to stay far
+/// below the server's probe cadence so nothing is starved by the wait.
+const IB_KEY_PROVIDER_FAST_PATH_GRACE: std::time::Duration =
+    std::time::Duration::from_secs(1);
+
+/// Put the 8-character Challenge/Response code on the wire as `XYZ 775` state=3.
+fn submit_swcr_code<S: Write>(stream: &mut S, code: &str) -> io::Result<()> {
+    let framed = xyz::xyz_wrap(&xyz::xyz_build_swcr_token_code_submission(code));
+    stream.write_all(&framed)?;
+    log::info!(
+        "2FA gate: submitted SWCR_TOKEN state=3 code (len={}, {} bytes framed)",
+        code.len(), framed.len(),
+    );
+    Ok(())
+}
+
+/// The provider thread dropped its sender without sending: it panicked.
+fn provider_panicked() -> io::Error {
+    ib_key_err(
+        io::ErrorKind::Other,
+        "2FA gate: code_provider panicked without returning a code",
+    )
+}
+
 /// Compact hex dump for diagnostic logging.
 fn hex_dump(bytes: &[u8]) -> String {
     bytes.iter().map(|b| format!("{:02x}", b)).collect::<Vec<_>>().join(" ")
@@ -769,10 +795,29 @@ pub fn do_ib_key_2fa<S: Read + Write>(
                             avth_url: approval_url.clone(),
                         };
                         let (tx, rx) = std::sync::mpsc::channel();
-                        std::thread::spawn(move || {
-                            let _ = tx.send(provider(challenge_info));
-                        });
-                        pending_code = Some(rx);
+                        std::thread::Builder::new()
+                            .name("ibkey-code-provider".into())
+                            .spawn(move || {
+                                let _ = tx.send(provider(challenge_info));
+                            })?;
+                        // An unattended provider (vault, env, file) resolves at
+                        // once and shouldn't have to wait for the next probe to
+                        // get its code on the wire, so give it a brief inline
+                        // grace period first. Far below the probe cadence, so
+                        // nothing is starved; a provider waiting on a human
+                        // falls through to the probe-driven path below.
+                        match rx.recv_timeout(IB_KEY_PROVIDER_FAST_PATH_GRACE) {
+                            Ok(result) => {
+                                submit_swcr_code(stream, &result?)?;
+                                code_submitted = true;
+                            }
+                            Err(std::sync::mpsc::RecvTimeoutError::Timeout) => {
+                                pending_code = Some(rx);
+                            }
+                            Err(std::sync::mpsc::RecvTimeoutError::Disconnected) => {
+                                return Err(provider_panicked());
+                            }
+                        }
                     }
                 }
             }
@@ -846,31 +891,25 @@ pub fn do_ib_key_2fa<S: Read + Write>(
             }
         }
 
-        // Submit the C/R code once the provider thread yields it. Checked after
-        // every inbound message, so submission trails the operator's entry by
-        // up to one server probe interval (~20 s). Polling on inbound traffic
-        // keeps the socket single-threaded; if that delay ever matters, give
-        // the stream a read timeout and select on both instead.
+        // A provider that outlived the grace period yields here instead. Polling
+        // on inbound traffic keeps the socket single-threaded, at the cost of
+        // trailing the operator's entry by up to one probe interval (~20 s); if
+        // that ever matters, give the stream a read timeout and select on both.
         if let Some(rx) = pending_code.as_ref() {
             match rx.try_recv() {
                 Ok(result) => {
                     pending_code = None;
-                    let code = result?;
-                    let submission = xyz::xyz_build_swcr_token_code_submission(&code);
-                    let framed = xyz::xyz_wrap(&submission);
-                    stream.write_all(&framed)?;
-                    log::info!(
-                        "2FA gate: submitted SWCR_TOKEN state=3 code (len={}, {} bytes framed)",
-                        code.len(), framed.len(),
-                    );
-                    code_submitted = true;
+                    // Don't burn a single-use code on a login we're about to
+                    // abandon: the deadline check at the top of the loop would
+                    // fire before the server's answer could be read.
+                    if Instant::now() < deadline {
+                        submit_swcr_code(stream, &result?)?;
+                        code_submitted = true;
+                    }
                 }
                 Err(std::sync::mpsc::TryRecvError::Empty) => {}
                 Err(std::sync::mpsc::TryRecvError::Disconnected) => {
-                    return Err(ib_key_err(
-                        io::ErrorKind::Other,
-                        "2FA gate: code_provider panicked without returning a code",
-                    ));
+                    return Err(provider_panicked());
                 }
             }
         }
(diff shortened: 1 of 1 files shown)


userFRM:
Pushed 7043319. The polling-only version made unattended logins pay for the fix: a provider that reads a code from a vault or an env var resolves immediately, but the code then sat until the server's next probe, up to ~20 s later. That is the case #176 is aimed at, so it should not be the slow one. The gate now waits inline for one second before falling back to the probe-driven path, which covers a programmatic provider outright and stays far below the probe cadence, so a provider waiting on a human still keeps the keepalives flowing.

Two smaller ones from the same pass: the submission is skipped once the client deadline has passed, so a single-use code is not burned on a login that is about to return `TimedOut`; and the provider thread goes through `thread::Builder`, so a spawn failure is an `io::Error` on the login path rather than a panic.

Tests split to match the two paths: the run-A replay covers the immediate provider and asserts no probe was needed, and a new test covers the human provider and asserts a heartbeat reply precedes the submission. `cargo test --lib auth::` — 98 pass.

userFRM:
A defect in the follow-up commit, fixed in `f24a3d3`.

**The inline grace period could straddle the deadline.** A provider consulted just before expiry returns inside the grace, and its code was submitted unconditionally — then the next loop returned `TimedOut`. Codes are single use, so that spends one on a connection already being abandoned and the operator's next attempt needs a fresh one. The polled path refused for exactly this reason; the fast path did not, which also made the commit's unqualified claim that submission is skipped after the deadline untrue for the route it had just added.

Both passes independently reproduced it with a short deadline and a provider slower than it.

**The re-entrancy question came back clean**, which was the one worth asking here. The callback's whole capability surface is the `Arc`'d closure environment and an owned challenge — no stream handle, no session object — and the socket thread holds no lock while looping. There is no frame through which it can re-enter or wedge the gate. A callback that never returns leaks one named thread while the gate keeps serving probes until the deadline ends the login; a panic drops the sender and both receive arms map that to a clean error.

**Documentation corrected.** `CodeProvider` predated the fast path and said the code goes out on the first inbound message after the callback returns. That holds only once the grace period has passed, and it said nothing about the deadline.

**Stated gaps rather than papered over:**

- The polled path's own deadline guard is still uncovered. Reaching it needs the deadline to expire between the top-of-loop check and the poll within the same iteration — a race, not a sequence. A test aiming at it would be timing-dependent, and I would rather leave it named than add a flaky one.
- The heartbeat test proves ordering, not that a deadline cannot be missed: the harness only produces probes when the client reads, and never disconnects for a late reply. Widening the grace to two seconds still passes it. The claim the commit can support is that the callback no longer blocks the socket thread for its full duration, not that no probe can ever be late.
- Provider-panic detection has no test; neutering both `Disconnected` arms passes the suite.
- Pre-existing, and acknowledged in the code already: the deadline only fires between inbound frames, so a server holding the socket open silently starves it. Live servers probe often enough that it fires in practice.

804 pass plus the two `config::expiry_tests` timezone failures. `--features python` and `--examples` clean.

userFRM:
Updated in `222abc2` — a comment correction.

The comment beside the second sub-case said it gives the polled-path guard its own coverage. It does not: the loop aborts on the deadline before the poll returns the code, so deleting that guard leaves the suite green. The commit message stated this correctly; the comment beside the test did not, and that is the copy a reader reaches first.

Verification worth recording, since it goes past the diff:

- The new guard is exactly complementary to the top-of-loop abort, and after a refusal there is no blocking call before that abort fires — so a refused code had no chance of winning the race. **The guard cannot withhold a legitimate submission.**
- `Instant` is monotonic, so there is no coarse-clock hazard, and the caller's deadline is eighteen minutes by default.
- Nothing leaks: a refusal leaves function-local state consistent and the next iteration returns `TimedOut`; the receiver drops either way; the provider thread was already detached before this change.
- One behaviour delta, left as is: a provider returning an *error* after the deadline now yields `TimedOut` rather than its own error. The sole caller propagates with `?` and never branches on the kind.

Known gaps, stated rather than implied:

- The polled path's own deadline guard is uncovered. Reaching it needs the deadline to expire between the top-of-loop check and the poll within one iteration — a race, not a sequence, and a test aiming at it would be timing-dependent.
- The heartbeat test proves ordering rather than that no probe can be late: the harness only produces probes when the client reads and never disconnects for a late reply, so widening the grace to two seconds still passes it. What the change supports is that the callback no longer blocks the socket thread for its full duration.
- Provider-panic detection has no test; neutering both `Disconnected` arms passes the suite.

804 pass plus the two `config::expiry_tests` timezone failures. All five offline targets clean.

userFRM:

Issue #192, opened 2026-05-27
Percent-based trailing stop (SubmitTrailingStopPct{,Ex}) builds an invalid FIX message — gateway rejects with 'Invalid value in field # 18'


deepentropy:
## Summary

`OrderRequest::SubmitTrailingStopPct` and `OrderRequest::SubmitTrailingStopPctEx` produce a FIX `D` (NewOrderSingle) that the upstream gateway rejects outright. The wrapper sees `order_status status="Inactive" filled=0 remaining=0` and (at default log level) **no error**. With `RUST_LOG=warn`, the rejection reason surfaces from `ccp.rs:603-606`:

```
ExecReport REJECTED: clord=<oid> reason='Invalid value in field # 18' 103=0
```

This affects any caller setting `Order { order_type: "TRAIL", trailing_percent: > 0.0, .. }` against a live IB gateway. The bug is wire-only — unit tests pass because they assert on the `OrderRequest` enum payload, not on the FIX bytes (`src/api/client/tests.rs:405-422`).

## Repro

`examples/hello_limit_order.rs`-style harness, paper account, `cdc1.ibllc.com`:

```rust
let sl = Order {
    action: "SELL".into(),
    total_quantity: 1.0,
    order_type: "TRAIL".into(),
    trailing_percent: 1.0,        // -> 100 basis points
    tif: "GTC".into(),
    outside_rth: true,
    parent_id,                    // bracket child
    oca_group: "bracket_X".into(),
    oca_type: 1,
    transmit: true,
    ..Default::default()
};
client.place_order(sl_id, &spy, &sl)?;
```

Observed status: `Inactive remaining=0 permId=6319067761754558039` (the permId is the FNV hash of the rejected ClOrdID prefix in `perm_id_from_fix_order_id`, which is why it's identical across runs — a useful diagnostic signal).

## Root cause

The amount-based path at `src/engine/hot_loop/order_builder.rs:457-476` is correct and carries the right fields, per its own comment (lines 453-456):

```
// Per ib-agent#136 capture: amount-based trailing stop carries
// the trail amount in both 99 (StopPx) and 211 (PegOffset),
// and requires 18=a (ExecInst = TrailingStop). Without 18,
// the gateway rejects with "Invalid value in field # 18".
```

The percent-based variants at `order_builder.rs:527-544` (`SubmitTrailingStopPct`) and `order_builder.rs:569-597` (`SubmitTrailingStopPctEx`) omit **both**:

[...]

PR #246, opened 2026-07-28
orders: route the adjustable stop through the shared encoder so a bracket child keeps parent, OCA and tif (ibx#240)

### LINKED ISSUE #240: Adjustable stop (SubmitAdjustableStop) ships without parent_id/oca_group and forces DAY -- bracket child is naked
## Summary

`build_order_request` now reads `adjusted_order_type` and routes to
`SubmitAdjustableStop` (good). But that arm bypasses the `SubmitEx` extended-attrs
path, so an adjustable stop used as a bracket child ships **naked, unlinked, and
DAY** -- the same class as #224, for the adjustable path.

## Detail

`client_core.rs:1456` returns early into `SubmitAdjustableStop`:

```rust
if !order.adjusted_order_type.is_empty() {
    ...
    return Ok(ControlCommand::Order(OrderRequest::SubmitAdjustableStop { ... }));
}
// the SubmitEx path below (parent_id 6107, oca_group 583, tif 59) is never reached
```

and the FIX encoder (`order_builder.rs:1042-1084`) emits neither tag 6107
(parent_id) nor tag 583 (oca_group), and hard-codes tag 59 = "0" (DAY):

```rust
let mut fields: Vec<(u32, &str)> = vec![
    ..., (54, side_str), (38, &qty_str),
    (40, "3"), (99, &stop_str),
    (59, "0"),                 // tif = DAY, hard-coded
    ...                        // no 6107, no 583
    (6257, "1"), (6261, adjusted_order_type.fix_code()),
    (6258, &trigger_str), (6259, &adj_stop_str),
];
```

`context.insert_order(Order::new(order_id, instrument, side, qty, 0, ...))` also sets
the tracked parent_id to 0.

## Impact

A move-stop-to-breakeven bracket child (parent + adjustable STP child + target)
ships with the child unlinked from the parent, no OCA with the target, and DAY tif.
It rests/triggers as a standalone order rather than as a bracket leg. This is the
exact failure #224 fixed for ordinary bracket children, still present for the
adjustable-stop encode path.

## Request

Route `SubmitAdjustableStop` through the same extended-attrs handling as `SubmitEx`:
carry parent_id (6107), oca_group (583), and the tif (59). Then a bracket can use an
adjustable stop child safely.

userFRM:
## Summary

- `build_order_request` returned early into `SubmitAdjustableStop` (`src/client_core.rs:1455`), which bypassed the extended-attributes path. The dedicated encoder arm emitted neither 6107 nor 583 and hard-coded 59=0, so an adjustable stop used as a bracket child shipped unlinked, outside its OCA group and DAY — exactly as described.
- The bypass dropped more than those three: outside-RTH, hidden, display size, trigger method, conditions and GTD expiry all live in the same `OrderAttrs` block and were discarded on this path too. Patching the three tags into the dedicated arm would have left the rest broken, so this routes the order type instead of adding tags.
- The adjustable stop is now `OrderKind::AdjustableStop` carried by `SubmitEx`, encoding through `send_order_ex` like every other kind. The standalone encoder arm is gone.
- `Context::submit_adjustable_stop` takes `tif` and `attrs`, matching `submit_limit_ex`.
- Closes #240.

## Wire layout is unchanged

40=3 and 99 sit with the other order-type tags, and the 6257/6261/6258/6259 group plus the conditional 6262 and 6260/6269 are appended after 204 and the attribute block — where the replaced encoder put them. An adjustable stop with no extended attributes serialises to the same tag sequence as before. Tag order should not carry meaning, but this path had a shipped layout and there was no reason to move it as a side effect of the fix.

## Test

The repo had no wire-level assertion on the order path — as noted in #192, the unit tests assert on the `OrderRequest` payload rather than the FIX bytes, which is why this shipped: the enum-level tests passed unchanged the whole time the child was going out naked.

`Connection::for_test()` (`src/protocol/connection.rs`) returns the peer socket alongside the connection, so a test can read what an encoder actually wrote. Two tests use it:

- `adjustable_stop_wire_carries_parent_oca_and_tif` — pins 6107, 583 and 59=GTC for a bracket child, plus 40/99/6257/6261/6258/6259.
- `adjustable_stop_wire_carries_trail_and_limit_tags` — pins 6262 for a stop-limit conversion and 6260/6269 for a trailing one, the *absence* of 6107/583 when no parent or OCA is set, and the relative order of the whole group against 204.

Both fail if the routing is reverted. `cargo test --lib` — 805 pass. The two `config::expiry_tests` failures and the `ib_paper_compat` E0063s are present on `main` before this branch and are untouched by it; the E0599s this change would have added to `ib_paper_compat` are migrated.

## Not fixed here, found while doing it

Two pre-existing issues in the shared path this now routes through. Both predate the change and affect every bracket child, not just adjustable ones, so they are not folded in:

1. **Tag 6107 assumes the parent is at version 0.** `send_order_ex` emits `format!("{}.0", attrs.parent_id)` while child ClOrdIDs come from `modify_versions`. A child submitted after its parent has been replaced references a stale ClOrdID.
2. **Replacement drops the adjustment and the bracket linkage.** The tracked `Order` keeps only ord type, price, tif and stop price, so a modify re-encodes as a plain stop with no 6257–6269, no 583/6209 and no 6107.

Happy to open these as separate issues if they are not already known.

## Not verified live

Paper validation of an actual adjustable bracket child is outstanding — the account hit a logon rate limit mid-session. Since the tag sequence is unchanged, the risk this carries is the routing, which the wire tests cover, rather than the encoding.

## Test plan

- [x] `cargo test --offline --lib` — 805 passed. The 2 failures are `config::expiry_tests::{named_zone_converts_with_dst, instant_round_trips_to_wire}`, which fail on the base commit too: the host has no legacy timezone files (fixed separately in #336).
- [x] Mutation check: reverting the shared-encoder route fails `adjustable_stop_wire_carries_parent_oca_and_tif` by name.


userFRM:
Three cleanups in the compat and encoder paths.

**A duplicate match arm.** Renaming the adjustable-stop request to `SubmitEx` left two identical arms in the compat helper, so the second is unreachable. That target is not built by the library test run, which is why the warning went unseen.

**A comment corrected.** It claimed the adjustable-specific tags "now sit ahead of 59/60/167, where the standalone arm put them after 204". They sit after 204 exactly as before, and the sibling test asserts precisely that — so the comment contradicted the test two screens below it.

**A doc comment reattached.** Inserting `for_test` above `send_fix` left the "state is committed only after `write_all`" paragraph documenting the test helper, and `send_fix` undocumented.

Verification worth recording: the wire layout was compared field by field against the old standalone arm and is identical apart from tag 59, which is the fix — it now carries the caller's time-in-force instead of a hardcoded DAY. The "silently discarded" claim was checked at the source: the client had both the TIF and the attribute block in hand, and the old request type could not carry them. The absence assertions in these tests have positive controls.

Filed separately: adaptive and algo orders bypass the extended-attribute path in the same way this fixes for the adjustable stop, so a GTC, parented, OCA or outside-RTH algo order still drops all of it — same shape, different early return.

userFRM:

PR #249, opened 2026-07-28
orders: assert outside-RTH on a replace from the resubmitted order (ibx#247)

### LINKED ISSUE #247: orders: a modify sends 6433=1 unconditionally, so any order placed with outside_rth=false becomes outside-RTH on first modify
## Summary

Every order modification sends `6433=1` (OutsideRTH) regardless of what the order was placed with, so an order placed with `outside_rth=false` — the default — comes back outside-RTH enabled after any `place_order` with a tracked id.

Unlike a dropped tag, this needs no assumption about how the gateway treats omissions: the tag is on the wire with an affirmative value the caller never asked for.

## Detail

`src/engine/hot_loop/order_builder.rs:1494` (v0.7.1, 9367845), in the Modify arm's field list:

```rust
(6122, "c"),         // Client version
(6433, "1"),         // OutsideRTH (preserve from original)
(38, &qty_str),      // OrderQty
```

It is unconditional. The comment says "preserve from original", but the record the replace is rebuilt from cannot express it — `types::Order` (`src/types.rs:251-266`) holds `{order_id, instrument, side, price, qty, filled, status, ord_type, tif, stop_price}` and has no outside-RTH field.

On placement the tag is conditional, exactly as expected: `order_builder.rs:138-140` and `send_order_ex` at `:1794-1796` emit `6433` only when `attrs.outside_rth` is set. So the value genuinely flips across a modify:

| | placed | after modify |
|---|---|---|
| `outside_rth = false` | no 6433 | **6433=1** |
| `outside_rth = true` | 6433=1 | 6433=1 |

## Impact

Any strategy that adjusts a resting order — a stop being trailed manually, a limit being repriced — silently opts that order into the pre/post-market session on the first adjustment. For a stop that is the difference between triggering only in RTH and triggering on a thin extended-hours print.

It hits every modified order, not only bracket children.

## Fix

The caller already supplies the value: `place_order` with a tracked id takes the full `Order` and routes to Modify, and `client_core` caches the placed `ApiOrder` including `outside_rth` (`track_order`, `src/client_core.rs:768-773`). Neither is consulted — `place_order` forwards only price and qty to `ControlCommand::Modify` (`src/api/client/orders.rs:32-41`).

So either carry the flag on `Modify` from the caller's `Order`, or read it from the cached `ApiOrder`. Emitting `6433` only when it is set matches the placement path.

This is one instance of a wider pattern on the modify path, which I have written up separately.

## Environment

v0.7.1 (9367845), read from the encoder. Not observed on a live modify — the value is unconditional in the source, so the wire result follows from the code rather than from a capture.

userFRM:
## Summary

- The replace hard-coded `6433=1` (`src/engine/hot_loop/order_builder.rs`, Modify arm) with a comment claiming it preserved the original. It cannot: `types::Order` (`src/types.rs:251-266`), the record the replace is rebuilt from, has no outside-RTH field.
- Placement emits 6433 only when the caller sets it, so an order placed with `outside_rth=false` — the default — was opted into the extended session by its first modify. For a resting stop that is the difference between triggering in RTH only and triggering on a thin extended-hours print. It hits every modified order, not only bracket children.
- `OrderRequest::Modify` now carries the flag from the resubmitted order and the replace emits 6433 only when set.
- Closes #247.

## Why the caller's order rather than the cached one

`track_order` overwrites the cached `ApiOrder` with the caller's order at the end of every `place_order`, modify branch included (`src/client_core.rs:768-773`), so the cache holds the *previous* resubmission. Reading it would use a stale flag and ignore an explicit change on this call. Price and qty already come from the resubmitted order, and `place_order` with a tracked id is the documented modify entry point, so caller-supplied is both the coherent source and the consistent one.

## Wire

With the flag set the message is unchanged — 6433 keeps its captured position between 6122 and 38. With it clear, the absent tag is the only difference. The conditional is a literal split rather than a positional insert, so the ordering is correct by construction.

## Test

Nothing in-tree exercised the Modify arm's encoder, which is why a hard-coded tag survived there. `modify_emits_outside_rth_only_when_the_caller_set_it` drives the arm over a loopback socket and asserts `|6122=c|6433=1|38=50|` when the flag is set and no `6433` when clear. Reintroducing the unconditional tag fails it; moving the tag to the tail fails it on the positional assertion.

`cargo test --lib` — 804 pass. The two `config::expiry_tests` failures and the `ib_paper_compat` E0063s are present on `main` before this branch; the six `Modify` literals in that target are updated so this adds no new errors.

## Scope and caveats

**Breaking:** `OrderRequest::Modify` gains a field and `Context::modify` gains an argument, both public. `EClient::place_order` and the Python signatures are unchanged. That reads like the 0.8.0 lane rather than 0.7.2 — your call.

**What this does not claim:** I have no capture for how the gateway treats an *absent* 6433 on a `35=G`. Under either reading the bug this targets is fixed — an RTH-only order can no longer come back extended-hours-enabled, and an explicit `true` still asserts the tag. The open case is placed-`true` then modified with `false`: if absence means "unchanged" rather than "false", the order stays outside-RTH. I could not settle that on paper, because the only readback path is `req_open_orders`, which serves local state that `track_order` has already overwritten with the caller's own values — so it echoes the input rather than reporting the gateway.

## Test plan

- [x] Mutation check: reverting the conditional 6433 fails `modify_emits_outside_rth_only_when_the_caller_set_it` by name.

### DIFF
--- a/src/api/client/orders.rs
+++ b/src/api/client/orders.rs
@@ -38,6 +38,7 @@ impl EClient {
                 order_id: oid,
                 price,
                 qty,
+                outside_rth: order.outside_rth,
             })
         } else {
             ClientCore::build_order_request(order, oid, instrument)?
--- a/src/api/client/tests.rs
+++ b/src/api/client/tests.rs
@@ -393,6 +393,42 @@ fn place_order_adjustable_trail_carries_trailing_amount_and_unit() {
 }
 
 #[test]
+fn modify_carries_outside_rth_from_the_resubmitted_order() {
+    // ibx#247: the replace asserted 6433=1 unconditionally, so an order placed
+    // with outside_rth=false came back outside-RTH after any modify. The flag
+    // has to travel with the modify, since the tracked record has no field for
+    // it.
+    let (client, rx, shared) = test_client();
+    shared.market.set_instrument_count(1);
+    let order = Order {
+        action: "BUY".into(), total_quantity: 1.0, order_type: "LMT".into(),
+        lmt_price: 100.0, outside_rth: false, ..Default::default()
+    };
+    client.place_order(70, &spy(), &order).unwrap();
+    let _submit = rx.try_recv().unwrap();
+
+    // Same id -> modify. Caller still says outside_rth=false.
+    let reprice = Order { lmt_price: 101.0, ..order.clone() };
+    client.place_order(70, &spy(), &reprice).unwrap();
+    match rx.try_recv().unwrap() {
+        ControlCommand::Order(OrderRequest::Modify { outside_rth, .. }) => {
+            assert!(!outside_rth, "a modify must not opt the order into the extended session");
+        }
+        cmd => panic!("expected Modify, got {:?}", cmd),
+    }
+
+    // And it survives when the caller does want it.
+    let rth_out = Order { lmt_price: 102.0, outside_rth: true, ..order.clone() };
+    client.place_order(70, &spy(), &rth_out).unwrap();
+    match rx.try_recv().unwrap() {
+        ControlCommand::Order(OrderRequest::Modify { outside_rth, .. }) => {
+            assert!(outside_rth, "an explicit outside_rth=true must reach the replace");
+        }
+        cmd => panic!("expected Modify, got {:?}", cmd),
+    }
+}
+
+#[test]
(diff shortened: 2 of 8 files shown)


userFRM:

PR #255, opened 2026-07-28
gateway: take the farm channel role from the slot, not the farm name (ibx#253)

### LINKED ISSUE #253: gateway: farm channel role is derived from the literal name "ushmds", so a regional HMDS (euhmds) is established on the trading channel
## Problem

`connect_farm` derives the farm's channel role from a literal US farm name (`src/gateway.rs:510`):

```rust
let channel_id = if farm_id == "ushmds" { "2" } else { "1" };
```

Every non-US HMDS name therefore falls to the `else` and requests channel `1` — the trading channel — for a historical-data transport.

The codebase supports those names elsewhere and expects them: `parse_farm_route` handles regional routes and has a test asserting `zdc1.ibllc.com/euhmds/4000` parses to `euhmds` (`:1913-1914`), and the routing comments say plainly that EU accounts are routed to `eufarm`/`euhmds`/`secdefeu` while only US accounts get `usfarm`/`ushmds` (`:1197-1198`, `:1607-1608`).

So the one place the name is compared for role is the one place that assumes US.

## Impact

An EU-routed account's HMDS transport completes logon and looks established while registered under the wrong channel role. That surfaces as historical-data behaviour that cannot be explained from the client side, on exactly the accounts least likely to be tested against.

## Solution

Derive the role from something that is not a US-specific literal. The slot already identifies HMDS unambiguously (`:1642`), so passing the role explicitly into `connect_farm` is the clean fix; matching on the `hmds` suffix is the one-line version.

## Notes

Read from v0.7.1 (9367845). I have a US-routed account, so I cannot exercise the EU path — this is from the source and from the repo's own regional-routing support.

userFRM:
## Summary

- `connect_farm` derived the farm's channel role from a literal US name (`src/gateway.rs:510`): `if farm_id == "ushmds" { "2" } else { "1" }`. Every regional HMDS name fell to the `else` and requested channel 1, the trading channel, for a historical-data transport.
- The file already supports those names — `parse_farm_route` handles regional routes with a test for `zdc1.ibllc.com/euhmds/4000`, and the comments at `:1197` and `:1607` state that EU accounts route to `eufarm`/`euhmds`/`secdefeu`. The role check was the one place that still assumed US.
- The choice is now `farm_channel_id`, keyed on the `hmds` suffix, so `ushmds`, `euhmds` and `aphmds` all take channel 2 and everything else stays on 1.
- Closes #253.

## Test

`hmds_farms_take_the_data_channel_in_every_region` pins the mapping for all three HMDS regions plus `usfarm`, `eufarm` and `secdefeu`. It is a free function beside `parse_farm_route`, so the mapping is testable without standing up a socket — which is why it had no coverage before.

`cargo test --lib` — 803 pass, clippy clean. The two `config::expiry_tests` failures are present on `main` before this branch.

## Not verified live

I have a US-routed account, so the EU path is unexercised here. The change is a pure widening: `ushmds` keeps the exact value it had, and only names that previously got the trading channel by accident are affected.

## Test plan

### DIFF
--- a/src/engine/hot_loop/mod.rs
+++ b/src/engine/hot_loop/mod.rs
@@ -10,7 +10,7 @@ use std::io;
 use crate::bridge::{Event, SharedState};
 use crate::engine::context::Context;
 use crate::config::chrono_free_timestamp;
-use crate::gateway::{connect_farm, reconnect_ccp, ReconnectAuth};
+use crate::gateway::{connect_farm, reconnect_ccp, ReconnectAuth, FARM_SLOT_HMDS, FARM_SLOT_TRADING};
 use crate::protocol::connection::Connection;
 use crate::protocol::fix;
 use crate::types::{ControlCommand, Fill, InstrumentId, Price, Qty, TbtQuote, TbtTrade, PRICE_SCALE, QTY_SCALE};
@@ -882,7 +882,7 @@ impl HotLoop {
                     &auth.host, "usfarm",
                     &auth.username, &auth.password, auth.paper,
                     &auth.server_session_id, &auth.session_key,
-                    &auth.hw_info, &auth.encoded, 18,
+                    &auth.hw_info, &auth.encoded, FARM_SLOT_TRADING,
                 );
                 let _ = tx.send(result);
             })
@@ -1018,7 +1018,7 @@ impl HotLoop {
                     &auth.hmds_host, &auth.hmds_farm,
                     &auth.username, &auth.password, auth.paper,
                     &auth.server_session_id, &auth.session_key,
-                    &auth.hw_info, &auth.encoded, 17,
+                    &auth.hw_info, &auth.encoded, FARM_SLOT_HMDS,
                 );
                 let _ = tx.send(result);
             })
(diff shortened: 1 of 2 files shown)


userFRM:
Reworked. The first revision keyed the role off the farm name, and the name is not a reliable discriminator.

`cashhmds` is the counterexample, and it is already in this repo (`tests/ib_paper_compat/connection.rs:47`): it ends in `hmds` but the extra-farm path connects it on the trading slot. My suffix rule would have moved it to channel 2, so the fix traded one wrong classification for another. Routing tags also carry whatever the server sends, so no naming invariant exists to lean on.

The caller never needed to guess. `connect_farm` already receives the slot as its last argument — 18 from the trading call site, 17 from the market-data one (`gateway.rs:1652`, `:1656`, and the two reconnect paths). Reading that is correct for every farm name, including both `euhmds` and `cashhmds`. The two literals are now named constants.

Also fixed: my previous insertion split `parse_farm_route`'s doc comment, leaving the parser undocumented and its doc describing the new helper. That was a plain mistake, now restored.

Still not established by this repo: that channel 2 is right for regional HMDS specifically. Tag 6171 assigns slot 17 for every region, so it follows that the routing request should match — but the only direct evidence for the value 2 is the original `ushmds` branch. If you have a capture of a regional account's farm logon, that settles it.

userFRM:

Issue #251, opened 2026-07-28
reconnect: an order marked Uncertain is never reported, and the open-order snapshot contradicts it


userFRM:
## Problem

An order working when the CCP link drops is marked `Uncertain`, and nothing tells the caller.

`handle_disconnect` (`src/engine/hot_loop/ccp.rs:1655-1660`) calls `mark_orders_uncertain` (`src/engine/context.rs:970-980`), which flips every working order — including `PendingSubmit` — by mutating the record in place. It pushes no `OrderUpdate`, so nothing reaches the API layer: `client_core`'s `TrackedOrder` still says `Submitted` or `PendingSubmit`, and `is_open_status` admits both.

The open-order snapshot then reasserts the old answer. `collect_open_orders` unions the client's own records with the cached view, and the cached view still says the order is working — so for the whole outage a caller asking which orders are open is told the opposite of what the callback told them, on the one surface they would check it against.

## Impact

The engine stops believing these statuses and says so to nobody. A caller sees a status the engine no longer holds, and `req_open_orders` corroborates it, for as long as the outage lasts.

## Solution

Report the flip. `mark_orders_uncertain` should surface what it changed so the dispatch layers can carry it through as an unknown status, and the open-order snapshot should stop re-importing an order whose status this client has withdrawn.

## Notes

Read from v0.7.1 (9367845). Verifiable offline: `mark_orders_uncertain` has no `shared` parameter and pushes nothing, and `collect_open_orders` merges the cached view unconditionally.

Split from the original filing, which bundled three further items — see #384 (deciding what became of an order absent from the recovery push), #385 (a `PendingSubmit` that is never acked), and the HMDS warm-up re-arm now fixed in #367.

userFRM:
Closing this to clear the tracker. #409 carries the fork's main if the detail here is still useful.

deepentropy:
Reopening. This was closed on 26/08/2026 as tracker cleanup, but the defect is still present on main (e9d919f, 23/09/2026). Its fix exists only in a fork that is not merged, so it stays open until it is fixed on main. A code review of main confirmed it.

deepentropy:
The official gateway 1040 sends no orderStatus to API clients when the CCP link drops; it sends error 1100 at once, and after the relogin it replays all orders with `35=H|11=*|55=*|54=*` and sends 1101/1102 only when that replay has ended.

[...]

PR #256, opened 2026-07-28
ccp: drop an order the gateway says it does not have, rather than restoring it (ibx#252)

### LINKED ISSUE #252: cancel-reject: an UnknownOrder rejection force-restores the order to Submitted in the engine while removing it from the cache
## Problem

When the gateway answers a cancel with `CxlRejReason 1` (UnknownOrder) — it is telling us the order does not exist on its side — ibx force-restores that order to `Submitted` in the engine and then removes only the shared-cache row. Engine and cache end up asserting opposite things.

`handle_cancel_reject` (`src/engine/hot_loop/ccp.rs:1148-1170`) does the restore first, unconditionally:

```rust
let restore_status = if order.filled > 0 { PartiallyFilled } else { Submitted };
// Deliberate regression (PendingCancel back to working) — the
// ibx#212 guard would rightly block it on the ordinary path.
context.set_order_status_forced(oid, restore_status);
```

and only afterwards branches on the reason:

```rust
if reason_code == 1 {
    shared.orders.remove_order_info(oid);
}
```

The restore is right for `TooLate` and `OrderInProcess`, where the order genuinely is still working and a follow-up exec report reconciles it. For `UnknownOrder` it contradicts the message being handled: the engine is left believing a nonexistent order is live, while the cache row that would have surfaced it is gone.

## Impact

The engine's own view is the one that governs subsequent cancels, modifies and reconnect bookkeeping, so a phantom order persists there with no cache row to make it visible — the inverse of the usual stale-cache problem and harder to notice. It compounds after a reconnect, where a cancel aimed at an order that no longer exists is exactly the expected shape.

## Solution

Branch before restoring: for `reason_code == 1`, remove or terminalize the engine order rather than forcing it back to working, and keep the cache removal. The other reasons keep today's restore.

## Notes

Read from v0.7.1 (9367845). The ordering is plain in the source; what a live gateway sends alongside a reason-1 reject is not asserted here.

userFRM:
## Summary

- `handle_cancel_reject` (`src/engine/hot_loop/ccp.rs:1147-1170`) force-restored the order to `Submitted`/`PartiallyFilled` before reading the reject reason, and only afterwards removed the shared-cache row for `CxlRejReason 1`.
- For UnknownOrder that inverted the message being handled: the gateway says the order does not exist, and ibx responded by putting it back to working in the engine while deleting the row that would have made it visible. Engine and cache then assert opposite things.
- The reason is now read first. UnknownOrder removes the order from the engine as well as the cache; every other reason keeps today's restore, since those mean the order really is still working.
- Closes #252.

## Test

Two tests either side of the branch: `cancel_reject_unknown_order_drops_the_order_rather_than_restoring_it` and `cancel_reject_too_late_restores_the_order_to_working`. Reverting the branch fails the first and leaves the second passing, so it pins the distinction rather than the mechanism.

`cargo test --lib` — 804 pass. The two `config::expiry_tests` failures are present on `main` before this branch.

## Note

The ibx#212 bypass is untouched — PendingCancel back to working is still a regression the ordinary guard would rightly block, so the restore path keeps `set_order_status_forced` and its comment.

Related: #251 covers the wider case where an order the gateway no longer has is never reconciled at all after a reconnect. This PR only fixes the path where the gateway tells us explicitly.

### DIFF
--- a/src/engine/hot_loop/ccp.rs
+++ b/src/engine/hot_loop/ccp.rs
@@ -1144,27 +1144,36 @@ impl CcpState {
 
         let Some(oid) = orig_clord else { return };
 
+        // FIX CxlRejReason 1 = UnknownOrder: the gateway is saying the order
+        // does not exist on its side, so there is nothing to restore it to.
+        // Restoring it anyway left the engine asserting a phantom order was
+        // working while the cache row that would surface it was removed —
+        // opposite halves of the same state disagreeing (ibx#252). Every other
+        // reason (TooLate, OrderInProcess, ...) means the order really is still
+        // working, so the cancel is undone and a follow-up exec report
+        // reconciles.
+        let unknown_order = reason_code == 1;
+
         // Update local context only if we tracked the order in this session.
         let instrument = if let Some(order) = context.order(oid).copied() {
-            let restore_status = if order.filled > 0 {
-                crate::types::OrderStatus::PartiallyFilled
+            if unknown_order {
+                context.remove_order(oid);
             } else {
-                crate::types::OrderStatus::Submitted
-            };
-            // Deliberate regression (PendingCancel back to working) — the
-            // ibx#212 guard would rightly block it on the ordinary path.
-            context.set_order_status_forced(oid, restore_status);
+                let restore_status = if order.filled > 0 {
+                    crate::types::OrderStatus::PartiallyFilled
+                } else {
+                    crate::types::OrderStatus::Submitted
+                };
+                // Deliberate regression (PendingCancel back to working) — the
+                // ibx#212 guard would rightly block it on the ordinary path.
+                context.set_order_status_forced(oid, restore_status);
+            }
             order.instrument
         } else {
             0
         };
 
-        // FIX CxlRejReason 1 = UnknownOrder. The gateway is telling us the
-        // order it just listed in the mass-status burst doesn't exist on its
-        // side — drop the stale cache entry so subsequent req_open_orders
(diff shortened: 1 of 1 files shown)


userFRM:
Closing: this trades one inconsistency for another rather than removing it. **Deleting the record throws away data a racing frame still needs.** `remove_order` discards instrument, side and filled quantity. A fill or terminal report arriving after the reject cannot be processed: the fill path requires `context.order` (`ccp.rs:855`), and completed-order delivery requires it too (`ccp.rs:1112`). So a cancel/fill race loses the fill callback entirely — worse than the split-brain this was fixing.

**The engine/cache agreement I claimed does not hold for same-session orders.** Removing the shared row leaves `ClientCore::open_orders` untouched, and `collect_open_orders` keeps returning the locally tracked `PendingCancel` row. The agreement only ever held for recovery orders that exist solely in the shared cache.

**Deletion also defeats the ibx#212 guard.** With no record, the monotonic status check has nothing to reject, so a delayed `150=0/39=0` frame — or the reconnect recovery burst — takes the missing-order path at `ccp.rs:684` and reinserts the order as Submitted. The phantom this removes can be recreated by the next reconnect.

The right shape is a retained non-open tombstone until authoritative reconciliation, not a delete — and not a synthetic `Rejected` either, which would fabricate a terminal outcome the gateway never reported. That is a design decision about how the engine represents "the gateway says this does not exist", so it belongs to you rather than to a patch from me.

The analysis is on #252 so nothing is lost.

Issue #258, opened 2026-07-28
contracts: byte-guarded string slicing panics on a non-ASCII schedule reply, taking down the hot loop


userFRM:
## Problem

`trim_session_endpoint` (`src/control/contracts.rs:675-687`) validates byte positions and then slices the string, so a multi-byte character straddling one of those boundaries panics:

```rust
let bytes = s.as_bytes();
if bytes.len() >= 14 && bytes[8] == b'-' && bytes[11] == b':' {
    out.push_str(&s[..8]);
    out.push(':');
    out.push_str(&s[9..11]);
    out.push_str(&s[12..14]);
```

The guard establishes that byte 8 is `-` and byte 11 is `:`, and says nothing about bytes 12-14. Slicing a `&str` at a non-char boundary is a panic, not an error.

Reaching it needs only one invalid UTF-8 byte in tag 6841/6842 of a `6040=107` schedule reply. Frame bodies are decoded with `from_utf8_lossy` (`src/protocol/fix.rs:151`), which replaces that byte with U+FFFD — three bytes wide — so a boundary that was ASCII in the intended payload now lands mid-character.

The same shape is nearby at `contracts.rs:656` and `:658` (`&s.trade_date[..8]`, `&s.start[..8]`), which also panic if the field is shorter than eight bytes or non-ASCII at the cut.

## Impact

This runs on the hot loop, so it is an engine-down on malformed input rather than a dropped message. Trading-hours strings are gateway-supplied and lossy-decoded, which is precisely the combination that makes byte-indexed slicing unsafe.

## Solution

Add `s.is_ascii()` to the guard — the format being parsed is ASCII by definition, so that makes the byte reasoning sound rather than incidental — and use `.get(..8)` with a fallback for the two closed slices, so a short or non-ASCII field degrades to the untrimmed value instead of aborting.

## Notes

Read from v0.7.1 (9367845). The `from_utf8_lossy` decode and the byte-guarded slice are both in the source; I have not fed a malformed schedule reply to a live gateway.

userFRM:
Closing this to clear the tracker. #409 carries the fork's main if the detail here is still useful.

deepentropy:
Reopening. This was closed on 26/08/2026 as tracker cleanup, but the defect is still present on main (e9d919f, 23/09/2026). Its fix exists only in a fork that is not merged, so it stays open until it is fixed on main. A code review of main confirmed it.

Issue #233, opened 2026-07-15
market data: MarketState has no unregister and active_count never decrements, so the 256 cap is cumulative-per-session and panics the hot loop


deepentropy:
## Summary

`MarketState` has no unregister and `active_count` only ever increments, so `MAX_INSTRUMENTS = 256` counts contracts **ever subscribed in a session**, not concurrent ones. Crossing it hits an `assert!` that panics the hot loop, killing the engine for the rest of the process.

`cancel_mkt_data` does not give the slot back.

## The code

`src/engine/market_state.rs:48-59`:

```rust
pub fn register(&mut self, con_id: i64) -> InstrumentId {
    if let Some(&id) = self.con_id_to_instrument.get(&con_id) {
        return id;
    }
    let id = self.active_count;
    assert!((id as usize) < MAX_INSTRUMENTS, "too many instruments");   // <-- :54
    self.con_id_to_instrument.insert(con_id, id);
    self.instrument_to_con_id[id as usize] = con_id;
    self.active_count += 1;
    id
}
```

There is no `unregister`. `active_count` has no decrement anywhere. `con_id_to_instrument` is never removed from. `cancel_mkt_data` → `unregister_mkt_data` (`client_core.rs:514-524`) cleans the **client** side (`req_to_instrument`, `last_quotes`, …) and sends `Unsubscribe`, but the engine's slot is gone for good.

So a session that subscribes and unsubscribes 257 distinct symbols — a watchlist app over a long day — panics on the 257th, even with one subscription live.

## What the panic does

`run_with_panic_recovery` (`mod.rs:193-211`) catches it, logs `"Engine hot loop panicked, emitting Disconnected: too many instruments"`, emits one `Event::Disconnected`, and **the hot loop is gone**. No restart. Every later command fails with `"Engine stopped"`.

Worse, the caller is blocked in `recv_registration(reply_rx)` on a `bounded(1)` whose sender was just dropped by the unwind — so `place_order` is where a user finds out.

## The related defect: one instrument, two req_ids, and the first goes silent

`client_core.rs:502`:

```rust
self.core.instrument_to_req.lock().unwrap().insert(instrument_id, req_id);
```

[...]

PR #259, opened 2026-07-28
engine: stop gateway input panicking the hot loop (ibx#257, ibx#258)

### LINKED ISSUE #257: ccp: a session-recovery frame panics the hot loop when the instrument table is full (register vs try_register)
## Problem

A recovery frame from the gateway can panic the hot loop, taking the whole engine down, when the instrument table is full.

The session-recovery insert in `handle_exec_report` registers the contract unconditionally (`src/engine/hot_loop/ccp.rs:703`):

```rust
if con_id != 0 && qty > 0 {
    let instrument = context.register_instrument(con_id);
```

`Context::register_instrument` (`src/engine/context.rs:893-894`) calls `MarketState::register`, which is the panicking wrapper (`src/engine/market_state.rs:94-96`):

```rust
pub fn register(&mut self, con_id: i64) -> InstrumentId {
    self.try_register(con_id).expect("too many instruments")
}
```

#233 converted the client-facing path to `try_register` plus a rejection (`mod.rs:232`), which is why a caller can no longer trigger this. The recovery path was not converted, and it is now the only non-test caller of the panicking wrapper.

## Impact

The trigger is not a caller mistake — it is inbound data. On reconnect the gateway replays every resting order across sessions, so an account holding orders on more than 256 distinct contracts panics the engine during recovery with zero market-data subscriptions taken out. The reconnect burst is exactly when the table is most likely to fill, and a panic there is unrecoverable rather than a rejected request.

## Solution

Use `try_register` here as well: log and skip the insert when the table is full, so recovery degrades to a missing order rather than a dead engine.

Once this caller is converted, `MarketState::register` has no non-test callers left — deleting it (or moving it behind `#[cfg(test)]`) closes the class rather than leaving the next caller a panicking convenience wrapper to reach for.

## Notes

Read from v0.7.1 (9367845). The panic is a single `expect` on a path fed directly by gateway frames; no live run needed to see it.

userFRM:
## Summary

Two panics reachable from gateway frames, both on the hot loop. A panic there is an engine-down, not a dropped message, which is what puts them together in one change.

**#257 — recovery registers through the panicking wrapper.** The session-recovery insert (`src/engine/hot_loop/ccp.rs:703`) called `context.register_instrument`, which resolves to `MarketState::register` — `try_register(..).expect("too many instruments")` (`market_state.rs:94-96`). #233 converted the client-facing path to `try_register` plus a rejection and left this caller on the old wrapper, and this is the one fed directly by inbound data: a reconnect replays every resting order across sessions, so an account holding orders on more than 256 distinct contracts aborted mid-recovery having taken out no subscriptions at all. It now uses a fallible `try_register_instrument` and drops the order with a warning.

**#258 — byte-guarded string slicing.** `trim_session_endpoint` (`src/control/contracts.rs:675`) checked byte positions and then sliced the `&str`. Frame bodies are decoded with `from_utf8_lossy` (`protocol/fix.rs:151`), so one invalid byte in tag 6841/6842 of a `6040=107` reply becomes a three-byte U+FFFD and `&s[12..14]` cuts inside it. The guard now requires `s.is_ascii()`, which is what makes reasoning in bytes sound rather than accidental. `format_sessions_string` had the same shape twice (`:656`, `:658`) and now uses `get(..8)`, which rejects a non-char boundary as well as a short field.

Closes #257 and #258.

## Test

`session_endpoint_survives_a_lossily_decoded_field` and `sessions_string_survives_a_non_ascii_trade_date` feed exactly the shape `from_utf8_lossy` produces. Removing the ASCII guard reproduces the original abort verbatim — `byte index 14 is not a char boundary; it is inside '\u{FFFD}' (bytes 12..15)` — so the test pins the panic rather than the phrasing of the fix.

## Note

`MarketState::register` now has no non-test callers. Leaving a panicking convenience wrapper in place is how this recurred after #233, so it is worth deleting or putting behind `#[cfg(test)]` — I left it alone here to keep the change to the defect.

## Test plan

### DIFF
--- a/src/control/contracts.rs
+++ b/src/control/contracts.rs
@@ -652,13 +652,12 @@ pub fn format_sessions_string(sessions: &[ScheduleSession]) -> String {
     for (i, s) in sessions.iter().enumerate() {
         if i > 0 { out.push(';'); }
         if s.start == s.end {
-            let date = if s.trade_date.len() >= 8 {
-                &s.trade_date[..8]
-            } else if s.start.len() >= 8 {
-                &s.start[..8]
-            } else {
-                s.start.as_str()
-            };
+            // get() rather than a length check plus a slice: it rejects a
+            // non-char boundary as well as a short field, so a lossily-decoded
+            // frame cannot panic here (ibx#258).
+            let date = s.trade_date.get(..8)
+                .or_else(|| s.start.get(..8))
+                .unwrap_or(s.start.as_str());
             out.push_str(date);
             out.push_str(":CLOSED");
         } else {
@@ -674,7 +673,11 @@ pub fn format_sessions_string(sessions: &[ScheduleSession]) -> String {
 /// Returns the input unchanged if the format does not match.
 fn trim_session_endpoint(s: &str) -> String {
     let bytes = s.as_bytes();
-    if bytes.len() >= 14 && bytes[8] == b'-' && bytes[11] == b':' {
+    // The reasoning below is in bytes, so it is only sound on ASCII. Frame
+    // bodies are decoded with from_utf8_lossy, and one invalid byte becomes a
+    // three-byte U+FFFD that can straddle a slice boundary — a panic on the
+    // hot loop rather than a bad value (ibx#258).
+    if s.is_ascii() && bytes.len() >= 14 && bytes[8] == b'-' && bytes[11] == b':' {
         let mut out = String::with_capacity(13);
         out.push_str(&s[..8]);
         out.push(':');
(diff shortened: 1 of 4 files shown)


userFRM:
Pushed a correction on both halves. **The early return hid live orders.** On a full instrument table I returned from the whole execution-report handler, which skipped the `last_clord` recording and the rich-order cache push further down. An order the engine could not track therefore also disappeared from `req_open_orders` and could not be cancelled by perm id — a worse outcome than the panic for anyone holding that order. Only the engine-book insert is skipped now; the rest of the handler runs.

**The trade-date test was passing against unfixed code.** `format_sessions_string` only reaches the `trade_date` slice on the closed-session branch, which needs `start == end`, and I had set them differently — so the test never exercised `get(..8)` at all. Corrected, and it now reproduces `byte index 8 is not a char boundary` when the fix is reverted. The endpoint test was already genuine.

One clarification on scope, since it came up: `try_register` still resolves a conId that is already registered even at capacity, so this only skips genuinely new contracts. And a full table can hold retained registrations rather than 256 live resting orders, so the trigger is broader than "an account with 256 orders" — reclaim behaviour matters as much as order count.

userFRM:
The review found a live panic of this exact class that the commit did not cover, and it fires at the worst possible moment.

`hmds.rs` byte-sliced a lossily decoded XML tag inside a debug log. A tag-6118 value of 199 ASCII bytes followed by one invalid byte panics with "byte index 200 is not a char boundary" — proven through the real parser, not reasoned about. So the hot loop aborts only when debug logging for that module is on, which is to say only while someone is diagnosing an incident. Made total, and the rest of the tree was swept for the same shape: every other slice is either bounds-guarded or anchored on an ASCII boundary, so that was the only remaining live one.

**The headline change had no regression test.** Reverting the fallible instrument registration back to the panicking form passed the entire suite. There is one now: it fills every slot, drives a recovery record through the handler, and requires it to return rather than abort — with the acknowledged cost, an untracked order, asserted rather than glossed.

Worth recording from the review, since it is the part I could not establish alone: the touched paths were enumerated for panic surface and every tag parse is fallible, every float-to-integer cast saturates, the order-id slice is anchored on an ASCII boundary, and instrument indices are bounded by the registration cap. Truncated, empty, oversized and non-UTF-8 payloads each terminate — frame extraction either breaks or consumes at least one byte per iteration, so none of them loop. And the guards swallow nothing silently: a full table warns and leaves the acknowledged gap, and a non-ASCII endpoint is returned unchanged per the function's stated contract.

userFRM:

PR #264, opened 2026-07-28
ccp: treat an absent position tag as no quantity, not as flat (ibx#261)

### LINKED ISSUE #261: ccp: an account-update frame without tag 6064 flattens a live position to zero instead of leaving it unchanged
## Problem

An account-update frame that omits the position tag flattens the position to zero (`src/engine/hot_loop/ccp.rs:1954-1957`):

```rust
let position: i64 = parsed.get(&6064)
    .and_then(|s| s.parse::<f64>().ok())
    .map(|v| v as i64)
    .unwrap_or(0);
```

Absent is read as flat rather than as unchanged. Everything downstream then treats the account as having no position: `PositionInfo.position` is set to 0, `context.positions` is reconciled to 0 through the delta at `:1985-1993`, `PortfolioState::positions` is overwritten, and a `PositionUpdate { position: 0 }` is emitted.

A marks-only update — a frame that carries the mark fields and no quantity — is exactly the shape that triggers it.

## Impact

`req_positions` and both P&L paths compute against a flat book while the account is holding. A strategy that sizes from current position, or a risk check that refuses to exceed a limit, sees zero and acts on it. It corrects on the next frame that does carry 6064, so it presents as an intermittent wrong number rather than a persistent one, which is worse to diagnose.

## Solution

Treat an absent 6064 as "no quantity in this frame": update the marks and leave the position alone, rather than defaulting to 0. The same file already distinguishes ack-shaped frames from data-shaped ones elsewhere (the tag-146 guard at `:352`, the what-if guard at `:761`), so the pattern is established.

## Notes

Read from v0.7.1 (9367845). The default is plain in the source; whether the live gateway emits marks-only frames on this message is not asserted here.

userFRM:
## Summary

- `handle_position_update` (`src/engine/hot_loop/ccp.rs:1954`) read tag 6064 with `unwrap_or(0)`, so a frame carrying marks and no quantity was read as the account being flat rather than as the frame carrying no quantity.
- The zero propagated everywhere: `PositionInfo.position`, `context.positions` via the delta reconciliation, `PortfolioState::positions`, and a `PositionUpdate { position: 0 }` event. `req_positions` and both P&L paths computed against a flat book while the account was holding, correcting only on the next frame that carried 6064.
- The quantity is now `Option<i64>`. Marks are applied either way — a marks-only frame is precisely the case being misread — and the quantity-dependent work is skipped when there is none. A frame that genuinely reports 0 still flattens the position.
- Closes #261.

## Test

`marks_only_frame_does_not_flatten_a_live_position` establishes 100 shares, sends a marks-only frame, and asserts both the engine and `req_positions` still report 100, then confirms an explicit `6064=0` still flattens. Restoring `unwrap_or(0)` fails it on the first assertion.

## Note

The intermittency is what makes this one unpleasant to diagnose from the outside: it self-corrects on the next full frame, so it presents as a position that reads zero for a moment rather than as a persistent wrong number. A strategy that sizes from current position, or a risk check that reads it, only has to look during that window.

I have not asserted how often the live gateway sends marks-only frames on this message — the defect is that an absent tag is read as a value.

## Test plan

- [x] `cargo test --offline --lib` — 806 passed. The 2 failures are `config::expiry_tests::{named_zone_converts_with_dst, instant_round_trips_to_wire}`, which fail on the base commit too: the host has no legacy timezone files (fixed separately in #336).

### DIFF
--- a/src/engine/hot_loop/ccp.rs
+++ b/src/engine/hot_loop/ccp.rs
@@ -1951,10 +1951,17 @@ pub(crate) fn handle_position_update(
         Some(v) => v,
         None => return,
     };
-    let position: i64 = parsed.get(&6064)
+    // An absent quantity means this frame carries no quantity, not that the
+    // account is flat. Defaulting to 0 reconciled the engine's position to zero
+    // off a marks-only frame and published a flat book to reqPositions and both
+    // P&L paths until the next frame that did carry 6064 (ibx#261).
+    let position: Option<i64> = parsed.get(&6064)
         .and_then(|s| s.parse::<f64>().ok())
-        .map(|v| v as i64)
-        .unwrap_or(0);
+        // `"NaN".parse()` succeeds and `NaN as i64` is 0, so a non-finite
+        // value would flatten a live position by the same route the absent
+        // tag did. Route it to no-data instead.
+        .filter(|v| v.is_finite())
+        .map(|v| v as i64);
     // Tag map verified against the updatePortfolio callback (ib-agent#172):
     // 6101 = averageCost, 6065 = marketPrice (per share), 6067 = marketValue,
     // 6100 = unrealizedPNL, 6099 = realizedPNL. Earlier code read 6065 as the
@@ -1963,7 +1970,13 @@ pub(crate) fn handle_position_update(
         .and_then(|s| s.parse::<f64>().ok())
         .map(|v| (v * PRICE_SCALE as f64) as Price)
         .unwrap_or(0);
-    let avg_cost: Price = price_tag(6101);
+    // The average cost is written into a row that persists, so an absent tag
+    // must not overwrite a real one with zero — the same rule the quantity
+    // above follows. Marks are refreshed every frame and are handled apart.
+    let avg_cost: Option<Price> = parsed.get(&6101)
+        .and_then(|s| s.parse::<f64>().ok())
+        .filter(|v| v.is_finite())
+        .map(|v| (v * PRICE_SCALE as f64) as Price);
     let market_price: Price = price_tag(6065);
     let market_value: Price = price_tag(6067);
     let unrealized_pnl: Price = price_tag(6100);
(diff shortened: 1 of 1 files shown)


userFRM:
Pushed a correction: the first revision introduced a window of its own. **The reordering was wrong.** I moved `set_position_marks` ahead of `set_position_info` so marks would apply to both frame shapes. But that setter inserts a default `PositionInfo` when the contract is unknown (`src/bridge.rs:733`), so for a quantity-bearing frame it briefly published a row with `position: 0` before the real quantity landed — a false flat visible to a concurrent reader, which is the same defect this PR exists to remove. The quantity path now writes the info row first, exactly as before, and a marks-only frame updates an existing row without creating one.

**A marks-only frame for an unseen contract must create nothing**, for the same reason. Added a test for it.

**Two things this deliberately does not do**, both flagged rather than guessed:

Marks-only frames no longer refresh `avg_cost`, symbol, security type, currency or multiplier. Updating those without touching the quantity needs a partial-update path the portfolio state does not have. They keep their previous values instead of being zeroed, which is the safer of the two wrong answers, but it is a limitation.

And the open question underneath all of this: if the gateway ever signals a closed position by *omitting* 6064 rather than sending `6064=0`, this change makes a closed position persist rather than a live one read flat. I have no evidence either way, and nothing in the repo establishes it. That is the mirror risk and it is worth a capture before this is considered settled.

userFRM:
The same defect had a price-side sibling inside the very function this fixes.

`avg_cost` was read with a bare default and written unconditionally into a row that persists, so a frame stating a quantity but omitting tag 6101 replaced a real average cost with zero — absent read as a value, exactly what the quantity half corrects. It now keeps what is stored when the tag is absent. Marks are refreshed every frame and are handled apart, so they do not need the same treatment.

Worth recording from the review, since it is the part I could not prove alone: the fix distinguishes absent from an explicit zero **by construction** — absence never writes and never emits, so no reader downstream can see a fabricated zero, and an explicit `"0"` still flattens. The skip cannot strand partial state either, because these frames are single-position tag maps rather than repeating groups. Shorts keep their sign through the finite filter.

The sweep for further sites of this class in that tree found two more, both already fixed on sibling branches rather than here: the group-level quantity on the lean position feed, and the midnight-seed quantity — the latter being the third occurrence, which #299 now covers. That makes four sites of one defect across this file, which is worth knowing about the shape of the original code rather than about any one of these changes.

userFRM:

PR #266, opened 2026-07-28
api: snapshot executions before replaying them into callbacks (ibx#265)

### LINKED ISSUE #265: api: req_executions holds the executions mutex across user callbacks, deadlocking a re-entrant wrapper and freezing the interpreter in Python
## Problem

`req_executions` holds the executions mutex while it invokes user callbacks (`src/api/client/orders.rs:163-173`):

```rust
let indices = self.core.filter_executions(filter);
let execs = self.core.executions.lock().unwrap();
for i in indices {
    let se = &execs[i];
    wrapper.exec_details(req_id, &se.contract, &se.execution);
    wrapper.commission_and_fees_report(&se.commission_and_fees);
}
```

The Python adapter does the same (`src/python/compat/client/orders.rs:188-241`), where the callback is arbitrary Python executed with the GIL held.

**Single thread.** A callback that re-enters any path locking `executions` deadlocks, because `std::sync::Mutex` is not reentrant. Re-requesting from inside `exec_details` is an ordinary ibapi pattern. In Python the deadlocked thread owns the GIL, so it is not one stuck call — the interpreter stops.

**Two threads.** While the replay loop sits in user code, the dispatch thread reaches `push_execution` (`src/api/client/dispatch.rs:102`) and blocks on the same mutex, so all fill and order dispatch stalls behind however long the callback takes.

There is also an indexing race: `filter_executions` returns raw indices and releases the lock, then the caller re-locks and does `&execs[i]`. `disconnect()` calls `core.reset()`, which clears the vector (`api/client/mod.rs:297`, `client_core.rs:405`). A disconnect landing between the two locks makes the index out of bounds and panics in the caller's thread.

## Impact

A hang rather than an error, in the callback pattern the API invites, and interpreter-wide in Python.

## Solution

Clone the matching entries under one short lock, release it, then dispatch — the discipline `dispatch.rs:142-150` already uses everywhere else in the same crate. That fixes the deadlock, the dispatch stall and the index race together, since indices never outlive the guard.

## Notes

Read from v0.7.1 (9367845), and verifiable in-process: evaluating `executions.try_lock().is_err()` inside `exec_details` returns true today.

userFRM:
## Summary

- `req_executions` (`src/api/client/orders.rs:163-173`) held the executions mutex across `exec_details` and `commission_and_fees_report`. A wrapper re-entering any path that locks `executions` deadlocks, since `std::sync::Mutex` is not reentrant — and re-requesting from inside `exec_details` is an ordinary ibapi pattern.
- The Python adapter (`src/python/compat/client/orders.rs:188`) held it while calling into Python with the GIL owned, so the same deadlock stops the interpreter rather than one thread.
- Concurrently, the dispatch thread blocks on `push_execution` (`dispatch.rs:102`) for as long as the callback runs, stalling fill and order delivery behind user code.
- The same lines handed raw indices across a lock release: `filter_executions` returns positions, the caller re-locks and indexes, and `disconnect()` clears the vector via `reset()` in between — an out-of-bounds panic in the caller's thread.
- Both paths now snapshot under one short lock and dispatch afterwards. Closes #265.

## Why this shape

`dispatch.rs:142-150` already does exactly this — lock, clone, drop, then call the wrapper — so this brings the one outlier in line rather than introducing a new pattern. Snapshotting also removes the index race for free, because positions never outlive the guard that validated them.

## Test

`req_executions_does_not_hold_the_lock_across_callbacks` drives a wrapper that calls `executions.try_lock()` from inside `exec_details` and asserts it succeeds, plus that the row is still replayed. Restoring the held-lock version fails it.

## Test plan

### DIFF
--- a/src/api/client/orders.rs
+++ b/src/api/client/orders.rs
@@ -161,10 +161,11 @@ impl EClient {
     /// Replays stored executions (optionally filtered), firing `exec_details` +
     /// `commission_and_fees_report` for each, then `exec_details_end`.
     pub fn req_executions(&self, req_id: i64, filter: &ExecutionFilter, wrapper: &mut impl Wrapper) {
-        let indices = self.core.filter_executions(filter);
-        let execs = self.core.executions.lock().unwrap();
-        for i in indices {
-            let se = &execs[i];
+        // Snapshot first: a callback may re-enter a path that locks
+        // `executions`, and the dispatch thread pushes fills through the same
+        // mutex — holding it across user code deadlocks one and stalls the
+        // other (ibx#265).
+        for se in self.core.snapshot_executions(filter) {
             wrapper.exec_details(req_id, &se.contract, &se.execution);
             wrapper.commission_and_fees_report(&se.commission_and_fees);
         }
--- a/src/api/client/tests.rs
+++ b/src/api/client/tests.rs
@@ -1035,6 +1035,86 @@ fn cancel_order_sends_cancel_command() {
     }
 }
 
+/// ibx#265: the executions mutex must not be held while user callbacks run.
+/// A wrapper that re-enters a path locking `executions` is an ordinary ibapi
+/// pattern (re-requesting from `exec_details`), and holding the lock across it
+/// deadlocks — in Python with the GIL held, freezing the interpreter.
+#[test]
+fn req_executions_does_not_hold_the_lock_across_callbacks() {
+    struct Reentrant<'a> {
+        core: &'a ClientCore,
+        observed_locked: bool,
+        rows: usize,
+    }
+    impl Wrapper for Reentrant<'_> {
+        fn exec_details(&mut self, _r: i64, _c: &Contract, _e: &crate::api::types::Execution) {
+            self.rows += 1;
+            // Re-entering while the lock is held is exactly the deadlock.
+            if self.core.executions.try_lock().is_err() {
+                self.observed_locked = true;
+            }
+        }
+    }
+
+    let (client, _rx, _shared) = test_client();
+    client.core.push_execution(
+        1,
+        crate::api::types::Contract { symbol: "AAPL".into(), ..Default::default() },
+        Default::default(),
+        Default::default(),
+    );
+
+    let mut w = Reentrant { core: &client.core, observed_locked: false, rows: 0 };
+    client.req_executions(1, &crate::api::types::ExecutionFilter::default(), &mut w);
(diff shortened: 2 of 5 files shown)


userFRM:
Pushed a correction. The first revision did not close the index race it claimed to — it relocated it.

`snapshot_executions` called `filter_executions` (which takes and releases the lock), then reacquired and dereferenced the returned indices. Using `get()` stopped the panic, but if `reset()` clears the store and new rows arrive between the two acquisitions, those indices address different executions and the replay delivers rows the filter never matched — quieter than the panic and worse. The snapshot now filters and clones under a single acquisition.

The matching rule is extracted to `execution_matches`, shared with `filter_executions`, so the index form and the snapshot form cannot disagree about what matches.

Two things this still does not do, both noted rather than hidden. `filter_executions` remains public and still hands out indices, so any future caller inherits the same trap — it has no callers left in-tree after this change and is a candidate for removal, but that is API surface and yours to decide. And the snapshot deep-clones matching rows while holding the lock; the store is unbounded and cleared only on reset, so an unfiltered replay on a long session is a real copy. Bounding the store, or replaying under a cursor, would be the fix if that ever bites.

userFRM:
Removed the index-returning helper, and corrected the commit message that justified keeping it.

`filter_executions` had **zero callers** — not in the library, not in the tests, nowhere. And it is precisely the racy index shape this change exists to replace: hand back positions into a vector that `reset()` can clear. The message said it was retained "for its own callers", which describes callers that do not exist. Deleted, and the message no longer claims otherwise. Both feature configurations build clean without it.

Also removed a doubled first line in the doc comment, where the old function's summary ended up heading the new one.

Worth recording from the review, because it verified the deadlock in both directions rather than reasoning about it: an external probe that re-calls `req_executions` from inside `exec_details` **completes** on this build and **freezes the interpreter** on the restored old shape, killed only by an external timeout. The Rust regression test fails under the same restoration. Callback re-entry was then enumerated mid-replay with nothing held — placing an order, cancelling, a recursive request, and disconnect all complete; the disconnect joins an engine thread that never runs user code and never needs the GIL. Lock ordering is GIL then executions on every path, with no inversion.

One coverage gap it flagged and I am leaving as such rather than papering over: the Python re-entry path is not pinned by the committed suite — its proof was an external probe — so a regression that reintroduced the hold only in the Python client would pass. Both paths share the same snapshot function, which is the mitigation, not a test.

userFRM:

PR #269, opened 2026-07-28
python: collect engine events before dispatching them into callbacks (ibx#268)

### LINKED ISSUE #268: python: the event drain holds the event_rx mutex across the 1100 callback, so disconnect() from that handler freezes the interpreter
## Problem

The Python event drain holds the `event_rx` mutex across the connectivity-loss callback, so a handler that responds by calling `disconnect()` deadlocks the interpreter.

`dispatch_once` (`src/python/compat/client/dispatch.rs:37-44`):

```rust
if let Some(rx) = self.event_rx.lock().unwrap().as_ref() {
    while let Ok(event) = rx.try_recv() {
        if matches!(event, Event::Disconnected) {
            call_wrapper!(self.wrapper, py, "error", (-1i64, 1100i64, "Connectivity ... lost", ""));
```

`rx` borrows from the temporary `MutexGuard`, so the guard is alive for the whole body — it has to be, or this would not compile. `disconnect()` then locks the same mutex (`src/python/compat/client/mod.rs:198`, `*self.event_rx.lock().unwrap() = None`), and `std::sync::Mutex` is not reentrant. Same thread, GIL held: the whole interpreter stops, not just the call. `connect()` from the same callback deadlocks identically (`mod.rs:170`).

Error 1100 is exactly the callback an ibapi application answers by tearing down and reconnecting, so the trigger is the documented usage rather than an exotic one — and it fires when a connection has just dropped, which is when a frozen interpreter costs the most.

This is the same defect as #265 in a different disguise. The `let`-binding shape that #265 fixed is greppable; this one hides behind an `if let` borrow. Worth noting `reference.rs:280` survives the identical `if let` shape only because it clones — an owned binding lets the guard drop before the body under edition 2024, a borrow does not.

## Solution

Collect the events under a short lock, release it, then dispatch — the same pattern #265 applied:

```rust
let events: Vec<Event> = {
    let guard = self.event_rx.lock().unwrap();
    guard.as_ref().map(|rx| rx.try_iter().collect()).unwrap_or_default()
};
```

## Notes

Read from v0.7.1 (9367845), borrow-checker-grade rather than probed: pyo3 is built here with `extension-module`, so there is no embedded interpreter to deadlock inside a Rust test.

userFRM:
## Summary

- `dispatch_once` (`src/python/compat/client/dispatch.rs:37`) bound its receiver out of the `event_rx` `MutexGuard` with an if-let borrow, so the guard was alive across the `error(-1, 1100, ...)` connectivity-loss callback.
- A Python handler answering 1100 by calling `disconnect()` re-locks that mutex (`mod.rs:198`). `std::sync::Mutex` is not reentrant, and the GIL is held, so it freezes the interpreter rather than failing the call. `connect()` from the same handler is identical (`mod.rs:170`).
- Events are now collected under a short lock and dispatched after release.
- Closes #268.

## Relationship to #265

Same defect, different disguise. #265 was a `let`-bound guard held across `exec_details`; this is an if-let borrow held across `error`. The distinction matters for finding the rest of them: under edition 2024 a `let` binding of an *owned* value drops the guard before the body, so cloning makes the shape safe — `reference.rs:280` uses the identical if-let and is fine precisely because it clones. A borrow out of the guard cannot drop early, and the borrow checker requires it to stay alive, which is why this one compiles and still deadlocks.

Both were swept for across all nine files in `src/python`; this is the only remaining instance.

## Trigger

Error 1100 is the callback an ibapi application answers by tearing down and reconnecting, so this is the documented usage rather than an unusual one — and it fires exactly when a connection has just dropped.

## Test

Not covered by a runtime test: pyo3 is built here with `extension-module`, so there is no embedded interpreter available to a Rust test, and the deadlock needs a Python callback re-entering. The change is borrow-scope only and `cargo check --lib --features python` is clean; `cargo test --lib` — 802 pass, with the two `config::expiry_tests` failures present on `main` beforehand.

If you would rather have this pinned by a test, the natural home is the Python suite under `tests/python` — a wrapper whose `error` handler calls `disconnect()` hangs the run today and returns cleanly after this.

## Test plan

### DIFF
--- a/src/python/compat/client/dispatch.rs
+++ b/src/python/compat/client/dispatch.rs
@@ -34,13 +34,28 @@ impl EClient {
     /// Single iteration of event dispatch: drain all shared queues and fire Python callbacks.
     pub(crate) fn dispatch_once(&self, py: Python<'_>, shared: &Arc<SharedState>) -> PyResult<()> {
         // Drain engine events — surface disconnects as error callbacks.
-        if let Some(rx) = self.event_rx.lock().unwrap().as_ref() {
-            while let Ok(event) = rx.try_recv() {
-                if matches!(event, Event::Disconnected) {
-                    call_wrapper!(self.wrapper, py, "error", (-1i64, 1100i64, "Connectivity between client and server has been lost", ""));
-                    self.connected.store(false, Ordering::Release);
-                }
-            }
+        //
+        // Collect under a short lock and dispatch after releasing it. Binding
+        // `rx` from the guard would hold the mutex across the 1100 callback,
+        // and a handler answering connectivity loss with disconnect() locks the
+        // same mutex — non-reentrant, GIL held, so the interpreter freezes
+        // rather than one call failing (ibx#268, same shape as ibx#265).
+        let events: Vec<Event> = {
+            let guard = self.event_rx.lock().unwrap();
+            guard.as_ref().map(|rx| rx.try_iter().collect()).unwrap_or_default()
+        };
+        // One batch is one session — the channel is replaced on reconnect — so
+        // several `Disconnected` events in it are one loss, and a network cut
+        // that takes both the farm and CCP down emits more than one. Firing per
+        // event meant a handler that reconnected on the first would then take a
+        // stale second 1100 into the new session, marking it disconnected.
+        if events.iter().any(|e| matches!(e, Event::Disconnected)) {
+            // Mark disconnected BEFORE the callback. A handler that answers
+            // 1100 with disconnect() then connect() establishes a new session
+            // and sets connected=true; storing false afterwards would clobber
+            // the new session's state (ibx#268).
+            self.connected.store(false, Ordering::Release);
+            call_wrapper!(self.wrapper, py, "error", (-1i64, 1100i64, "Connectivity between client and server has been lost", ""));
         }
 
         // Drain fills -> execDetails + orderStatus
--- a/src/python/compat/client/orders.rs
+++ b/src/python/compat/client/orders.rs
@@ -245,7 +245,12 @@ impl EClient {
     #[pyo3(signature = (api_only=false))]
     fn req_completed_orders(&self, py: Python<'_>, api_only: bool) -> PyResult<()> {
         let _ = api_only;
-        if let Some(shared) = self.shared.lock().unwrap().clone() {
+        // Bind the clone out of the guard first. A MutexGuard temporary in an
+        // if-let scrutinee lives to the end of the body, so cloning alone does
+        // not release it — a callback re-entering disconnect() would deadlock
+        // on this same mutex (ibx#268).
+        let shared = self.shared.lock().unwrap().clone();
+        if let Some(shared) = shared {
             let completed = shared.orders.drain_completed_orders();
             for co in &completed {
                 let status_str = crate::client_core::order_status_str(co.status);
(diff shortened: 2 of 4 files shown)


userFRM:
Pushed a correction. The central claim in this PR does not hold, and the difference matters.

**Cloning out of the guard does not release it.** The earlier reasoning was that `reference.rs:280` is safe because it clones, and that under edition 2024 an owned binding drops the guard before the body. That is not what edition 2024 changed: it drops an if-let scrutinee temporary before the **else** block, not before a successful body. So the `MutexGuard` in `if let Some(shared) = self.shared.lock().unwrap().clone()` is alive for the whole body.

That leaves two further sites with the same defect: `req_completed_orders` holds `shared` across `completed_order`/`completed_orders_end` (`orders.rs:248`), and `req_market_rule` holds it across `market_rule` (`reference.rs:280`). A callback re-entering `disconnect()` deadlocks on `self.shared` at either. Both now bind the clone in a preceding `let`, which does drop the guard, and my "only remaining instance" claim is withdrawn.

**The disconnected flag was stored after the callback.** A handler answering 1100 with `disconnect()` then `connect()` establishes a new session and sets the flag true — and the stale `store(false)` afterwards clobbered it. It is now set before the callback.

**Still open, and I have not fixed them here:** dispatch continues with the old `SharedState` after a re-entrant `disconnect()` has reset the client, so stale-session callbacks can still fire and repopulate reset state. And `req_executions` in this file holds `core.executions` across its callbacks — that one is fixed in #266, on a separate branch, so whichever lands first the other needs a rebase.

**Test gap I accept:** a deadlock regression wants a killable subprocess with a hard timeout, since an in-process timeout cannot recover once the GIL is frozen. That belongs in `tests/python` and is not written yet.

userFRM:
The watchdog could not fire for the regression it was written to catch.

I used a `threading.Timer` calling `pytest.fail`. That callback needs the GIL — which is exactly what the frozen main thread is holding — so on a real regression the timer never runs and the suite wedges until something outside kills it. The docstring claimed each test "reports instead of wedging"; it did the opposite.

Replaced with `faulthandler.dump_traceback_later`, which arms a timer in C and does not need the interpreter. Verified by restoring the held lock, rebuilding the extension, and running: the suite now aborts at ten seconds with a non-zero exit and names the frozen frame —

```
Timeout (0:00:10)!
Thread 0x... [python] (most recent call first):
  File "tests/python/test_issue_268.py", line 46 in error
  File "tests/python/test_issue_268.py", line 56 in test_disconnect_from_the_1100_handler_does_not_deadlock
```

— pointing straight at the callback that re-entered. With the fix restored, all three pass in hundredths of a second.

Worth recording from the review, because it checked something I had asserted rather than proven: the crate is edition 2024, so whether an `if let` scrutinee's guard still lives through the success body is not obvious. It does — verified with an in-crate probe that fails if you assume otherwise — so both of the order-path sites were real deadlocks rather than prophylactic changes. Each of the four behaviours was also confirmed by mutation against the real extension: restoring the held drain freezes the interpreter, firing per event yields `saw 2`, storing the flag after the callback fails the reconnect test, and reverting the scrutinee shape freezes a probe that disconnects from `completed_order`.

It also swept the rest of the Python adapter: after this commit the only remaining lock held across a callback is in `req_executions`, which is the sibling PR #266. The two touch disjoint hunks and merge cleanly.

userFRM:

PR #273, opened 2026-07-28
tick decode: remove two overflow traps on the price path (ibx#272)

### LINKED ISSUE #272: tick decode: unvalidated VLQ length and extended-width magnitude overflow the price arithmetic — a panic in debug, a fabricated price in release
## Problem

Two unvalidated widths on the tick path reach arithmetic that overflows. In a debug build each is a panic that kills the engine thread; in a release build — which is what ships, since `[profile.release]` sets only `lto = "fat"` and no `overflow-checks` — each silently produces a wrong number that is published as a price.

**Unterminated VLQ in a 35=E frame.** `read_vlq` (`src/protocol/tick_decoder.rs:289-303`) walks to the end of the buffer when no byte sets bit 7, returning however many bytes remained. `vlq_signed` (`:306-314`) then does:

```rust
let bits = 7 * num_bytes;
let half: u64 = 1 << (bits - 1);
```

Eleven trailing bytes gives `bits = 77` and `1u64 << 76`. Debug: shift-overflow panic. Release: the shift is masked (`77 & 63 = 13`), so the sign threshold is computed against the wrong bit width and the decoded price delta is garbage. The existing `if pos >= body.len() { break; }` guards stop `num_bytes == 0`, not a long unterminated run, and the two price fields at `:474` and `:504`/`:508` pass the raw byte count straight through. Reached from the wire via `decode_ticks_35e` (`src/engine/hot_loop/hmds.rs:465`).

**Extended-width 35=P tick.** The extended format reads `byte_width` as a full 8-bit field (`tick_decoder.rs:243-246`), so the magnitude can be 63 bits — up to `i64::MAX`. The apply path (`src/engine/hot_loop/farm.rs:225-231`) then multiplies it unchecked:

```rust
tick_decoder::O_BID_PRICE => { q.bid = tick.magnitude * mts; }
```

with `mts` = `min_tick_scaled`, 1_000_000 for a one-cent tick against `PRICE_SCALE` 1e8. Debug: panic. Release: wraps to an arbitrary signed price, written into the seqlock quote and handed to every consumer. Seven price sites, no clamp between decode and multiply.

## Impact

The release behaviour is the worse of the two. A malformed or truncated frame does not crash and does not log — it fabricates a bid, ask or last that looks like an ordinary number, and everything downstream treats it as one.

## Solution

`vlq_signed` should reject `num_bytes > 9` (nine 7-bit groups already cover the whole `u64` range, so anything longer is malformed by definition) or clamp `bits` to 63.

For the multiply, `saturating_mul` at the price sites is the small fix; bounding `byte_width` and range-checking the magnitude in the decoder is the better one, since the same magnitude also feeds sizes and volume.

Worth considering `overflow-checks = true` on the release profile independently. It would have turned both of these into a loud failure instead of a quiet wrong price.

## Notes

Read from v0.7.1 (9367845). Both were reproduced against this commit with throwaway probes: an unterminated VLQ run overflows the shift, and a bit-packed extended-width frame decodes to `i64::MAX` whose `checked_mul` is `None`.

userFRM:
## Summary

Two unvalidated widths on the tick path reached arithmetic that overflows. Closes #272.

**VLQ width.** `read_vlq` (`src/protocol/tick_decoder.rs:289`) consumes to the end of the buffer when no byte sets bit 7, and `vlq_signed` (`:306`) computed `1u64 << (7 * num_bytes - 1)` from that count — eleven trailing bytes asked for `1u64 << 76`. The width is now clamped to 63. Nine 7-bit groups already cover the whole `u64` range, so a longer run is malformed by definition, and the sign convention is untouched for well-formed widths.

**Price scaling.** The extended 35=P format reads `byte_width` as a full 8-bit field (`:243`), so a magnitude can reach `i64::MAX`, and the apply path multiplied it by `min_tick_scaled` unchecked at seven price sites (`src/engine/hot_loop/farm.rs:225-231`). Those now saturate.

## Why this matters more in release than in debug

`[profile.release]` sets only `lto = "fat"` — no `overflow-checks` — so the shipped build takes the wrapping branch for both. A malformed frame therefore does not panic and does not log: it writes a fabricated bid, ask or last into the seqlock quote, and every consumer downstream treats it as an ordinary number.

Enabling `overflow-checks` on the release profile is worth considering on its own merits, independently of these two sites. I have not changed it here — that is a project-wide call about the performance trade on the hot path.

## Test

`unterminated_vlq_does_not_overflow_the_shift` feeds a run that nothing terminates and asserts the sign convention still holds either side of the one-byte boundary. Restoring the unclamped width reproduces `attempt to shift left with overflow` at the exact line.

## Note on scope

The decoder is arguably the better place to bound the magnitude, since it also feeds sizes and volume, which are assigned without scaling and so cannot overflow today but carry the same unvalidated width. I fixed the sites that can produce a wrong price and left the wider bound to you.

## Test plan

- [x] Mutation check: reverting either tick-by-tick accumulator fails `a_tick_by_tick_price_that_cannot_be_scaled_is_dropped` by name.

### DIFF
--- a/src/engine/hot_loop/farm.rs
+++ b/src/engine/hot_loop/farm.rs
@@ -221,14 +221,56 @@ impl FarmState {
             let mts = context.market.min_tick_scaled(instrument);
             let q = context.market.quote_mut(instrument);
 
+            // The extended 35=P format carries a full 8-bit byte width, so a
+            // magnitude can reach i64::MAX and the scaling multiply wrapped to
+            // an arbitrary price in release builds (ibx#272). Drop the price
+            // rather than pinning it: a saturated value is ~92e9 at
+            // PRICE_SCALE, which reads downstream as an ordinary quote, and
+            // leaving the previous one standing is the honest failure.
+            let scaled = |m: i64| m.checked_mul(mts);
             match tick.tick_type {
-                tick_decoder::O_BID_PRICE => { q.bid = tick.magnitude * mts; }
-                tick_decoder::O_ASK_PRICE => { q.ask = tick.magnitude * mts; }
-                tick_decoder::O_LAST_PRICE => { q.last = tick.magnitude * mts; }
-                tick_decoder::O_HIGH_PRICE => { q.high = tick.magnitude * mts; }
-                tick_decoder::O_LOW_PRICE => { q.low = tick.magnitude * mts; }
-                tick_decoder::O_OPEN_PRICE => { q.open = tick.magnitude * mts; }
-                tick_decoder::O_CLOSE_PRICE => { q.close = tick.magnitude * mts; }
+                tick_decoder::O_BID_PRICE => {
+                    match scaled(tick.magnitude) {
+                        Some(v) => q.bid = v,
+                        None => log::warn!("35=P bid price out of range (magnitude={}), tick dropped", tick.magnitude),
+                    }
+                }
+                tick_decoder::O_ASK_PRICE => {
+                    match scaled(tick.magnitude) {
+                        Some(v) => q.ask = v,
+                        None => log::warn!("35=P ask price out of range (magnitude={}), tick dropped", tick.magnitude),
+                    }
+                }
+                tick_decoder::O_LAST_PRICE => {
+                    match scaled(tick.magnitude) {
+                        Some(v) => q.last = v,
+                        None => log::warn!("35=P last price out of range (magnitude={}), tick dropped", tick.magnitude),
+                    }
+                }
+                tick_decoder::O_HIGH_PRICE => {
+                    match scaled(tick.magnitude) {
+                        Some(v) => q.high = v,
+                        None => log::warn!("35=P high price out of range (magnitude={}), tick dropped", tick.magnitude),
+                    }
+                }
+                tick_decoder::O_LOW_PRICE => {
+                    match scaled(tick.magnitude) {
+                        Some(v) => q.low = v,
+                        None => log::warn!("35=P low price out of range (magnitude={}), tick dropped", tick.magnitude),
+                    }
+                }
+                tick_decoder::O_OPEN_PRICE => {
+                    match scaled(tick.magnitude) {
+                        Some(v) => q.open = v,
+                        None => log::warn!("35=P open price out of range (magnitude={}), tick dropped", tick.magnitude),
+                    }
+                }
+                tick_decoder::O_CLOSE_PRICE => {
+                    match scaled(tick.magnitude) {
+                        Some(v) => q.close = v,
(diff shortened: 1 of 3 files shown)


userFRM:
Pushed a correction. The previous revision had a defect I should not have shipped, so recording it plainly.

**The clamp moved the panic rather than removing it.** At the nine-byte width `bits == 63`, and the sign correction then evaluated `val as i64 - (1i64 << 63)`, which is outside i64 for a well-formed negative value. The byte sequence `40 00 00 00 00 00 00 00 80` is a valid nine-byte VLQ and reproduces `attempt to subtract with overflow` at the same function — the original abort, relocated from the shift to the subtraction. The subtraction is now done in u64 and reinterpreted, which gives the same two's-complement answer at every width with no trap, and the test now pins that boundary either side.

**Saturating was the wrong call.** `i64::MAX` at PRICE_SCALE reads downstream as roughly 92e9 — an ordinary-looking quote a consumer cannot tell from real data. The seven price sites now use `checked_mul` and drop the tick when it does not fit, leaving the previous quote standing, which is the honest failure.

**Scope this does not cover**, since the title claims less than the issue does: the 35=E path accumulates decoded deltas with unchecked addition and multiplies unchecked (`hmds.rs:474`, `:490`, `:530`, `:537`), and the same 8-byte magnitude is published unvalidated as size and volume while 35=E's u64 sizes are cast to i64 (`farm.rs:235`, `hmds.rs:478`, `:492`). Those are real and I have added them to #272 rather than widening this branch.

userFRM:
The failure this change removes survived on its own branch, by another route.

The tick-by-tick path carries the same shape: a wire-controlled delta accumulated into the running price with `+=`, then scaled with an unchecked multiply. A well-formed nine-byte delta — which this very commit makes decodable rather than trapping inside the decoder — flows straight into that and traps in debug or wraps in release, publishing an arbitrary trade or quote price. Same defect, same price path, one route over.

Fixed with it rather than filed separately, since a commit titled "remove the overflow traps on the price path" that leaves one on that path is not finished. Both sides of the book now advance together or not at all, so a partial update cannot leave the pair inconsistent.

Independent verification of the rest, which went further than mine: the signed decoder was checked at every width — zero, one through eight, nine at the boundary in both directions, and ten or more — and the extended tick path confirmed to abort cleanly above eight bytes rather than overflow. Two of the commit's own factual claims were checked rather than taken: the release profile really does carry no overflow checks, and the scale limit really is about 92 billion. Every mutation was caught, including the wrapping multiply reproducing the exact wrapped value the message quotes, to the digit.

Also confirmed: the previously vacuous test really was vacuous, and its replacement discriminates.

userFRM:

Issue #275, opened 2026-07-28
protocol: inbound frame HMAC is computed but never enforced, an absent tag reports valid, and a failed verification still advances the read IV


userFRM:
> **Correction to the threat model below.** The claim that every socket runs inside TLS holds for the auth connection only. `connect_farm` opens a plain `TcpStream` and runs the key exchange and the signed frame stream over it — its own comment calls it "raw TCP" — so the trading and historical-data farm channels have no TLS beneath them. On those channels this is not a defence-in-depth layer sitting behind a working one; it is the layer.

## Threat model first

Every socket here runs inside TLS, and `accept_invalid_certs` is hardcoded `false` at all four construction sites (`src/api/client/mod.rs:182`, `src/python/compat/client/mod.rs:132`, `src/bin/benchmark.rs:158`, `src/gateway.rs:2200`). So none of the below is remotely exploitable against a correct TLS path, and I am not reporting a live vulnerability.

What it is: the application-layer integrity check the protocol specifies is computed on every inbound frame and acted on nowhere, so certificate validation is the only control actually carrying weight. That matters mainly where TLS is terminated by something in the middle, which is common on institutional networks.

I am filing this publicly because the repo has no `SECURITY.md` and private vulnerability reporting is disabled, so there is no private channel — and because the exposure is a missing defence-in-depth layer rather than a remote exploit.

## Problem

`fix_unsign` returns a validity flag (`src/protocol/fix.rs:322`, `:377`, `:385`) and `Connection::unsign` forwards it (`src/protocol/connection.rs:276-287`). Every caller discards it:

`ccp.rs:227`, `:246`, `:253`; `farm.rs:88`, `:107`, `:114`; `hmds.rs:118`, `:137`, `:144`; `gateway.rs:568`, `:589`, `:604`, plus a direct `fix_unsign` with `_valid` at `gateway.rs:285`. Thirteen sites, all binding `_valid` or `_`. There is no site anywhere in the tree that acts on the result, so a tampered frame — order ack, fill, quote, account push — is parsed and applied exactly like an authentic one.

Two things compound it.

**Enforcing the flag would not be enough on its own.** `connection.rs:281-283` returns `(msg, true)` when the frame carries no `8349=` tag at all, so a frame with the tag stripped is reported valid even to a caller that checks.

[...]

PR #277, opened 2026-07-28
connection: do not chain the read IV off a frame that failed verification (ibx#275)

userFRM:
## Summary

`Connection::unsign` (`src/protocol/connection.rs:285`) assigned `self.read_iv = new_iv` before knowing whether the frame verified. Since the undistortion step XORs byte positions using that IV (`fix.rs:338-344`), one frame that failed verification permanently desynchronised the read chain, and every genuine frame afterwards was silently corrupted before parsing.

The IV now advances only when verification succeeded, and a failure logs at warn.

Partially addresses #275.

## What this deliberately does not do

#275 has three legs. This PR fixes one:

1. ~~a failed verification advances the read IV~~ — fixed here
2. the validity flag is discarded at all thirteen call sites
3. `unsign` returns `(msg, true)` when the frame carries no `8349=` tag, so checking the flag would not be sufficient

Legs 2 and 3 both change what the client accepts from the gateway. If unsigned frames legitimately occur on a signed connection — which I cannot determine from the source — enforcing would break connectivity rather than harden it. That is a capture question and yours to answer, so I have left it rather than guess on a path where guessing wrong takes the client offline.

This leg needs no such judgement: chaining off a frame you have just decided not to trust is wrong under either answer, and today it corrupts the stream silently.

## Test

`failed_verification_does_not_advance_the_read_iv` signs a frame, flips a payload byte, and asserts the rejected frame leaves the IV untouched and that a genuine frame still verifies after it. Restoring the unconditional assignment fails it.

`cargo test --lib` — 803 pass. The two `config::expiry_tests` failures are present on `main` before this branch.

## Note

The warn log is the point as much as the IV fix: with the flag discarded everywhere, there is currently no way to know this is happening. Even before legs 2 and 3 are settled, a line in the log turns an invisible stream corruption into something diagnosable.

### DIFF
--- a/src/protocol/connection.rs
+++ b/src/protocol/connection.rs
@@ -282,7 +282,18 @@ impl Connection {
             return (msg.to_vec(), true);
         }
         let (undistorted, new_iv, valid) = fix::fix_unsign(msg, &self.read_key, &self.read_iv);
-        self.read_iv = new_iv;
+        if valid {
+            self.read_iv = new_iv;
+        } else {
+            // Do not chain off a frame that failed verification. The
+            // undistortion step XORs byte positions using this IV, so advancing
+            // it here desynchronised the read chain permanently and every later
+            // genuine frame arrived corrupted, with nothing surfaced (ibx#275).
+            log::warn!(
+                "inbound frame failed HMAC verification ({} bytes) — read IV not advanced",
+                msg.len(),
+            );
+        }
         (undistorted, valid)
     }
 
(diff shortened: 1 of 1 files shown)


userFRM:
Closing this — the reasoning behind it is wrong, and I would rather withdraw it than have it sit here looking plausible.

I claimed not advancing the IV was "correct under either answer". It is not. `fix_unsign` derives the next IV as `iv[i] ^ hmac_res[i]` (`src/protocol/fix.rs:379-383`), where `hmac_res` is computed over `(iv, body)` — entirely independently of whether the signature matched. So if only the `8349` field is damaged in transit, the body is intact, the computed `new_iv` **is** the sender's next IV, and discarding it guarantees the next genuine frame fails. The gateway has already advanced its own chain and cannot know we rejected anything.

Retaining the old IV is right only for an injected or replayed frame the gateway never chained. If the body is damaged, neither policy recovers. So the choice is not between a safe default and a risky one — an HMAC failure needs resynchronisation or a reconnect, not a fixed rule.

My test did not catch this because it modelled the one case the change helps: `tampered.len() - 12` lands inside the `8349=` hex rather than the payload (the commit message calls it a payload byte, which is wrong), and it then re-submitted the same frame signed with the *initial* IV, which is an injected duplicate rather than a genuine frame N+1.

The analysis is on #275 so it is not lost. It matters there: "just stop advancing the IV" is the obvious first fix and it would make things worse in the more likely failure mode.

PR #280, opened 2026-07-28
auth: take the IBKey token sub-type from AUTH_START (ibx#279)

### LINKED ISSUE #279: auth: the IBKey token sub-type is hardcoded rather than read from AUTH_START, so live login fails for any account whose sub-type is not "2a"
## Problem

The IBKey token sub-type is compiled in rather than read from the wire, so the second-factor gate only works on accounts whose sub-type happens to match the captured default. Every other account has its `SWCR_TOKEN` rejected and the socket closed before a challenge is issued — no push is ever dispatched.

`AUTH_START` field 4 names the token type and its sub-type, for example `"5.2i"`: token type 5 (DSA / IBKey), sub-type `"2i"`. The main connect path receives that payload and discards it (`src/gateway.rs:1019`, `let _auth_start = ...`), then enters the gate with `config.ib_key_token_sub_type` (`:1067`), which defaults to `IB_KEY_DEFAULT_TOKEN_SUB_TYPE` = `"2a"` from the ib-agent#123 capture.

The code already documents the hazard, at `src/protocol/xyz.rs:131`:

> `token_sub_type` is computed server-side per session — for accounts matching the captured profile it's `"2a"`, but other accounts/SWCR configurations may differ.

So the value is known to be per-account and per-session, and is then hardcoded anyway. The `ib_key_token_sub_type` override exists but a caller cannot know what to set it to without capturing their own `AUTH_START`.

## Observed

On a live account whose sub-type is `2i`, against v0.7.1:

```
Auth complete
Live login: waiting for second-factor approval (mobile push)
2FA gate: sent SWCR_TOKEN state=1 (40 bytes inner, 48 bytes framed)
Error: 2FA gate: server closed socket before issuing a challenge
```

SRP succeeds; the gate is where it dies. The same account authenticates normally through IB Gateway and through ib_async, both of which take the sub-type from the auth response.

With the sub-type read from `AUTH_START` instead, the same account:

```
2FA gate: token sub-type "2i" (from AUTH_START)
2FA gate: awaiting approval (session_id=205 252, approval_url=...)
2FA gate: approved
Auth: trading farm route = ndc1.ibllc.com/usfarm
Auth: market-data farm route = ndc1.ibllc.com/ushmds/4000
```

The push arrives on the phone, approval completes, and the connection proceeds through CCP logon and farm routing.

## Impact

Live connection is impossible for any account whose sub-type is not `2a`, with a failure that reads like a protocol or account-configuration problem rather than a wrong constant.

## Solution

Parse field 4 of `AUTH_START` and use the sub-type after the dot, keeping `ib_key_token_sub_type` as an override for the case where the field is absent.

## Prior art

This was diagnosed correctly in #129, which listed "Parse token subtype from AUTH_START" among its changes and documented the `"5.2i"` → sub-type `"2i"` derivation. That PR was closed when #111 landed, and the implementation that shipped took the captured constant instead of the parse.

userFRM:
## Summary

`AUTH_START` field 4 names the second-factor token type and its per-session sub-type — `"5.2i"` is type 5 (IBKey), sub-type `"2i"`. The main connect path received that payload and discarded it (`src/gateway.rs:1019`), then entered the gate with `config.ib_key_token_sub_type`, defaulting to the `"2a"` captured in ib-agent#123.

For any account whose sub-type is not `"2a"`, the server rejects the `SWCR_TOKEN` and closes the socket before issuing a challenge. No push is dispatched, so there is nothing for the user to approve, and the error reads like a protocol or account-configuration fault.

The sub-type now comes from the field the server just sent. `ib_key_token_sub_type` remains as an override for the case where `AUTH_START` carries none, and the gate logs which source it used.

Closes #279.

## Verified live

On an account with sub-type `2i`, before:

```
Auth complete
2FA gate: sent SWCR_TOKEN state=1 (40 bytes inner, 48 bytes framed)
Error: 2FA gate: server closed socket before issuing a challenge
```

After:

```
2FA gate: token sub-type "2i" (from AUTH_START)
2FA gate: awaiting approval (session_id=205 252, approval_url=...)
2FA gate: approved
Auth: trading farm route = ndc1.ibllc.com/usfarm
Auth: market-data farm route = ndc1.ibllc.com/ushmds/4000
```

The push arrived on the phone, approval completed, and the session went on to subscribe market data on three contracts. The same account authenticates normally through IB Gateway and ib_async, both of which take the sub-type from the auth response — which is what made a wrong constant, rather than a wrong frame layout, the likely explanation.

## Note on the code's own warning

`src/protocol/xyz.rs:131` already says the value is "computed server-side per session — for accounts matching the captured profile it's `2a`, but other accounts/SWCR configurations may differ", and suggests capturing the live value if the default does not trigger the push. The field carrying that value is in the message immediately preceding the gate, so the capture step is not needed.

## Prior art

#129 diagnosed this correctly and listed "Parse token subtype from AUTH_START" among its changes, with the same `"5.2i"` to `"2i"` derivation. It was closed when #111 landed, and the implementation that shipped used the captured constant instead. This restores the parse on top of the current code.

## Test plan

- [x] Mutation check: dropping the type-5 ownership check fails `the_ib_key_sub_type_comes_from_the_ib_key_entry` by name.
- [ ] Not covered: the call site that uses the parse is inside `connect`, which needs a socket and a completed SRP exchange, so reverting it to an inline split still fails nothing.

### DIFF
--- a/docs/book/src/api/python-reference.md
+++ b/docs/book/src/api/python-reference.md
@@ -46,7 +46,7 @@ def connect(host="cdc1.ibllc.com".to_string(), port=0, client_id=0, username="".
 | `paper` | `bool` | If `true`, connect to paper trading. If `false`, connect blocks on the live second-factor approval window (see method note). |
 | `core_id` | `usize or None` | CPU core affinity for the hot loop thread. Use a distinct value per engine when running several in one process. |
 | `ib_key_timeout_secs` | `int or None` | Live second-factor approval timeout in seconds (default ~18 min). Lower it to fail fast on unattended live logins; ignored for paper. |
-| `ib_key_token_sub_type` | `str or None` | Account-specific second-factor token sub-type (default `"2a"`); ignored for paper. |
+| `ib_key_token_sub_type` | `str or None` | Fallback second-factor token sub-type (default `"2a"`), used only when the server states none for the session; ignored for paper. |
 
 ---
 
--- a/src/gateway.rs
+++ b/src/gateway.rs
@@ -899,11 +899,13 @@ pub struct GatewayConfig {
     /// server-side deadline). Set lower to fail fast for unattended logins.
     /// Only consulted on non-paper logins; paper logins skip the gate entirely.
     pub ib_key_timeout_secs: u64,
-    /// Account-specific second-factor token sub-type used in the SWCR_TOKEN
-    /// state=1 init body (`M.D` field). Default `"2a"` matches the captured
-    /// reference profile; other accounts may use a different value. Capture
-    /// the live value via ib-agent's `SWCR_TOKEN_SUBTYPE` hook if the default
-    /// doesn't trigger the IBKey push for your account.
+    /// Fallback second-factor token sub-type for the SWCR_TOKEN state=1 init
+    /// body (`M.D` field).
+    ///
+    /// `AUTH_START` states the sub-type for the session, and that value wins:
+    /// it is account- and session-specific, so a fixed setting can only ever
+    /// match the profile it came from. This is what gets used when the server
+    /// states none. Default `"2a"` matches the captured reference profile.
     pub ib_key_token_sub_type: String,
     /// If set, the IBKey gate uses the **Challenge/Response** path instead
     /// of waiting for a mobile push approval. After the server delivers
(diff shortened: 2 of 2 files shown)


userFRM:
Updated: as first written this could break an account that worked before.

**It was a regression, not just an omission.** Field 4 carries a comma-separated list when the account has more than one factor enabled, and `split_once('.')` ran on the whole field. So `5.2a,4` produced the sub-type `2a,4` — not a sub-type at all. An account logging in fine on the compiled-in `2a` would have had its token rejected and the socket closed before any challenge, which is the exact symptom this PR exists to remove. `4.auth,5.2i` gave `auth`, the authenticator's sub-type, sent in the IBKey init.

The list is now split before the dot, and the sub-type taken is the one belonging to type 5, since that is the token this gate sends. A sub-type stated against another factor is not IBKey's.

**The change had no tests at all, and that was provably consequential.** Reverting the entire production hunk so the file matched the base byte-for-byte failed nothing — 802 passed either way. Changing `nth(4)` to `nth(3)`, which reads a different field entirely, also passed. The parse is now a named function with a table test covering the single-factor case, both list orders, a sub-type belonging to another factor, and every shape that falls back.

**The setting's documentation was backwards on both clients.** `ib_key_token_sub_type` was described as an account-specific override, and the code silently ignored it whenever the server stated anything. The server's value is right — it is session-specific and a fixed setting cannot be correct for a session it predates — so the setting is a fallback for when `AUTH_START` states none. Rust and Python docs now say that.

**Known gap, stated rather than papered over.** Reverting the *call site* to an inline whole-field split still fails nothing: it lives inside `connect`, which needs a socket and a completed SRP exchange. The parse itself is pinned; the one-line call that uses it is not, and I would rather say so than add a test that only appears to cover it.

Also filed from this pass: the decrypted frame is trusted as `AUTH_START` without checking its inner message type, so any frame with a dotted fifth field is read as the token declaration. Separate concern from this fix.

803 pass plus the two `config::expiry_tests` timezone failures. `--features python` clean.

Note for review order: #284 builds directly on this branch and generalises the helper to return the factor type as well, which is what lets it route an authenticator account.

userFRM:

PR #284, opened 2026-07-29
auth: support the authenticator-code second factor (ibx#282)

### LINKED ISSUE #282: auth: the authenticator-code second factor has no implementation, so those accounts fail live login as a socket close
## Problem

`AUTH_START` names which second factor the account uses. The connect path discards it and routes every account into the IBKey gate.

https://github.com/deepentropy/ibx/blob/main/src/gateway.rs#L1019

```rust
let _auth_start = match session::recv_secure(&mut tls, &mut channel) {
```

https://github.com/deepentropy/ibx/blob/main/src/gateway.rs#L1065

```rust
match session::do_ib_key_2fa(
    &mut tls,
    &config.ib_key_token_sub_type,
```

The token field is `<type>.<subtype>`, and the type selects the exchange. Type 5 is the IBKey push that `do_ib_key_2fa` implements. Type 4 is an authenticator code — a different message with a different shape — and there is no code for it anywhere in the crate.

An account on type 4 therefore has an IBKey message sent at it. The server has no use for it, answers nothing, and eventually closes the socket.

## Impact

Login is impossible on these accounts, and the failure does not say so. What surfaces is a connection torn down partway through auth, which reads as a protocol or network fault. The gate then spends its full deadline waiting for a reply that is never coming before reporting a timeout, so the symptom is an 18-minute hang followed by a misleading error. There is no log line naming the account's actual factor, so nothing points at the cause.

This is separate from #279. That one is about the sub-type being hardcoded within the IBKey path, and fixing it does not help here — type 4 is not an IBKey exchange at all, so no sub-type value makes it work.

## Observed exchange

For type 4 the code goes out as `XYZ 774`, message code 1, with the code in the security-token slot, framed and written raw like the IBKey messages, immediately after SRP passes. The server answers with code 2 carrying `PASSED` or `FAILED`.

The one part that is easy to get wrong: nothing prompts for the code first. The prompt is local to the client, not a server message, so a gate that waits to be asked waits until its deadline.

I have this working against a live type-4 account and will send a PR.

userFRM:
## Summary

- Adds the type 4 second factor, so accounts on an authenticator code can log in — previously there was no implementation and the attempt died as a socket close partway through auth.
- The code goes out as `XYZ 774` message code 1 with the code in the security-token slot, framed and written raw like the IBKey messages, straight after SRP passes. The server answers with code 2; `PASSED` is the accepted verdict and anything else is treated as a rejection.
- Nothing prompts for the code first. The prompt is local to the client, not a server message, so a gate that waits to be asked waits until its deadline. This one sends as soon as the next poll comes round after the provider returns.
- The provider runs off the socket thread, because it normally waits on a human and the loop has to keep answering keepalives meanwhile or the connection is dropped for going quiet. The gate polls the socket while it waits so a code cannot sit unsent for a keepalive interval — an authenticator code is good for about thirty seconds.
- Polling means a read can return part of a frame. `ns_recv` is two `read_exact` calls and drops what it consumed if the second times out, so the next read resumes mid-frame and the connection dies on the magic check. The gate buffers instead, and bounds the claimed length before believing it so a corrupt header cannot grow that buffer toward 4 GiB.
- A rejection is now reported as one. A 774 code 2 `FAILED`, a denial delivered as `AUTH_FINISH`, an unexpected 774 code, and NS error frames each return immediately. Only the first was recognised before, so a refused code was answered with keepalives until the deadline and surfaced as a timeout.
- `IbKeyChallenge` names the factor being answered. The two want different codes — eight characters against the IBKey display, six digits from an authenticator app — and the shipped provider example rejected anything that was not eight characters, so it could not serve this factor.
- Routing is now one pure function, `second_factor_route(paper, token_type)`, matched on at the call site. An `AUTH_START` naming no type keeps going to the IBKey gate, as before — that gate opens by sending its init and reports `Skipped` once the server answers `AUTH_FINISH PASSED`, which is how an account with no second factor completes. Routing it past the gate would send no init and leave the server waiting on one.
- A verdict is judged against whether the code was on the wire when that frame's **first byte arrived**. Judging by the moment it was matched is wrong twice: the socket is read before the provider channel is polled, and a frame can span polls, so a send landing between a verdict's header and its body would make a verdict that predates the code look like an answer to it. The deadline is rechecked between the read and the send for a related reason.
- `EClientConfig` gains `code_provider`, which is also what ibx#208 asks for. This is a new field on a public struct, so struct-literal callers need one line; the in-tree examples are updated.

Stacked on ibx#280 — that PR reads the token type out of `AUTH_START`, which is what selects this path. Review after it, or I can rebase onto main if you would rather take them in the other order.

Closes #282.

## Test plan

- [x] `cargo test --lib` — 826 pass. The two `config::expiry_tests` failures and the `E0063` in `tests/ib_paper_compat` are pre-existing on main, unrelated to this change.
- [x] Twenty-one tests over the gate, one over the routing. Among them: happy path with a byte-exact assertion on the frame, a reply split one byte per read, keepalives answered mid-wait, both rejection routes, the code kept out of error text, an unsolicited verdict ignored, provider absent, deadline honoured.
- [x] Each was checked by reverting the real code it covers rather than by inventing a mutation: restoring the old reader fails the split-read and frame-length tests, calling the provider inline fails the keepalive test, and removing the pre-send guard, the read-time verdict judgment, the length cap, the status sanitisation or the deadline recheck each fails its own test.
- [x] Byte-exact builder tests, verified to fail when the code is moved to an adjacent slot.
- [x] Live login on a type 4 account: gate accepted, both farms connected, historical data farm up, 41s wall clock.
- [ ] Live login on an account with no second factor at all — **I could not construct this case.** Every live account appears to be forced through some factor (app, SMS or authenticator), so an `AUTH_START` naming no type may not be reachable in practice. The route is kept anyway because main called that gate unconditionally, and the gate's own `Skipped` outcome exists for exactly this shape; removing it would be an unforced behaviour change on a path I cannot observe. Worst case the arm is simply never taken. It cannot behave differently from main, because it *is* main's behaviour.
- [x] Live login on a type 4 account with a deliberately wrong code: the server answers `774` code 2 `FAILED` and the gate returns `PermissionDenied` naming it, immediately. Before the rejection arms were added this same reply fell through and was answered with keepalives until the client deadline, so it surfaced as a timeout ~18 minutes later.
- [x] Live login on a type 5 (IBKey push) account: approval delivered, login completed. This PR moved the decision that selects that route into `second_factor_route`, so it is new code on a path that previously worked.
- [x] Paper login: no gate is entered at all, the `None` route. Same reason — paper used to be a `!config.paper` guard at each branch and now goes through the same function, so every existing paper caller is on new code.
- [x] All live runs above were made against this branch's head, not an earlier revision of it — the reader, the verdict handling and the routing all changed after the first live run.

### DIFF
--- a/docs/book/src/api/python-reference.md
+++ b/docs/book/src/api/python-reference.md
@@ -46,7 +46,7 @@ def connect(host="cdc1.ibllc.com".to_string(), port=0, client_id=0, username="".
 | `paper` | `bool` | If `true`, connect to paper trading. If `false`, connect blocks on the live second-factor approval window (see method note). |
 | `core_id` | `usize or None` | CPU core affinity for the hot loop thread. Use a distinct value per engine when running several in one process. |
 | `ib_key_timeout_secs` | `int or None` | Live second-factor approval timeout in seconds (default ~18 min). Lower it to fail fast on unattended live logins; ignored for paper. |
-| `ib_key_token_sub_type` | `str or None` | Account-specific second-factor token sub-type (default `"2a"`); ignored for paper. |
+| `ib_key_token_sub_type` | `str or None` | Fallback second-factor token sub-type (default `"2a"`), used only when the server states none for the session; ignored for paper. |
 
 ---
 
--- a/examples/account_whatif.rs
+++ b/examples/account_whatif.rs
@@ -86,6 +86,7 @@ fn connect() -> Result<EClient, Box<dyn std::error::Error>> {
         host,
         paper: true,
         core_id: None,
+        code_provider: None,
     })?)
 }
 
--- a/examples/ex111_live_2fa_aapl.rs
+++ b/examples/ex111_live_2fa_aapl.rs
@@ -85,7 +85,7 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
     println!("== Connecting LIVE ({}). Approve the second-factor push on your phone when it arrives.", host);
     let t0 = Instant::now();
     let client = EClient::connect(&EClientConfig {
-        username, password, host, paper: false, core_id: None,
+        username, password, host, paper: false, core_id: None, code_provider: None,
     })?;
     println!("== Connected in {:.1}s", t0.elapsed().as_secs_f64());
 
--- a/examples/ex134_ccp_grace.rs
+++ b/examples/ex134_ccp_grace.rs
@@ -53,7 +53,7 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
     println!("== ex134_ccp_grace: connecting to paper ({})...", host);
     let t0 = Instant::now();
     let client = EClient::connect(&EClientConfig {
-        username, password, host, paper: true, core_id: None,
+        username, password, host, paper: true, core_id: None, code_provider: None,
     })?;
     println!("== Connected in {:.1}s", t0.elapsed().as_secs_f64());
 
--- a/examples/ex136_trail_adaptive.rs
+++ b/examples/ex136_trail_adaptive.rs
@@ -123,7 +123,7 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
     println!("== Connecting to paper ({})...", host);
     let t0 = Instant::now();
     let client = EClient::connect(&EClientConfig {
-        username, password, host, paper: true, core_id: None,
+        username, password, host, paper: true, core_id: None, code_provider: None,
     })?;
     println!("== Connected in {:.1}s", t0.elapsed().as_secs_f64());
 
--- a/examples/ex138_lit_rel_opg.rs
+++ b/examples/ex138_lit_rel_opg.rs
@@ -122,7 +122,7 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
     println!("== Connecting to paper ({})...", host);
     let t0 = Instant::now();
     let client = EClient::connect(&EClientConfig {
-        username, password, host, paper: true, core_id: None,
+        username, password, host, paper: true, core_id: None, code_provider: None,
     })?;
     println!("== Connected in {:.1}s", t0.elapsed().as_secs_f64());
 
--- a/examples/ex146_per_request_mode.rs
+++ b/examples/ex146_per_request_mode.rs
@@ -69,7 +69,7 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
     let duration: u64 = env::var("DURATION_SECS").unwrap_or_else(|_| "20".to_string()).parse()?;
 
     let client = EClient::connect(&EClientConfig {
-        username, password, host, paper: true, core_id: None,
+        username, password, host, paper: true, core_id: None, code_provider: None,
     })?;
 
     let contract = Contract { con_id, symbol: symbol.clone(), sec_type: "STK".into(), exchange: "SMART".into(), currency: "USD".into(), ..Default::default() };
--- a/examples/ex150_misc_urls.rs
+++ b/examples/ex150_misc_urls.rs
@@ -15,7 +15,7 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
     let host = env::var("IB_HOST").unwrap_or_else(|_| "cdc1.ibllc.com".to_string());
 
     let client = EClient::connect(&EClientConfig {
-        username, password, host, paper: true, core_id: None,
+        username, password, host, paper: true, core_id: None, code_provider: None,
     })?;
 
     println!("ccp_session_id: {:?}", client.ccp_session_id());
--- a/examples/ex176_live_2fa_code_provider.rs
+++ b/examples/ex176_live_2fa_code_provider.rs
@@ -1,12 +1,14 @@
-//! Verify issue #176: live login through the Challenge/Response variant of the
-//! second-factor approval gate. Instead of tapping Approve on the mobile push,
-//! the example reads the 8-character code from stdin and submits it.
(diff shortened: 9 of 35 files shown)


userFRM:
The extraction that decides which factor to route to had no test at all — and it is the exact code whose wrongness caused both live failures this branch was written for. Changing `fields.get(4)` to `get(3)` passed the entire suite.

It now lives in a function beside the routing decision, with a table pinning field four, a bare type, a type with a sub-type, a comma-separated list, padding, and a short frame. That also gives the sub-type half of this branch a test that fails when it is reverted, which it did not have.

Hardened while it was being extracted: the field was split on the first `.` before being split on `,`, so a hypothetical `5.2i,4` produced a sub-type of `2i,4` — a garbage value in the init frame, which the server answers by closing the socket. Entries are split first now, so a sub-type on one entry cannot swallow the ones after it. Only bare lists have been seen live, so this is hardening rather than an observed failure.

**A code with surrounding whitespace went to the wire verbatim.** The empty check trimmed; the send did not. A provider returning `" 123456 "` put the spaces on the wire and burned the one attempt the operator gets. Trimmed on send now, with a test.

Also adopted a frame larger than one read buffer. Every existing test used a frame that fits in a single 4KiB read — the one size where the read clamp does nothing — so removing the clamp passed everything. A 5KB frame now has to reassemble and stop at its own boundary; without the clamp it does not.

Worth recording from the review: a trace-level capturing logger was installed and the accepted path plus a rejection where the server echoes the code back in every reply slot were both run — no log line and no error string contained the code, only its length. `EClientConfig` structurally cannot derive `Debug`, and the gate logs identifiers rather than fields.

userFRM:
Updated in `9152ca9`.

**Sub-type now follows the routed factor.** Field 4 states a sub-type per entry, and the parser kept the first one it saw regardless of which entry it belonged to — so `4.auth,5.2i` routed to IBKey and sent `auth`. The sub-type is now taken from the entry the gate routes to, using the same preference order as the routing itself.

**Two hunks the feature depends on are now pinned.** Reverting either left the suite green:

- The provider forwarded from `EClientConfig` to the gateway is what makes this reachable from the high-level client, and it was one line in a struct literal built inside `connect`. That mapping is now a function, checkable without a socket, and the test drives the forwarded closure rather than asserting some closure is present.
- The challenge tells the callback which factor it is answering. The authenticator path asserted that; the IBKey path did not. Both now do.

**Documentation corrected.** `code_provider` described itself as IBKey-only and said `None` leaves behaviour unchanged — untrue on an authenticator account, where there is no push to fall back to and `None` fails the connect. The routing rationale had also come detached onto the parsing helper, and the example announced an IBKey challenge on runs that use neither.

## Tracked separately

- **#350** — Python `EClient.connect()` exposes no provider and always passes `None`, so an authenticator account can log in from Rust and not from Python. Fixing it means changing the Python signature and releasing the GIL around a callback that blocks for as long as an operator takes to read a code.
- **#351** — the gate can accept a verdict that predates its own submission: the `sent` guard is evaluated when the frame is read rather than when it became readable.
- **#244 / #245** — the IBKey provider still runs inline on a mixed-factor account. Both changes are needed; neither is a regression from the other.

## Known limitation

Reverting the type-4 dispatch fails no test: the route-to-gate step runs inside `connect`, which needs a socket. Both exchanges were driven against a live account during development — `775` for IBKey and `774` carrying the code for the authenticator — which is evidence for the wiring rather than a substitute for a test.

userFRM:
Rebased onto the corrected #280 and squashed to one commit on top of it.

#280's parse turned out to have a regression of its own — reading the whole of field 4 as a single `type.sub-type` pair, so a list like `5.2a,4` produced the sub-type `2a,4`. Since this branch generalises that helper to return the factor type as well, the fix belonged down there rather than arriving here. #280 now carries the list-aware parse and its table test; this branch extends the same helper with the type, which is what lets it route an authenticator account at all.

Net effect on this PR: the two commits became one, and the parse hunks that were duplicating #280's fix are gone. The feature itself is unchanged from the last review, including the sub-type following the factor rather than the field, the forwarding test that drives the closure, and the factor label pinned on both paths.

830 pass plus the two `config::expiry_tests` timezone failures. `--features python` and `--examples` clean.

userFRM:
It found the sub-type bug surviving in a shape my test did not cover. Fixed in `a2a3ada`.

**`4.auth,5` still sent the wrong sub-type.** I had made the parser prefer type 5 and then type 4 when searching the entries for a sub-type. That is not the same thing as taking the *routed* type's sub-type. On `4.auth,5` routing selects IBKey, because type 5 is present; the search then found no sub-type on the 5 entry and fell through to the authenticator's, sending `auth` in the IBKey init. Which is exactly the defect the change exists to remove.

The routed type is decided first now, by the same rule `second_factor_route` uses, and only that entry's sub-type is read. An entry the gate is not using cannot supply one, and a routed entry stating none falls back to the configured value as intended. The test covers both orders of the mixed field and the case where the routed entry has no sub-type of its own.

**The doc generator carried its own copy of the wording.** `scripts/gen_api_docs.py` still described `ib_key_token_sub_type` as an account-specific setting, so running it — as the release workflow does — rewrote the corrected "fallback used only when the server states none" back to the old text. The correction would have survived exactly until the next release build. Fixed at the generator and regenerated.

830 pass plus the two `config::expiry_tests` timezone failures. `--features python` and `--examples` clean. Reverting the routed-type decision fails `auth_start_token_comes_from_field_four` by name.

userFRM:

PR #286, opened 2026-07-29
market data: key server tags rather than indexing them (ibx#281)

### LINKED ISSUE #281: market data: server_tag cap of 65536 silently drops every tick on a live session (observed tags ~274555)
## Problem

`MarketState` maps `server_tag` to `InstrumentId` through a flat array capped at 65536, and drops any tick whose tag exceeds it.

https://github.com/deepentropy/ibx/blob/main/src/engine/market_state.rs#L9

```rust
/// Max server_tag value for flat lookup array. Gateway sessions assign tags
/// monotonically and can climb well past 1024 for long-running paper accounts
/// or thin-volume / sub-penny names (which seem to consume blocks of tags).
/// 65536 covers a busy day on a single session at ~256 KB heap; tags that
/// exceed this are logged and dropped.
const MAX_SERVER_TAG: usize = 65536;
```

The cap holds on paper. It does not hold on a live session, where the tags do not start near zero. Subscribing to front-month NQ and MNQ on a live account, the very first tags assigned were:

```
server_tag 274555 exceeds MAX_SERVER_TAG (65536); ticks for instrument 0 will be dropped
server_tag 274556 exceeds MAX_SERVER_TAG (65536); ticks for instrument 1 will be dropped
server_tag 274557 exceeds MAX_SERVER_TAG (65536); ticks for instrument 2 will be dropped
server_tag 274558 exceeds MAX_SERVER_TAG (65536); ticks for instrument 3 will be dropped
```

Four consecutive tags, all 4.2x over the cap, on the first four subscriptions of the session. Nothing had a chance to climb into range — the session simply starts there.

## Impact

Every tick for every instrument is discarded at `register_server_tag`, so the engine looks healthy and delivers no data. Logon succeeds, the farms connect, `req_mkt_data` is accepted, no error surfaces on the API — the callbacks just never fire. The only signal is the warning above, which is easy to miss against a live session's log volume, and which names the cap rather than saying "you will receive no market data".

Both call sites gate on the same bound, so the write and the lookup fail together:

https://github.com/deepentropy/ibx/blob/main/src/engine/market_state.rs#L133
https://github.com/deepentropy/ibx/blob/main/src/engine/market_state.rs#L176

## Suggested direction

Raising the constant only moves the cliff — 274555 is what one session happened to assign, not a ceiling, and the tag space is the gateway's to choose.

The table is sized by the tag space, but the thing it holds is bounded by `MAX_INSTRUMENTS` (256). It is 256 KB of array for at most 256 useful entries, and the mismatch is what makes the cap load-bearing in the first place. Keying on the tag instead of indexing by it removes the bound entirely at a fraction of the memory. If the flat-array lookup is worth keeping in the hot path, a direct-mapped table that stores the tag alongside the id and falls back on collision would keep O(1) without assuming anything about the tag range.

Happy to send a PR for whichever direction you prefer.

userFRM:
## Summary

- `MarketState` mapped `server_tag` to `InstrumentId` through a flat array capped at 65536 and dropped any tick whose tag exceeded it. That cap holds on paper but not on a live session, where the tags do not start near zero — subscribing to two front-month futures on a live account, the first four tags assigned were 274555 to 274558, every one of them 4.2x over the bound.
- The consequence is silent: every tick is discarded at the lookup, so logon succeeds, the farms connect, `req_mkt_data` is accepted, no error reaches the API, and the callbacks simply never fire. The only signal was a warning naming the cap rather than saying no market data would arrive.
- Raising the constant would only move the cliff — the tag space is the gateway's to choose, and 274555 is what one session happened to assign, not a ceiling.
- The table was sized by the tag space while the thing it holds is bounded by `MAX_INSTRUMENTS`, so it spent 256 KB of array on at most 256 useful entries. That mismatch is what made the bound load-bearing. Keying on the tag rather than indexing by it removes the bound entirely at a fraction of the memory, and matches how `con_id` is already mapped in the same struct.
- `unregister` now drops an instrument's own tags instead of scanning 65536 slots for them.

Closes #281.

## Test plan

- [x] `cargo test --lib` — 805 pass. The two `config::expiry_tests` failures are pre-existing on main.
- [x] `server_tags_resolve_far_above_any_fixed_bound` uses the four tags observed on the live session and asserts they resolve, plus that `clear_server_tags` stops a reused tag resurrecting an old mapping. Reinstating the 65536 bound fails it.
- [x] `unregister_drops_only_its_own_server_tags` pins that removing one instrument leaves another's tags intact.
- [x] Live confirmation that ticks flow end to end. Subscribed a front-month CME future on a live account with the fix in place and captured 84 ticks over 45 seconds — trade prices tracking correctly, plus the daily high/low/open/close. The same subscription on the unfixed lookup delivered nothing at all: the tags assigned were 274555 and up, so every message was discarded before reaching a callback.
- [x] Live confirmation that the warning was the only symptom: the logon, the farm connections and the `req_mkt_data` all reported success in both runs. Only the tick flow differed.

### DIFF
--- a/src/engine/hot_loop/farm.rs
+++ b/src/engine/hot_loop/farm.rs
@@ -958,7 +958,16 @@ impl FarmState {
         if tick_type != 0x1E90 { return; }
 
         let server_tag = u32::from_be_bytes([body[2], body[3], body[4], body[5]]);
-        let instrument = context.market.instrument_by_server_tag(server_tag).unwrap_or(0);
+        // Instrument 0 is a real instrument — the first one registered — so an
+        // unrecognised tag must be dropped rather than defaulted, or the
+        // article is attributed to whatever was subscribed first.
+        let instrument = match context.market.instrument_by_server_tag(server_tag) {
+            Some(id) => id,
+            None => {
+                log::debug!("News for unknown server tag {server_tag}; dropping");
+                return;
+            }
+        };
 
         let batch_count = u32::from_be_bytes([body[8], body[9], body[10], body[11]]) as usize;
         let mut pos = 12;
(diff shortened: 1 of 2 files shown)


userFRM:
One correctness fix from the second review, and a test of mine that was passing for the wrong reason.

**Unknown news tags were attributed to a real instrument.** The news path fell back to instrument `0` when a tag did not resolve — and `0` is not a sentinel, it is the first instrument registered. So an article for an unrecognised tag was pinned on whatever was subscribed first. This mapping change narrows the exposure a lot, since previously *every* live-session news tick carried a tag above the old bound and hit that path, but the fallback itself was still wrong. Unknown tags are dropped now.

**The test I wrote to prove that could not fail.** It asserted no article was delivered for an unknown tag — and no article was delivered for *any* tag, because I hand-built the payload without the frame the handler unwraps first. It passed with the bug restored. It now carries a positive control: the same payload with a registered tag must deliver exactly one article, so the drop is attributable to the tag rather than to a frame the handler never accepted. With that control in place the mutation fails.

Independent confirmation of the substance: high tags register through both the subscription ack and the ticker setup, a tick updates and emits the right instrument, disconnect clears the mappings, and — the risk I most wanted checked — **a stale tag cannot resolve to the wrong instrument after slot reuse**; unregistering removes every tag belonging to that instrument and the reused slot does not resurrect the old one. The maximum 31-bit wire tag decodes correctly, and no numeric cap remains.

The hot-path cost was measured rather than estimated, pinned to one core on a release build: the full tick path goes 238ns → 270ns, about +13%, with the hashed lookup itself around 6.6ns. Left as is, per the trade the commit describes.

userFRM:
Two further sessions on the same front month, and the tag range moves each time.

| session | first tags assigned |
| --- | --- |
| earlier run above | 274555 to 274558 |
| 2026-07-29 | 110927, 110928 |
| 2026-07-31 | 223733, 223734 |

Three ranges with no relation between them beyond all sitting far above the old bound, and the third is lower than the second. Nothing here is counting up from zero within a session, so there is no constant that survives contact with the next login.

The 2026-07-31 run also shows the two tags arriving from different messages: 223733 on the `35=Q` subscription ack and 223734 on the `35=L` ticker setup, with the ticks then carried on the second. Both get registered and both are over the bound.

userFRM:
This is not confined to futures. It happens on a SMART routed SPY subscription, which makes it the default path rather than an edge case.

Same example source built twice, once at 9367845 and once on this branch, run against the same account and the same contract (SPY, con_id 756733, STK/SMART) for 45 seconds each.

At 9367845 the run prints its summary and there is nothing in it:

```
=== published tick_price ===
=== published tick_size ===
```

The only trace is three warnings naming tags 304718 and 304719.

On this branch, the same 45 seconds:

```
=== published tick_price ===
  type   1  n=6     first=744.6         last=744.56
  type   2  n=4     first=744.8         last=744.77
  type   4  n=2     first=744.69        last=744.56
  type   6  n=1     first=741.69
  type   7  n=1     first=737.68
  type  14  n=1     first=748.89
=== published tick_size ===
  type   0  n=6
  type   3  n=4
  type   8  n=2
```

The logon, the farm connections and `req_mkt_data` all reported success in both runs, and no error reached the API in either. So the released engine returns no market data for the most heavily traded ETF there is, and the only signal is a warning naming a constant.

That run also supplies a fourth tag range. Across four sessions on this account the first tag assigned was 274555, then 110927, then 223733, then 304718. All four are above the bound and the sequence does not climb, so no larger constant would have held either.

userFRM:

PR #293, opened 2026-07-29
market data: scale wire quantities into the fixed-point the readers expect (ibx#287)

### LINKED ISSUE #287: market data: sizes and volume are stored unscaled but read as QTY_SCALE fixed-point, so every quantity reaches the API 10,000x too small
## Problem

`Quote`'s quantity fields are `QTY_SCALE` fixed-point (`QTY_SCALE = 10_000`). Both readers divide by it:

https://github.com/deepentropy/ibx/blob/main/src/client_core.rs#L913

```rust
value: fields[idx] as f64 / QTY_SCALE as f64,
```

https://github.com/deepentropy/ibx/blob/main/src/python/compat/client/market_data.rs#L238

```rust
dict.set_item("bid_size", q.bid_size as f64 / qs)?;
```

The live decode path does not multiply on the way in. Prices are scaled, quantities are stored raw:

https://github.com/deepentropy/ibx/blob/main/src/engine/hot_loop/farm.rs#L232

```rust
tick_decoder::O_BID_PRICE  => { q.bid = tick.magnitude * mts; }   // scaled
...
tick_decoder::O_BID_SIZE   => { q.bid_size = tick.magnitude; }    // not scaled
tick_decoder::O_ASK_SIZE   => { q.ask_size = tick.magnitude; }
tick_decoder::O_LAST_SIZE  => { q.last_size = tick.magnitude; }
tick_decoder::O_VOLUME     => { q.volume    = tick.magnitude; }
```

So a wire quantity of `1` is stored as `1` and delivered as `1 / 10_000`.

## Observed

Streaming a front-month CME future on a live account, every `tick_size` for volume arrived as a fraction of a contract:

```
ms,tick_type,value
1627,8,0.0002
2382,8,0.0001
3138,8,0.0001
```

Tick type 8 is volume. Those are 2 contracts and 1 contract. Prices on the same subscription were correct (`tick_type 4` = 27936.75, and high/low/open/close all sane), which is consistent with the price path being scaled and the quantity path not.

## Why the tests do not catch it

`quote_ticks_dispatch` builds the `Quote` by hand with the scaling already applied, so it never exercises the decode path that omits it:

https://github.com/deepentropy/ibx/blob/main/src/api/client/tests.rs#L1761

```rust
bid_size: 1000 * QTY_SCALE as i64, ask_size: 2000 * QTY_SCALE as i64,
last_size: 500 * QTY_SCALE as i64,
volume: 10_000 * QTY_SCALE as i64,
```

and the assertions only check that a tick fired, not what value it carried:

```rust
assert!(w.events.iter().any(|e| e.starts_with("tick_size:1:8:")));   // volume
```

That combination means the producer and the consumer disagree about the units and nothing fails.

## Impact

Every size and volume delivered to both the Rust and the Python API from live market data is wrong by a factor of 10,000. Sizes are also compared against `last` for change detection, so a quantity that changes by less than one scaled unit could in principle be suppressed as well. Anything sizing an order off `last_size`, tracking volume, or reading top-of-book depth is working from wrong numbers, and nothing about the values looks obviously broken beyond being small.

## Fix

Either scale on store in the decode path, to match what the rest of the code assumes:

```rust
tick_decoder::O_BID_SIZE => { q.bid_size = tick.magnitude * QTY_SCALE; }
```

or drop the divisor in both readers and treat the fields as raw counts. The first keeps the existing field contract that `test_helpers.rs` and the Python bridge already encode, so it is the smaller change.

Worth a test that pins the producer and the reader to the same units, rather than asserting only that a tick appeared.

Correction to an earlier version of this issue: I wrote that the same shape applies to `hmds.rs`. It does not. `TbtQuote` sizes are delivered raw at `dispatch.rs` with no divisor, so that path is internally consistent and scaling it would break it.

Happy to send a PR.

userFRM:
## Summary

- `Quote`'s quantity fields are `QTY_SCALE` fixed-point (10^4). Both readers divide by it — the Rust dispatch in `client_core.rs` and the Python bridge in `market_data.rs` — but the live decode path stored the raw wire magnitude. Prices on the adjacent lines are scaled (`tick.magnitude * mts`); quantities were not.
- So every size and volume reaching either API from live market data was 10,000x too small. Streaming a front-month CME future, volume ticks arrived as `0.0001` and `0.0002` for one and two contracts, while prices on the same messages were correct — which is exactly what the split between the two paths predicts.
- The conversion is now a named function beside the constant, so producer and reader refer to the same thing. It saturates: the magnitude is server-supplied, and a clamped quantity beats a wrapped one that would surface as a plausible negative size.
- Scaled on store rather than dropping the readers' divisors, because the fixed-point form is the intended contract — order quantities are documented fixed-point, the Python test helper multiplies user units in, and both bridges divide on the way out. Removing the divisors would make fractional sizes unrepresentable.
- The tick-by-tick quote path is deliberately untouched. `TbtQuote` sizes are a different type and are delivered undivided, so that path is internally consistent; adding a divisor there breaks its existing test.

Closes #287.

## Test plan

- [x] `cargo test --lib` — 806 pass. The two `config::expiry_tests` failures are pre-existing on main.
- [x] `decoded_quantities_are_stored_as_fixed_point` pushes a real binary `35=P` body through the decode path and asserts what lands in the quote. This is the first test to reach that function — before it, reverting all four conversions left the suite bit-identical, so the defect could have returned silently. Each of the four is now individually killed by mutation.
- [x] A round-trip test pins the conversion against the reader's divisor; replacing the helper with the identity fails it with the live symptom, one contract arriving as 0.0001.
- [x] `process_msgs_dispatches_all_quote_fields` asserts the delivered quantities. It previously asserted only that a tick appeared, which is a large part of why this shipped.
- [x] A price test pins that the quantity change did not disturb the scaling that was already correct.
- [ ] Not covered: the Python bridge is feature-gated and outside a default `cargo test --lib`, and the two benchmark binaries carry their own copies of the apply loop (updated here, but not reached by the test run).

## Note

`tick_decoder::VOLUME_MULT` is dead and documents "IB encodes volume * 10000", which would mean volume must not be scaled again. It does not describe this field: across a live capture the magnitudes were single digits and fell as often as they rose, so whatever `O_VOLUME` carries is neither pre-scaled nor cumulative. That it does not look like cumulative volume at all is filed separately as ibx#290.

### DIFF
--- a/src/api/client/tests.rs
+++ b/src/api/client/tests.rs
@@ -1782,11 +1782,19 @@ fn process_msgs_dispatches_all_quote_fields() {
     assert!(w.events.iter().any(|e| e.starts_with("tick_price:1:7:")));   // low
     assert!(w.events.iter().any(|e| e.starts_with("tick_price:1:9:")));   // close
     assert!(w.events.iter().any(|e| e.starts_with("tick_price:1:14:"))); // open
-    // tick_size for: bid_size(0), ask_size(3), last_size(5), volume(8)
-    assert!(w.events.iter().any(|e| e.starts_with("tick_size:1:0:")));   // bid_size
-    assert!(w.events.iter().any(|e| e.starts_with("tick_size:1:3:")));   // ask_size
-    assert!(w.events.iter().any(|e| e.starts_with("tick_size:1:5:")));   // last_size
-    assert!(w.events.iter().any(|e| e.starts_with("tick_size:1:8:")));   // volume
+    // tick_size for: bid_size(0), ask_size(3), last_size(5), volume(8).
+    // Assert the delivered quantity, not just that a tick appeared — the
+    // scaling defect in ibx#287 fired every one of these with a value four
+    // orders of magnitude off, and a `starts_with` check passed throughout.
+    let delivered = |prefix: &str| -> Option<f64> {
+        w.events.iter().find(|e| e.starts_with(prefix))
+            .and_then(|e| e.rsplit(':').next())
+            .and_then(|v| v.parse().ok())
+    };
+    assert_eq!(delivered("tick_size:1:0:"), Some(1000.0), "bid_size");
+    assert_eq!(delivered("tick_size:1:3:"), Some(2000.0), "ask_size");
+    assert_eq!(delivered("tick_size:1:5:"), Some(500.0), "last_size");
+    assert_eq!(delivered("tick_size:1:8:"), Some(10_000.0), "volume");
 }
 
 #[test]
--- a/src/bin/bench_decode.rs
+++ b/src/bin/bench_decode.rs
@@ -295,10 +295,10 @@ fn apply_tick(q: &mut Quote, tick: &RawTick, min_tick_scaled: i64) {
         tick_decoder::O_LOW_PRICE => q.low = tick.magnitude * min_tick_scaled,
         tick_decoder::O_OPEN_PRICE => q.open = tick.magnitude * min_tick_scaled,
         tick_decoder::O_CLOSE_PRICE => q.close = tick.magnitude * min_tick_scaled,
-        tick_decoder::O_BID_SIZE => q.bid_size = tick.magnitude,
-        tick_decoder::O_ASK_SIZE => q.ask_size = tick.magnitude,
-        tick_decoder::O_LAST_SIZE => q.last_size = tick.magnitude,
-        tick_decoder::O_VOLUME => q.volume = tick.magnitude,
+        tick_decoder::O_BID_SIZE => q.bid_size = ibx::types::qty_from_wire(tick.magnitude),
+        tick_decoder::O_ASK_SIZE => q.ask_size = ibx::types::qty_from_wire(tick.magnitude),
+        tick_decoder::O_LAST_SIZE => q.last_size = ibx::types::qty_from_wire(tick.magnitude),
+        tick_decoder::O_VOLUME => q.volume = ibx::types::qty_from_wire(tick.magnitude),
         tick_decoder::O_TIMESTAMP | tick_decoder::O_LAST_TS => {
             q.timestamp_ns = tick.magnitude as u64;
         }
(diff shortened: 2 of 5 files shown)


userFRM:
Second review is clean on the change itself, with the scaling verified rather than asserted.

An arm-by-arm table confirms every quantity site is converted and no price site is: a whole-`Quote` probe using distinct price and quantity factors showed each production arm changing only its intended field. Overflow behaves — a four-byte magnitude scales to about 2.1e13, well inside the target type, and extended widths saturate rather than wrap. Every one-to-four-byte value round-trips exactly.

One gap it found on this commit in isolation: quantity-scaling the other six price arms survived, because only `bid` had a regression test here. That is closed by the commit stacked on top of this one, which pins where each decoded type lands — I verified by mutating five price arms to the quantity scale and it fails there by name. Worth knowing if these are considered separately.

Also confirmed: the commit is exactly its message, five files, no unrelated hunks, and no leftover debug hook anywhere in the tree.

userFRM:
Reproduced on an equity, where the right answer is not in doubt.

Streaming SPY with the decode path traced, these are the bid size magnitudes on the wire and what the API delivers for them:

```
wire       40      80    8040    8120    8320
API     0.004   0.008   0.804   0.812   0.832
```

A 0.832 share bid on SPY is not a quantity any venue quotes. 8320 shares is. Ask sizes and volume from the same session land the same way, and the prices carried on those messages are correct, which is the split between the two paths.

There is also a control for this inside the tree, so it does not rest on judging what a plausible size looks like. Depth sizes arrive on the same connection and `parse_depth_fields` reads them straight to `f64`, with no fixed point step in either direction, and `dispatch.rs` hands that number to the wrapper unchanged. For the same book at the same moment, depth reports 8320 and L1 reports 0.832. Whichever number is right, the two cannot both be, and they differ by exactly the divisor the readers apply.

userFRM:

PR #294, opened 2026-07-29
market data: re-subscribe L1 from the record that survives a disconnect (ibx#288)

### LINKED ISSUE #288: market data: a farm reconnect re-subscribes zero L1 instruments, because the list it iterates was cleared on disconnect
## Problem

`handle_disconnect` clears the subscription list:

https://github.com/deepentropy/ibx/blob/main/src/engine/hot_loop/farm.rs#L886

```rust
self.md_req_to_instrument.clear();
self.instrument_md_reqs.clear();
// Clear depth wire-state (server_tags become invalid after disconnect).
// depth_resub_info is preserved for resubscription on reconnect.
```

`reconnect` then builds its resubscribe set by iterating that same list:

https://github.com/deepentropy/ibx/blob/main/src/engine/hot_loop/farm.rs#L917

```rust
let active: Vec<(InstrumentId, i64, String, ...)> = self.instrument_md_reqs.iter()
    .filter_map(|(id, _)| { ... })
    .collect();
```

By then it is empty, so `active` is empty and the loop below issues nothing.

The data needed to rebuild the subscriptions did survive — `md_resub_info` is populated on subscribe and pruned on unsubscribe, exactly like `depth_resub_info`. It is read only to decorate entries that `active` no longer contains, and then discarded:

https://github.com/deepentropy/ibx/blob/main/src/engine/hot_loop/farm.rs#L930

```rust
let old_resub = std::mem::take(&mut self.md_resub_info);
for (instrument, con_id, sym, ...) in active {   // empty
    self.send_mktdata_subscribe(...);
}
drop(old_resub);
```

Depth takes the other route in the same function and works: `depth_resub_info` is preserved on disconnect and iterated directly on reconnect. L1 preserves the equivalent record and then iterates the cleared list instead.

## Impact

After any real farm drop, the connection comes back healthy and delivers no L1 market data for the rest of the session. No `35=Q` is sent, so no acks, no server tags, no ticks. Nothing errors — the reconnect reports success, `is_connected()` stays true, and the callbacks simply stop. Depth subscriptions keep working, which makes it look like a data-vendor problem rather than a client one.

## Why the test suite does not catch it

The only reconnect test drives the test-only hook, which sets the flag without clearing anything:

https://github.com/deepentropy/ibx/blob/main/src/engine/hot_loop/farm.rs#L899

```rust
/// Test-only: set disconnected without clearing state or emitting events.
pub fn handle_disconnect_for_test(&mut self) {
    self.disconnected = true;
}
```

So the test reconnects with `instrument_md_reqs` still populated and resubscribes correctly. The production path is the one that clears it, and nothing exercises that.

## Fix

Drive the resubscribe from `md_resub_info`, the record that is meant to survive, the way the depth path already does. `con_id` is still reachable per instrument through `market.con_id`.

A test that goes through `handle_disconnect` rather than the test hook would pin it. Worth checking the log line after a reconnect: re-subscribing zero instruments should be loud rather than silent.

## Note

Related but separate, in the same area: `send_mktdata_unsubscribe` does not purge `md_req_to_instrument`, so an in-flight `35=Q` ack arriving after an unsubscribe-then-reclaim can map its tag and the previous contract's `min_tick` onto a reused instrument slot. Narrow window, roughly one round trip, and not caused by the above — mentioning it here rather than opening a second issue since a fix in this area would likely touch both.

Happy to send a PR.

### LINKED ISSUE #291: market data: unsubscribing tick-by-tick or news reclaims the instrument slot even when a live L1 subscription is still using it
## Problem

`try_reclaim_instrument` guards against three kinds of live reference before freeing a slot:

https://github.com/deepentropy/ibx/blob/main/src/engine/hot_loop/mod.rs#L258

```rust
fn try_reclaim_instrument(&mut self, instrument: InstrumentId) {
    if !self.context.open_orders_for(instrument).is_empty() {
        return;
    }
    if self.hmds.tbt_subscriptions.iter().any(|(id, _, _)| *id == instrument) {
        return;
    }
    if self.ccp.news_subscriptions.iter().any(|(id, _)| *id == instrument) {
        return;
    }
    if self.context.market.unregister(instrument).is_some() {
```

A live L1 market-data subscription is not among them, and it is a reference just as much as the other three. It is called from three places, two of which have nothing to do with market data:

https://github.com/deepentropy/ibx/blob/main/src/engine/hot_loop/mod.rs#L419

```rust
ControlCommand::UnsubscribeTbt { instrument } => {
    ...
    self.try_reclaim_instrument(instrument);
```

https://github.com/deepentropy/ibx/blob/main/src/engine/hot_loop/mod.rs#L431

```rust
ControlCommand::UnsubscribeNews { instrument } => {
    ...
    self.try_reclaim_instrument(instrument);
```

So: subscribe L1 on a contract, also subscribe tick-by-tick or news on it, then drop only the tick-by-tick or news subscription. All three guards pass, and the slot is freed while `farm.instrument_md_reqs` still holds an active L1 request for it.

## Impact

`unregister` drops the instrument's server tags, so the L1 subscription the caller still holds stops delivering. Nothing errors — the gateway keeps sending, the tags no longer resolve, and the ticks are discarded. The caller sees a subscription that is still registered on its side and simply goes quiet.

The freed slot is then reusable, so a later `register` can hand that id to a different contract while the original L1 subscription is still live on the wire, and the stale `md_resub_info` entry for it can be replayed on the next reconnect against whatever now occupies the slot.

The `#233` caller contract already states the invariant — nothing may still reference the id — and an active market-data request is exactly such a reference. This is the one referent the check does not cover.

## Fix

A fourth guard in the same shape as the existing three, checking `instrument_md_reqs` for the instrument. The `Unsubscribe` path at line 410 already removes that entry before calling, so it is unaffected; only the tick-by-tick and news paths change behaviour, which is the point.

userFRM:
## Summary

- `handle_disconnect` clears `instrument_md_reqs`, and `reconnect` selected the subscriptions to re-issue by iterating that same list. By then it is empty, so a farm reconnect re-subscribed nothing: no `35=Q` went out, no acks came back, no server tags were assigned, and no ticks arrived for the rest of the session.
- The failure is silent. The reconnect reports success, `is_connected()` stays true, and the callbacks simply stop. Depth keeps working, which makes it look like a data problem rather than a client one.
- The record needed to rebuild the subscriptions already survived. `md_resub_info` is written on subscribe and pruned on unsubscribe, exactly as `depth_resub_info` is — and the depth path a few lines below already re-subscribes by draining its own record. The L1 path read `md_resub_info` only to decorate entries the empty list never produced, then dropped it unused.
- Selection now drains `md_resub_info` the same way, skipping any instrument whose slot was reclaimed while the farm was down. It is a named method so it is reachable without a live `Connection`; the previous shape was only reachable through `reconnect`, which is a large part of why this survived.
- The unsubscribe prune is hoisted above the lookup that preceded it. That lookup early-returns when an instrument has no active requests, which is always true while the farm is down — so an unsubscribe issued during an outage left the record standing, and now that reconnect replays the record it would come back as a subscription the caller had explicitly cancelled, or re-subscribe whatever contract had since taken the slot using the cancelled one's parameters.

Closes #288.
Closes #291.

## Test plan

- [x] `cargo test --lib` — 805 pass. The two `config::expiry_tests` failures are pre-existing on main.
- [x] The tests drive the real `handle_disconnect` rather than `handle_disconnect_for_test`. That hook sets the disconnected flag without clearing anything, so the existing reconnect test re-subscribed from a list the production path would have emptied — nothing exercised the real disconnect, which is why this was invisible. The real one needs only state, no socket.
- [x] Selecting from the cleared list again fails the first test at 0 targets against 1, reproducing the defect.
- [x] Putting the prune back below the early return fails the unsubscribe-while-down test.
- [x] Dropping the `con_id` filter fails the reclaimed-slot test.
- [x] Re-issuing with no connection still leaves the record standing, so a later reconnect can retry rather than losing the subscription.
- [ ] Not covered: `reconnect()` itself, its clear ordering, and the exact wire fan-out. Replacing the production selection with an empty vector still passes these tests, because they exercise the selection helper rather than the caller. Closing that needs a listener socket and decoding the emitted `35=V`, which is integration scaffolding this PR does not add.

## Note

This does not restore live market data on its own. A reconnect will now re-subscribe correctly, but the resulting acks carry the high server tags that ibx#281 drops — both are needed for ticks to actually flow after a farm drop.

### DIFF
--- a/src/engine/hot_loop/farm.rs
+++ b/src/engine/hot_loop/farm.rs
@@ -32,6 +32,19 @@ pub(crate) struct FarmState {
 }
 
 impl FarmState {
+    /// Whether this instrument still has market-data state that would be
+    /// repointed by a slot reuse: a live subscription, a request id waiting on
+    /// its ack, or a record kept for the next reconnect.
+    ///
+    /// The resubscribe record is the reason this matters. Before it existed a
+    /// reclaimed slot replayed nothing; now it replays the old contract's
+    /// descriptor against whatever contract holds the id at reconnect (ibx#288).
+    pub(crate) fn holds_market_data(&self, instrument: InstrumentId) -> bool {
+        self.instrument_md_reqs.iter().any(|(id, _)| *id == instrument)
+            || self.md_req_to_instrument.iter().any(|(_, id)| *id == instrument)
+            || self.md_resub_info.iter().any(|r| r.0 == instrument)
+    }
+
     pub(crate) fn new() -> Self {
         Self {
             next_md_req_id: 1,
@@ -1111,5 +1124,39 @@ mod resub_tests {
         market.unregister(instrument);
 
         assert!(farm.take_resub_targets(&market).is_empty());
+    }
+
+    /// The case the test above does not reach: the slot is not merely freed but
+    /// handed to another contract before the reconnect. `md_resub_info` holds
+    /// no con_id of its own, so the record is combined with whatever con_id the
+    /// id now resolves to — the old contract's descriptor subscribing the new
+    /// contract's instrument. The guard is that a slot holding market-data
+    /// state is not reclaimable in the first place.
+    #[test]
+    fn an_instrument_holding_a_resubscribe_record_is_not_reclaimable() {
+        let mut farm = FarmState::new();
+        let mut market = MarketState::new();
+        let instrument = market.register(756733);
+        farm.md_resub_info.push((
+            instrument, "SPY".into(), "SMART".into(), "STK".into(), String::new(),
+            0.0, String::new(), String::new(), 0,
+        ));
+
+        assert!(
+            farm.holds_market_data(instrument),
+            "the record alone must keep the slot resident",
+        );
+
+        // And each of the other two references does the same on its own.
+        let mut farm = FarmState::new();
+        farm.instrument_md_reqs.push((instrument, vec![7]));
+        assert!(farm.holds_market_data(instrument), "a live subscription");
+
+        let mut farm = FarmState::new();
+        farm.md_req_to_instrument.push((7, instrument));
+        assert!(farm.holds_market_data(instrument), "a request awaiting its ack");
+
+        // An instrument with none of the three is free to go.
+        assert!(!FarmState::new().holds_market_data(instrument));
     }
 }
--- a/src/engine/hot_loop/farm.rs
+++ b/src/engine/hot_loop/farm.rs
@@ -462,6 +462,13 @@ impl FarmState {
         farm_conn: &mut Option<Connection>,
         hb: &mut HeartbeatState,
     ) {
+        // Drop the resubscribe record first. The lookup below early-returns
+        // when the instrument has no active requests, which is always the case
+        // while the farm is down — `handle_disconnect` cleared that list — so
+        // an unsubscribe issued during an outage would otherwise leave the
+        // record standing and the reconnect would re-subscribe an instrument
+        // the caller explicitly cancelled (ibx#288).
+        self.md_resub_info.retain(|(id, ..)| *id != instrument);
         let reqs = match self.instrument_md_reqs.iter()
             .position(|(id, _)| *id == instrument)
         {
@@ -471,7 +478,6 @@ impl FarmState {
             }
             None => return,
         };
-        self.md_resub_info.retain(|(id, ..)| *id != instrument);
 
         let conn = match farm_conn.as_mut() {
             Some(c) => c,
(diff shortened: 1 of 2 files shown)


userFRM:
The same defect shows from two directions — from this branch and from #298 — and it is the most serious thing this round turned up. Fixed in `941b30b`.

**Replaying a resubscribe record made a hole in the reclaim guard reachable.** That guard keeps an instrument slot resident while open orders, tick-by-tick or news still reference it, precisely because a reused id repoints those references at the wrong contract. Market data was not on the list.

So: subscribe L1 and tick-by-tick for one contract — they share a slot, since registration dedupes by conId — then cancel the tick-by-tick. The slot is freed underneath the live L1 subscription and handed to the next contract. `md_resub_info` carries no conId of its own, so on reconnect the old contract's descriptor is combined with whatever conId the id now resolves to, and the wrong instrument is subscribed. Prices arrive, they are plausible, and they are for something else.

Before this branch a reclaimed slot replayed nothing, which is why the gap only becomes reachable here. That makes it this PR's to close rather than a pre-existing one to file.

The fix is to complete the guard rather than to validate at replay time: a slot holding a live subscription, a request still awaiting its ack, or a resubscribe record is not reclaimable. Each of the three keeps it resident on its own, and each is pinned separately — dropping any one arm fails the test by name.

The existing test covered a slot freed and left free. The case that costs money is a slot freed and handed on, which it could not reach.

**One gap I am leaving named.** Reverting `reconnect` itself to the old selection still fails nothing: all three tests call `take_resub_targets` directly and none drives `reconnect`, which needs a live socket. The commit text said only the wire write was uncovered; the call site is too. If the loopback-pair helper from the 2FA branch lands, closing that is small — but it is not this branch's to invent.

806 pass plus the two `config::expiry_tests` timezone failures. `--features python` clean.

userFRM:

PR #298, opened 2026-07-29
market data: forget a request id when its subscription is cancelled (ibx#289)

### LINKED ISSUE #289: market data: an unsubscribe leaves its request id mapped, so a late 35=Q ack binds the old subscription's server_tag and minTick onto a reused instrument slot
## Problem

`send_mktdata_unsubscribe` removes the instrument's entry from `instrument_md_reqs` and prunes `md_resub_info`, but it never removes the request ids from `md_req_to_instrument`:

https://github.com/deepentropy/ibx/blob/main/src/engine/hot_loop/farm.rs#L465

```rust
let reqs = match self.instrument_md_reqs.iter()
    .position(|(id, _)| *id == instrument)
{
    Some(idx) => {
        let (_, reqs) = self.instrument_md_reqs.remove(idx);
        reqs
    }
    None => return,
};
self.md_resub_info.retain(|(id, ..)| *id != instrument);
```

That map is what the `35=Q` ack handler resolves against, and it is only otherwise emptied on disconnect or reconnect:

https://github.com/deepentropy/ibx/blob/main/src/engine/hot_loop/farm.rs#L289

```rust
let instrument = match self.md_req_to_instrument.iter()
    .position(|(id, _)| *id == req_id)
{
    Some(idx) => {
        let (_, instr) = self.md_req_to_instrument.remove(idx);
        instr
    }
    None => return,
};

context.market.register_server_tag(server_tag, instrument);
context.market.set_min_tick(instrument, min_tick);
```

So: subscribe, unsubscribe before the ack lands, let the slot be reclaimed and reused by a different contract, then the late ack arrives. It still resolves the stale request id, and applies that subscription's `server_tag` **and its `minTick`** to whatever contract now occupies the slot.

## Impact

The tag mix-up alone would route one contract's ticks to another. The `minTick` is worse: prices are scaled by it on the decode path (`tick.magnitude * min_tick_scaled`), so the reused instrument's prices are then computed with the previous contract's tick size. A future whose minTick differs from the stock that replaced it does not produce an obviously broken number — it produces a plausible, wrong price, silently, for as long as the subscription lives.

The window is roughly one round trip, so it needs an unsubscribe that races an in-flight ack, then a reclaim and reuse of that slot. Not common, but not exotic either for anything that rotates symbols — a scanner, or a strategy cycling watchlists while holding positions.

## Fix

Drop the cancelled request ids from `md_req_to_instrument` in `send_mktdata_unsubscribe`, alongside the `instrument_md_reqs` removal that already happens there. The ids are in hand at that point — they are the `reqs` vector the function goes on to send cancels for.

A late ack for an unknown request id then hits the existing `None => return` and is ignored, which is the correct outcome.

Worth a test that acks a request id after its unsubscribe and asserts nothing is registered against the instrument.

userFRM:
## Summary

- `send_mktdata_unsubscribe` removed the instrument from `instrument_md_reqs` and pruned `md_resub_info`, but left its request ids in `md_req_to_instrument` — the map the `35=Q` ack handler resolves against, which nothing else empties short of a disconnect.
- So an ack still in flight when the unsubscribe went out resolved afterwards. If the slot had been reclaimed and reused by then, the ack applied that subscription's `server_tag` **and its `minTick`** to whatever contract now held the slot.
- The tag alone misroutes ticks. The `minTick` is worse: prices are scaled by it on the decode path, so the reused instrument's prices are computed with the previous contract's tick size — a plausible wrong price rather than an obvious fault, for as long as the subscription lives.
- The ids are already in hand there; they are the ones the function goes on to send cancels for. A later ack for an unknown request now takes the existing `None => return`.

Closes #289.

## Test plan

- [x] `cargo test --lib` — 803 pass. The two `config::expiry_tests` failures are pre-existing on main.
- [x] `a_late_ack_for_an_unsubscribed_request_is_ignored` subscribes, records the pending request ids, unsubscribes, and asserts none still resolves. Dropping the prune fails it.
- [ ] Not covered: the race itself needs an ack arriving between the unsubscribe and a slot reuse, which needs a live connection. The test pins the state that makes the race possible, not the timing.

### DIFF
--- a/src/engine/hot_loop/farm.rs
+++ b/src/engine/hot_loop/farm.rs
@@ -472,6 +472,13 @@ impl FarmState {
             None => return,
         };
         self.md_resub_info.retain(|(id, ..)| *id != instrument);
+        // Forget the pending acks too. A `35=Q` still in flight when the
+        // unsubscribe goes out would otherwise resolve its request id after
+        // the slot has been reclaimed and reused, binding this subscription's
+        // server_tag AND its minTick onto whatever contract now holds the
+        // slot — prices for the new contract then scale by the old one's tick
+        // size, which reads as plausible rather than broken (ibx#289).
+        self.md_req_to_instrument.retain(|(req_id, _)| !reqs.contains(req_id));
 
         let conn = match farm_conn.as_mut() {
             Some(c) => c,
@@ -1015,3 +1022,38 @@ impl FarmState {
     }
 }
 
+
+#[cfg(test)]
+mod stale_ack_tests {
+    use super::*;
+    use crate::engine::context::Context;
+
+    /// A `35=Q` in flight when the unsubscribe goes out used to resolve its
+    /// request id afterwards. If the slot had been reclaimed and reused by
+    /// then, the ack bound its server_tag and minTick onto the new contract,
+    /// whose prices then scaled by the previous contract's tick size — a
+    /// plausible wrong price rather than an obvious fault (ibx#289).
+    #[test]
+    fn a_late_ack_for_an_unsubscribed_request_is_ignored() {
+        let mut farm = FarmState::new();
+        let mut context = Context::new();
+        let mut hb = HeartbeatState::new();
+        let instrument = context.market.register(756733);
+
+        farm.send_mktdata_subscribe(
+            756733, "SPY", "SMART", "STK", "", 0.0, "", "", instrument, 0,
+            &mut None, &mut hb,
+        );
+        let pending: Vec<u32> = farm.md_req_to_instrument.iter().map(|(r, _)| *r).collect();
+        assert!(!pending.is_empty(), "the subscribe must register at least one request");
+
+        farm.send_mktdata_unsubscribe(instrument, &mut None, &mut hb);
+
+        for req_id in pending {
+            assert!(
+                !farm.md_req_to_instrument.iter().any(|(r, _)| *r == req_id),
+                "request {} must not resolve after its unsubscribe", req_id,
+            );
+        }
+    }
+}


userFRM:
Reviewed: merge. The claim holds by trace and the one new test discriminates.

The prune runs with the exact id list it then cancels, and sits ahead of the early return for a downed connection, so it fires even when the farm is already gone. Every route reaches it — explicit cancel and the shutdown unsubscribe-all — the ack handler self-removes on resolve, disconnect clears the map, and reconnect clears before re-subscribing. Client request ids are monotonic and never reset, so a forgotten id cannot be re-issued and land on a stale mapping.

Reverting the prune fails `a_late_ack_for_an_unsubscribed_request_is_ignored` by name. The test is not vacuous: it asserts the pending ids are non-empty before unsubscribing, so it would catch a prune that worked by there being nothing to prune.

One thing this pass found that is **not** this PR's, but is worth recording here because it was found from this branch: the same misattribution survives through slot reclamation. Cancelling tick-by-tick or news on an instrument that also holds a live L1 subscription frees the slot underneath it, because the reclaim guard checks orders, tick-by-tick and news but not market data. That is fixed in #294, where replaying a resubscribe record is what makes it reachable — until then a reclaimed slot replayed nothing.

803 pass plus the two `config::expiry_tests` timezone failures. One production hunk and one test module, both accounted for by the message.

userFRM:
No further changes needed here. Independently confirmed the claimed stale request-id binding is fixed and that the regression test kills the reverted mutant by name, with no scope creep or leftover instrumentation in the diff. Nothing further to change here.

userFRM:

PR #299, opened 2026-07-29
ccp: treat a missing position-feed quantity as no quantity, not as flat (ibx#296)

### LINKED ISSUE #296: ccp: the 6040=75 position feed still reads a missing or unparseable quantity as flat
## Problem

#261 covers the account-update path, where an absent tag 6064 was read as a position of zero. The sibling handler for the `6040=75` position feed has the same shape and was not part of that fix.

The running quantity starts at zero:

https://github.com/deepentropy/ibx/blob/main/src/engine/hot_loop/ccp.rs#L1799

```rust
let mut qty: i64 = 0;
```

and an unparseable value falls back to zero rather than being rejected:

https://github.com/deepentropy/ibx/blob/main/src/engine/hot_loop/ccp.rs#L1821

```rust
} else if let Some(v) = part.strip_prefix("6064=") {
    qty = v.parse::<f64>().unwrap_or(0.0) as i64;
```

Both flush sites then publish that value unconditionally — to the position row, to the shared position, and to the event:

https://github.com/deepentropy/ibx/blob/main/src/engine/hot_loop/ccp.rs#L1807

```rust
shared.portfolio.set_position_info(PositionInfo {
    con_id, position: qty, avg_cost, ..Default::default()
});
...
shared.portfolio.set_position(instrument, qty);
emit(event_tx, Event::PositionUpdate { instrument, con_id, position: qty, .. });
```

So a group entry that carries a conId but no parseable 6064 flattens a live position, publishes it to `reqPositions` and both P&L paths, and emits a `PositionUpdate` saying the position is zero.

## Impact

Same as #261: a strategy reading its own position sees flat when it is not, and can open a duplicate. The value is a plausible zero rather than an obvious fault, so nothing downstream can tell it apart from a genuine flat.

The repeating group here is structurally `(conId, qty, avgCost)`, so an entirely absent quantity is less likely than on the account-update path. The unparseable case needs no protocol anomaly at all — one bad byte reaches `unwrap_or(0.0)`. Note also that `"NaN".parse::<f64>()` succeeds and `NaN as i64` is 0, so a non-finite value takes the same route without failing the parse.

## Fix

Mirror #261: parse the quantity per entry as an `Option<i64>`, filtered to finite values, and skip the flush for an entry that does not carry one. Roughly four lines across the two flush sites. A genuine flat still arrives as an explicit `6064=0` and must keep flattening.

Found while auditing #264, which fixes the account-update path only.

userFRM:
## Summary

- The `6040=75` position feed defaults its running quantity to zero and falls back to zero on an unparseable value, then both flush sites publish it unconditionally — position row, shared position, and a `PositionUpdate`.
- So an entry carrying a conId but no parseable `6064` flattened a live position and published it to `reqPositions` and both P&L paths. A strategy reading its own position sees flat when it is not, and can open a duplicate.
- Same defect ibx#261 fixed on the account-update path; that fix did not reach this sibling handler. The unparseable case needs no protocol anomaly — one bad byte reaches `unwrap_or(0.0)`, and `"NaN".parse::<f64>()` succeeds while `NaN as i64` is zero, so a non-finite value took the same route without failing the parse.
- The running quantity is now an `Option`, filtered to finite values; an entry without one is skipped rather than flushed. A genuine flat still arrives as an explicit `6064=0` and still flattens.

Closes #296.

## Test plan

- [x] `cargo test --lib` — 804 pass. The two `config::expiry_tests` failures are pre-existing on main.
- [x] Three shapes that must leave a live position alone: absent tag, unparseable value, non-finite value.
- [x] One that must still flatten: an explicit `6064=0`. Restoring the zero default fails the first three and leaves this one passing, so the fix cannot be mistaken for "ignore zeros".

### DIFF
--- a/src/client_core.rs
+++ b/src/client_core.rs
@@ -983,6 +983,7 @@ impl ClientCore {
         let mut total_unrealized: f64 = 0.0;
         let mut total_realized: f64 = 0.0;
         let mut priced = 0usize;
+        let mut unpriceable = 0usize;
 
         for con_id in con_ids {
             let seed = seeds.get(&con_id);
@@ -993,10 +994,12 @@ impl ClientCore {
             total_realized += seed.map(|s| s.realized_pnl).unwrap_or(0.0);
 
             // A position we held at midnight and cannot currently size is not
-            // a flat one. Pricing the absence as zero reports the whole
-            // overnight holding as sold; skipping leaves the gateway's own
-            // account-level figure to stand (ibx#296).
+            // a flat one: pricing the absence as zero reports the whole
+            // overnight holding as sold. It is counted as unpriceable rather
+            // than merely skipped, because a total missing one position is not
+            // a smaller correct answer (ibx#296).
             if seed.is_some() && pi.is_none() {
+                unpriceable += 1;
                 continue;
             }
             let qty_now = pi.as_ref().map(|p| p.position).unwrap_or(0);
@@ -1005,6 +1008,7 @@ impl ClientCore {
             // parse says the position is not intraday, only that its size is
             // unknown, so there is nothing to price it against.
             let Some(qty_midnight) = seed.map_or(Some(0), |s| s.qty_midnight) else {
+                unpriceable += 1;
                 continue;
             };
 
@@ -1046,7 +1050,13 @@ impl ClientCore {
         // to the gateway's account-level P&L, which the gateway pushes independently
         // of any market-data subscription. Without this the quote-derived totals stay
         // [0,0,0] and no callback ever fires (ibx#239).
-        if priced == 0 {
+        // A position that could not be priced makes the client-side sum an
+        // incomplete account total, not a smaller correct one — and the realized
+        // figure has already accrued for it, so the three would not even agree
+        // with each other. The gateway's own account-level numbers are complete
+        // by construction, so one unpriceable position sends the whole account
+        // to them rather than reporting a partial sum as if it were the total.
+        if priced == 0 || unpriceable > 0 {
             let acct = shared.portfolio.account();
             total_daily = acct.daily_pnl as f64 / PRICE_SCALE_F;
             total_unrealized = acct.unrealized_pnl as f64 / PRICE_SCALE_F;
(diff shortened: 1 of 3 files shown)


userFRM:
Updated in `a9e7610`. Two corrections to how the skip interacts with the totals.

**A skipped position now sends the whole account to the gateway figure.** The account-level fallback only ran when *nothing* priced, so a portfolio holding one ordinary position and one that could not be sized reported the ordinary one's P&L as the account total — a partial sum presented as complete. Realized had already accrued for the skipped position, so the three totals did not agree with each other either. A total missing a position is not a smaller correct total, so one unpriceable position now defers the whole account to the gateway's own numbers, which are complete by construction.

**`pnlSingle` keeps reporting what it knows.** There the unknown costs only the daily figure: position, value, unrealized and realized are all still available. Suppressing the callback left every one of them stale on the caller's side — a quote could move and nothing would be reported. It now emits the known fields and holds the daily figure at its last reported value, rather than recomputing it from an assumption that is wrong in a specific direction.

## On the public field type

`MidnightSeed.qty_midnight` changing from `i64` to `Option<i64>` is source-breaking for anyone constructing or reading that struct. Representing "the row arrived and stated no quantity" needs somewhere to put it, and adding a field to a struct with public fields breaks literal construction just the same — so the break is not avoidable, only relocatable. Flagging it rather than burying it; happy to take a different shape if you prefer one.

## Note on one test

`a_position_feed_entry_with_a_quantity_publishes_it_everywhere` passes with the fix reverted by design. It is the positive control, and it exists so the absence assertions beside it cannot pass against a feed that publishes nothing at all.

810 pass plus the two `config::expiry_tests` timezone failures. All five offline targets clean. Reverting the whole-account fallback fails `one_unpriceable_position_sends_the_whole_account_to_the_gateway`; reverting the single-callback change fails `an_unknown_seed_does_not_suppress_the_rest_of_a_single_callback`. Both by name.

userFRM:
It rejected my fix, and it was right on both counts. Corrected in `a9e7610`.

**The fallback I claimed does not exist for a mixed portfolio.** I wrote that skipping an unpriceable position "leaves the gateway's own account-level figure to stand". That fallback only runs when *nothing* prices. So an account with one ordinary position and one that could not be sized reported the ordinary one's P&L **as the account total** — a partial sum presented as complete, which is a quieter version of the wrong number I was removing. Worse, realized had already accrued for the skipped position, so the three totals did not agree with each other.

A total missing a position is not a smaller correct total. One unpriceable position now sends the whole account to the gateway's numbers, which are complete by construction.

**And `pnlSingle` had the opposite problem.** There the unknown costs only the daily figure — position, value, unrealized and realized are all still known. My `continue` suppressed the entire callback, so a quote could move and the caller would be told nothing, leaving every field stale. It now reports what it knows and holds the daily figure at its last reported value, rather than recomputing it from an assumption that is wrong in a specific direction.

**One finding I am not acting on, with reasons.** `MidnightSeed.qty_midnight` changing from `i64` to `Option<i64>` is source-breaking for anyone constructing or reading that struct, and it is public and re-exported. But representing "the row arrived and stated no quantity" needs somewhere to put it, and adding a field to a struct with public fields breaks literal construction just the same. So the break is not avoidable, only relocatable — flagging it here rather than hiding it, and happy to take a different shape if you would rather.

Also noted: `a_position_feed_entry_with_a_quantity_publishes_it_everywhere` passes with the fix reverted. That is deliberate — it is the positive control, and it exists precisely so the absence assertions beside it cannot pass against a feed that publishes nothing at all.

810 pass plus the two `config::expiry_tests` timezone failures. All five offline targets clean. Mutation-verified: reporting a partial total fails `one_unpriceable_position_sends_the_whole_account_to_the_gateway`, and suppressing the single callback fails `an_unknown_seed_does_not_suppress_the_rest_of_a_single_callback`, both by name.

userFRM:

PR #300, opened 2026-07-29
ci: run the test suite on push and pull request

### LINKED ISSUE #301: tests: the integration targets did not compile, and nine fail once they do; the Python suite has 23 failures and is never compiled
## Why this exists

#300 adds a test workflow and had to fix the test targets to compile at all. Doing that revealed what had been hidden behind the compile error. Filing it separately so the workflow can land green rather than red from its first run.

## The integration targets did not compile

Three struct changes left them behind: `ControlCommand::FetchContractDetails` gained `filters`, and `OrderRequest`'s trailing-stop and adjustable-stop variants gained `trail_stop_price` and the adjustable trailing pair. Thirteen literals across `tests/ib_paper_compat` and `tests/scenarios`. Fixed in #300, since no workflow can exist without it.

Because they never compiled, they never ran — so the failures below have been invisible rather than absent.

## Nine failures once they compile

`cargo test --no-fail-fast` gives 911 passed, 11 failed, 8 ignored. Two of the eleven are the tzdata aliases described in #300 and pass on a runner with the full database. The other nine:

- `coverage::control_command_manifest_tracks_all_variants` — *"Missing coverage mapping for variants: [\"Ping\"]"*. The manifest is meant to track every `ControlCommand` variant and `Ping` was added without one. This is the test doing its job; it just had no way to say so.
- `pnl_subscription_fires_on_change`, `pnl_single_dispatches_position_info`, `cancel_pnl_stops_dispatch` — all three fail with *"PnL callback expected"*. Not credential-dependent.
- `compat_suite`, `ccp_reconnect_with_cached_credentials`, `farm_reconnect_with_cached_credentials`, `hotloop_auto_reconnect_on_farm_disconnect` — reconnect and compat paths.

I have not diagnosed the PnL three or the reconnect four. They may be stale expectations rather than live defects, but they cannot be assumed either way after a period of never running.

## The Python bindings are never compiled by a test run

`src/python` sits behind a non-default feature, so `cargo test` does not build it. Nothing automated exercises any Python path.

Building the extension and running its suite gives **276 passed, 23 failed, 115 skipped**. Every failure is `RuntimeError: Not connected` — `test_compat_tier1`, `tier2`, `tier3` and a few others expect a session. Either they need one, or they need the skip the other 115 already take.

This matters beyond the count: a wrong fix in that module cannot be caught by any current automated run, only by reading it.

## Suggested order

1. The manifest mapping for `Ping` — small, and it is a real gap in variant coverage.
2. Triage the PnL three; they share a failure mode.
3. Decide whether the connection-dependent Python tests skip or gate on a session, then add the Python job to the workflow.
4. Once green, extend the workflow from `--lib` to the full `cargo test`.

userFRM:
## Summary

- There is no test workflow. `.github/workflows` holds only the two documentation jobs, so nothing runs the 802 unit tests on a push or a pull request.
- Adding one first needs the test targets to compile. `tests/ib_paper_compat` and `tests/scenarios` were left behind by three struct changes — `ControlCommand::FetchContractDetails` gained `filters`, and `OrderRequest`'s trailing-stop and adjustable-stop variants gained `trail_stop_price` and the adjustable trailing pair. Thirteen literals, all mechanical.
- The workflow installs tzdata first. `config::expiry_tests` resolves zones by their legacy aliases (`US/Eastern`), which live in tzdata's backward-compatibility file and are absent from minimal installs — those two tests fail on such a box for that reason alone, which is why they look like permanent failures locally.

Closes #301.

## What this deliberately does not run, and why

- **The integration tests.** The workflow builds `--all-targets`, so they must keep compiling, but it runs only `--lib`. With the compile fixed, nine of them fail: a stale coverage manifest (`ControlCommand::Ping` has no mapping), three PnL scenarios, and several reconnect tests. A workflow that is red from its first run gets ignored rather than fixed, so those are worth a separate look before they gate anything.
- **The Python bindings.** `src/python` sits behind a non-default feature, so a plain `cargo test` never compiles it — every Python path is currently unexercised by any automated run. Its suite has 276 passing and 23 failing, the failures all `RuntimeError: Not connected`, so it needs a session or those tests need skipping before it can gate merges.

Both are in ibx#300.

## Test plan

- [x] `cargo build --lib --all-targets` — clean; every test target compiles for the first time.
- [x] `cargo test --lib` — 802 pass. The two `config::expiry_tests` failures are the tzdata aliases above and pass on a runner with the full database, which is why the workflow installs it.
- [x] `cargo test --no-fail-fast` across every target, to establish what the integration suite actually does once it compiles: 911 passed, 11 failed, 8 ignored.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -1,0 +1,45 @@
+name: tests
+
+on:
+  push:
+    branches: [main]
+  pull_request:
+  workflow_dispatch:
+
+concurrency:
+  group: tests-${{ github.ref }}
+  cancel-in-progress: true
+
+env:
+  CARGO_TERM_COLOR: always
+
+jobs:
+  rust:
+    name: cargo test --lib
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+
+      # `config::expiry_tests` resolves zones by their legacy aliases
+      # (`US/Eastern`). On 24.04 — which `ubuntu-latest` resolves to — those
+      # aliases are not in `tzdata` at all: they moved to `tzdata-legacy`, so a
+      # full `tzdata` install still leaves `/usr/share/zoneinfo/US/` absent and
+      # both tests failing.
+      - name: Install tzdata
+        run: sudo apt-get update && sudo apt-get install -y tzdata tzdata-legacy
+
+      - uses: dtolnay/rust-toolchain@stable
+
+      - uses: Swatinem/rust-cache@v2
+
+      - name: Build
+        run: cargo build --lib --all-targets
+
+      - name: Test
+        run: cargo test --lib
+
+      # The bindings are behind a non-default feature, so nothing above builds
+      # them. Their tests need a live session and cannot run here, but keeping
+      # them compiling is most of the value and costs one step.
+      - name: Build the Python bindings
+        run: cargo build --lib --features python
--- a/tests/ib_paper_compat/account.rs
+++ b/tests/ib_paper_compat/account.rs
@@ -388,7 +388,7 @@ pub(super) fn phase_enriched_order_cache(conns: Conns) -> Conns {
     // Fetch secdef first to populate contract cache with exchange/localSymbol/tradingClass
     control_tx.send(ControlCommand::FetchContractDetails {
         req_id: 9999, con_id: 756733, symbol: String::new(),
-        sec_type: String::new(), exchange: String::new(), currency: String::new(),
+        sec_type: String::new(), exchange: String::new(), currency: String::new(), filters: Default::default(),
     }).unwrap();
 
     let order_id = next_order_id();
@@ -547,7 +547,7 @@ pub(super) fn phase_enriched_open_orders(conns: Conns) -> Conns {
     // Fetch secdef to populate contract cache
     control_tx.send(ControlCommand::FetchContractDetails {
         req_id: 9998, con_id: 756733, symbol: String::new(),
-        sec_type: String::new(), exchange: String::new(), currency: String::new(),
+        sec_type: String::new(), exchange: String::new(), currency: String::new(), filters: Default::default(),
     }).unwrap();
 
     let order_id = next_order_id();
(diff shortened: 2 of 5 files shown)


userFRM:
Reviewed, and the finding is one the commit argues against itself: **this workflow was red on its own first run.** Fixed in `fde8fce`.

The message says a workflow red from the start gets ignored rather than fixed, and then installed `tzdata` to satisfy `config::expiry_tests`. On 24.04 — which `ubuntu-latest` resolves to — the legacy aliases those tests use are not in `tzdata` at all. They were split into `tzdata-legacy`. Verified on a 24.04 box here: `tzdata` 2026b installed, `dpkg -L tzdata` lists no `zoneinfo/US` path, `/usr/share/zoneinfo/US/` does not exist, and exactly those two tests fail. The step's comment blamed "minimal installs"; they are absent from full ones.

Now installs both. Note that #336 fixes the same failures in the library, by resolving the legacy names when the system does not carry them — the two are complementary, and CI should be green whichever lands first.

**Also added: a build of the Python bindings.** They sit behind a non-default feature, so nothing else in the workflow compiles them, and nothing was keeping them compiling. Their tests need a live session and stay out, but the build is one step and most of the value.

**Corrected in the message:** it claimed `tests/scenarios` was left behind by the struct changes. No hunk touches it and it compiles unmodified — all thirteen literals are in `tests/ib_paper_compat`.

**Not taken: pinning the actions to SHAs.** It is the safer practice, but the repository's other two workflows both float on tags, and quietly diverging in a PR about running tests is the wrong place to change that policy. Worth doing across all three deliberately if wanted.

The rest checks out: failures do fail the job, no `continue-on-error` or `|| true` masking anything, `--all-targets` builds so the integration tests stay honest about compiling, and the added struct fields use the documented not-set sentinels. Both workflow steps run clean locally.

userFRM:
No further changes needed here. The compile restoration and the CI claims both check out.

One caveat it raised, which I am leaving as designed: `push` only covers `main`, so a branch pushed without a pull request runs nothing. That is deliberate — pull requests are where the gate matters, and building every branch push costs runner time for no extra signal. Worth revisiting only if branches routinely sit unpushed to a PR for long stretches.

userFRM:

PR #304, opened 2026-07-29
tick decode: correct five 35=P field identities (ibx#303)

### LINKED ISSUE #303: tick decode: five 35=P field identities are wrong — volume, last size, high, open and the timestamp
## What I measured

I logged every decoded `35=P` entry as `(wire tick type, magnitude)` before any mapping, on a live front-month MNQ subscription for three minutes: 733 entries. MNQ's minTick is 0.25, and the last-price magnitudes confirm the scale — 111870 × 0.25 = 27967.50, a correct MNQ price.

| wire type | ibx calls it | n | measured |
|---|---|---|---|
| 2 | `O_LAST_PRICE` | 213 | 27967.50 → 27992.00, oscillating — correct |
| 3 | `O_HIGH_PRICE` | 1 | **27922.00**, constant |
| 6 | `O_VOLUME` | 120 | values 1–6, **60 decreases in 119 transitions** |
| 8 | `O_OPEN_PRICE` | 1 | **28177.25**, constant |
| 9 | `O_LOW_PRICE` | 1 | 27640.50, constant — consistent |
| 10 | `O_TIMESTAMP` | 228 | 983371 → 986141, **0 decreases in 227 transitions** |
| 12 | `O_LAST_SIZE` | 1 | 0, constant |
| 20 | not decoded | 2 | 1785325554, and 20260728 |
| 21 | not decoded | 164 | 0 → 179, **monotonic, +1 per second** |
| 22 | `O_CLOSE_PRICE` | 1 | 27962.25, constant |

## What that establishes

**Type 6 is not volume.** It decreased 60 times. Cumulative volume cannot decrease. Its values are 1–6 and it arrives with trades — it is a per-trade size.

**Type 10 is the volume.** 228 samples, zero decreases across 227 transitions, and its increments are 1, 2, 3, 4, 5, 6 — the same distribution as the type-6 values. ibx writes this to `q.timestamp_ns`.

**Type 3 is not the high.** It reads 27922.00 while the last trade over the same window ranged 27967.50–27992.00. A session high below the last traded price is not possible.

**Type 8 is not the open.** It reads 28177.25, above every last price observed, which is what a session high looks like. Type 9 at 27640.50 sits below everything, consistent with the low it is already mapped to — so 8 and 9 look like the high/low pair, and 3 is one of open/close.

**Type 12 is not the last size.** One sample, value 0, while 120 genuine per-trade sizes arrived on type 6.

**Types 20 and 21 carry the timestamp and are dropped.** Type 21 increments by exactly 1 per wall-clock second across 164 samples; type 20 carries 1785325554 — a Unix timestamp for the capture window — and 20260728, a `yyyymmdd` date. That is the trade time ibx currently sources from type 10.

## Impact

Everything reading volume gets a per-trade size that oscillates — a VWAP, a volume filter or a bar aggregation is working from noise. `last_size` is always zero. The session high is reported as a value below the last trade, and the open is reported as the high. `timestamp_ns` carries a running volume count, so anything treating it as a time is far wrong rather than slightly.

None of it looks broken from the outside: every value is a plausible number of the right magnitude.

This supersedes #290, which reported the volume half of it from the API side.

## Suggested checks

Each is independent and reproducible from a capture:

- type 10 never decreases across a session; type 6 does
- type 8 ≥ every type 2 seen; type 9 ≤ every type 2 seen
- type 21 advances one per wall-clock second
- sum of type 6 over a window ≤ the type 10 span over the same window

I have not established which of 3 and 22 is the open and which the close, and one sample each is not enough to say. Type 13 (`O_LAST_EXCH`) only ever read 0 here, so this capture says nothing about it either way.

userFRM:
## Summary

Logging every decoded `35=P` entry as `(wire type, magnitude)` before any mapping, on a live front-month MNQ subscription for three minutes — 733 entries — measures these fields rather than inferring them from what the API produces. MNQ's minTick is 0.25 and the last-price magnitudes confirm the scale: 111870 × 0.25 = 27967.50.

| wire type | was | is | what the capture shows |
|---|---|---|---|
| 6 | `O_VOLUME` | `O_LAST_SIZE` | 120 samples, values 1–6, **60 decreases in 119 transitions** — a running total cannot decrease |
| 10 | `O_TIMESTAMP` | `O_VOLUME` | 228 samples, **0 decreases in 227**, increments 1–6 matching the type-6 sizes |
| 8 | `O_OPEN_PRICE` | `O_HIGH_PRICE` | 28177.25, matching the daily bar high for the session exactly |
| 3 | `O_HIGH_PRICE` | `O_CLOSE_PRICE` | 27922.00 while the last trade ran 27967.50–27992.00. **A high cannot be below the last trade.** Matches the prior session's daily close exactly |
| 22 | `O_CLOSE_PRICE` | `O_OPEN_PRICE` | 27962.25, matching the daily bar open for the session exactly |
| 12 | `O_LAST_SIZE` | not decoded | one sample, value 0, while 120 real sizes arrived on type 6 |
| 20 / 21 | not decoded | timestamp | 21 advances **+1 per wall-clock second** over 164 samples; 20 carries 1785325554 |

Closes #303. Supersedes ibx#290, which reported the volume half from the API side.

## What the caller sees today

Volume is a per-trade size that oscillates, so a VWAP or a volume filter is working from noise. `last_size` is always zero. The session high is reported below the last trade and the open is reported as the high. `timestamp_ns` carries a running volume count. Every one of those is a plausible number of the right magnitude, which is why none of it looks broken.

## Deliberately not settled here

- **How 20 and 21 combine.** The timestamp is taken from 20, guarded on the value looking like an epoch, since the same type also carried a `yyyymmdd` value and a date read as an epoch is worse than no timestamp. Type 21 is left undecoded rather than guessed at.
- **Type 13** (`O_LAST_EXCH`) only ever read 0 here, so this capture says nothing about it.

## Test plan

- [x] `cargo test --lib` — 803 pass. The two `config::expiry_tests` failures are pre-existing on main.
- [x] `wire_tick_identities_match_what_the_feed_sends` asserts each measured wire number and that no two fields share one.
- [x] Remapping all five fields broke **nothing** in the suite beforehand — no test tied a wire number to a meaning, which is the other half of why this survived. That test now exists, so a future remap has to change a test that states its reason.
- [ ] Not covered: a capture spanning a session boundary, which is what would settle open versus close and how the two timestamp parts combine.

### DIFF
--- a/src/api/client/tests.rs
+++ b/src/api/client/tests.rs
@@ -1782,11 +1782,19 @@ fn process_msgs_dispatches_all_quote_fields() {
     assert!(w.events.iter().any(|e| e.starts_with("tick_price:1:7:")));   // low
     assert!(w.events.iter().any(|e| e.starts_with("tick_price:1:9:")));   // close
     assert!(w.events.iter().any(|e| e.starts_with("tick_price:1:14:"))); // open
-    // tick_size for: bid_size(0), ask_size(3), last_size(5), volume(8)
-    assert!(w.events.iter().any(|e| e.starts_with("tick_size:1:0:")));   // bid_size
-    assert!(w.events.iter().any(|e| e.starts_with("tick_size:1:3:")));   // ask_size
-    assert!(w.events.iter().any(|e| e.starts_with("tick_size:1:5:")));   // last_size
-    assert!(w.events.iter().any(|e| e.starts_with("tick_size:1:8:")));   // volume
+    // tick_size for: bid_size(0), ask_size(3), last_size(5), volume(8).
+    // Assert the delivered quantity, not just that a tick appeared — the
+    // scaling defect in ibx#287 fired every one of these with a value four
+    // orders of magnitude off, and a `starts_with` check passed throughout.
+    let delivered = |prefix: &str| -> Option<f64> {
+        w.events.iter().find(|e| e.starts_with(prefix))
+            .and_then(|e| e.rsplit(':').next())
+            .and_then(|v| v.parse().ok())
+    };
+    assert_eq!(delivered("tick_size:1:0:"), Some(1000.0), "bid_size");
+    assert_eq!(delivered("tick_size:1:3:"), Some(2000.0), "ask_size");
+    assert_eq!(delivered("tick_size:1:5:"), Some(500.0), "last_size");
+    assert_eq!(delivered("tick_size:1:8:"), Some(10_000.0), "volume");
 }
 
 #[test]
--- a/src/bin/bench_decode.rs
+++ b/src/bin/bench_decode.rs
@@ -295,10 +295,10 @@ fn apply_tick(q: &mut Quote, tick: &RawTick, min_tick_scaled: i64) {
         tick_decoder::O_LOW_PRICE => q.low = tick.magnitude * min_tick_scaled,
         tick_decoder::O_OPEN_PRICE => q.open = tick.magnitude * min_tick_scaled,
         tick_decoder::O_CLOSE_PRICE => q.close = tick.magnitude * min_tick_scaled,
-        tick_decoder::O_BID_SIZE => q.bid_size = tick.magnitude,
-        tick_decoder::O_ASK_SIZE => q.ask_size = tick.magnitude,
-        tick_decoder::O_LAST_SIZE => q.last_size = tick.magnitude,
-        tick_decoder::O_VOLUME => q.volume = tick.magnitude,
+        tick_decoder::O_BID_SIZE => q.bid_size = ibx::types::qty_from_wire(tick.magnitude),
+        tick_decoder::O_ASK_SIZE => q.ask_size = ibx::types::qty_from_wire(tick.magnitude),
+        tick_decoder::O_LAST_SIZE => q.last_size = ibx::types::qty_from_wire(tick.magnitude),
+        tick_decoder::O_VOLUME => q.volume = ibx::types::qty_from_wire(tick.magnitude),
         tick_decoder::O_TIMESTAMP | tick_decoder::O_LAST_TS => {
             q.timestamp_ns = tick.magnitude as u64;
         }
--- a/src/bin/bench_decode.rs
+++ b/src/bin/bench_decode.rs
@@ -51,7 +51,7 @@ fn main() {
             (1, 2, 20055, false),
             (2, 2, 20052, false),
             (4, 2, 200, false),
-            (6, 4, 1_000_000, false),
+            (tick_decoder::O_VOLUME as u64, 4, 1_000_000, false),
         ]),
     ]);
 
@@ -299,8 +299,8 @@ fn apply_tick(q: &mut Quote, tick: &RawTick, min_tick_scaled: i64) {
         tick_decoder::O_ASK_SIZE => q.ask_size = ibx::types::qty_from_wire(tick.magnitude),
         tick_decoder::O_LAST_SIZE => q.last_size = ibx::types::qty_from_wire(tick.magnitude),
         tick_decoder::O_VOLUME => q.volume = ibx::types::qty_from_wire(tick.magnitude),
-        tick_decoder::O_TIMESTAMP | tick_decoder::O_LAST_TS => {
-            q.timestamp_ns = tick.magnitude as u64;
+        tick_decoder::O_TS_BASE if tick.magnitude > 1_000_000_000 => {
+            q.timestamp_ns = (tick.magnitude as u64).saturating_mul(1_000_000_000);
         }
         _ => {}
     }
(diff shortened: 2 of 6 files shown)


userFRM:
The tests pinned the constants table but not where each type lands, which left this change's own named failure mode uncovered.

Swapping the `open` and `close` apply arms — table intact — passed the entire suite. That is exactly the case the commit message calls out: two plausible prices exchanged, silently, with the P&L path reading the wrong one. Storing the timestamp as raw seconds instead of nanoseconds also passed, and so did dropping the guard that keeps a date-shaped magnitude out of the timestamp field.

Two tests now drive real `35=P` bodies through the decode path with distinct magnitudes per field, asserting each one lands where it belongs, plus the epoch-second conversion and the date-shaped value being dropped rather than stored. All three mutations fail.

Also removed `VOLUME_MULT`. It had no consumers and its comment asserted that IB encodes volume times 10,000 — which the sibling commit on this branch measured as false while correcting the quantity scaling, and it survived both commits as live misinformation in the middle of the constants block. Types 12 and 23 now carry a line each saying what they are no longer read as, so the two deliberate gaps read as deliberate.

userFRM:
Two follow-ups from the second review, both about things that read as one identity and now mean another.

**The benchmark payloads hardcoded wire type 6.** That used to be volume; after this change it is last size, so both benchmarks silently began measuring a different tick — with a last size of 1,000,000, which is not a realistic one. They now name the constant they mean rather than the number, so a future identity change cannot quietly repoint them.

**A doc comment still named the old close identity.** Corrected to name the constant instead of a number that is no longer true.

Independent confirmation of the part that matters, from an end-to-end probe over all seven price identities and all four size identities: every price lands in its own `Quote` member scaled by the tick size, every size scaled by the quantity factor, and none crosses over. Swapping only the open and close destinations with the constants intact fails the binding test — open received 505 where 504 was expected — so the landing is pinned independently of the table, which is what was missing before.

Both timestamp mutations also fail: storing raw seconds, and dropping the guard that keeps a date-shaped magnitude out of the field.

On the parent commit in this stack: quantity-scaling the other six price arms survived its own tests, since only `bid` had a regression test there. The binding test added here catches it — I checked by mutating five price arms to the quantity scale, and it fails by name.

userFRM:
Confirmed on two more instruments, with a check that does not rest on monotonicity: an OHLC set has to be internally consistent, and the current mapping produces one that cannot exist.

SPY, minTick 0.01, captured while the session was still open so no close had been stamped:

| wire | magnitude | price | mapped today | mapped here |
| --- | --- | --- | --- | --- |
| 2 | 74456, 74469 | 744.56, 744.69 | last | last |
| 3 | 74169 | 741.69 | high | open |
| 8 | 74889 | 748.89 | open | high |
| 9 | 73768 | 737.68 | low | low |

As mapped today that is a session high of 741.69 while the last trade printed at 744.69, and an open of 748.89 sitting above that high. Neither is a state a book can be in. Read as this PR maps them the set orders cleanly: low 737.68, open 741.69, last 744.69, high 748.89.

Front month ES, minTick 0.25, captured after the close:

| wire | magnitude | price |
| --- | --- | --- |
| 3 | 29890 | 7472.50 |
| 8 | 30164 | 7541.00 |
| 9 | 29710 | 7427.50 |
| 22 | 29918 | 7479.50 |

Same shape: today's mapping gives an open of 7541.00 above a high of 7472.50. Remapped, the range low 7427.50 to high 7541.00 contains both the open at 7472.50 and type 22 at 7479.50.

The type 10 reading holds up as well. That ES subscription carried a single type 10 of 1737932. Under the nanosecond timestamp it is assigned today that is 1.7 ms past the epoch. As the front month's session volume it is the right order of magnitude. Type 20 carried 20260730 on the same message and an epoch second value of 1785534833 on the SPY subscription, which is the split described above.

The type 6 pairing reproduces exactly. Across the SPY capture type 6 appeared twice and both times as the tick immediately after a type 2 on the same tag in the same message, carrying 80 and 90 against last prices of 744.69 and 744.56. It never appeared on its own, and in a five minute window with no trades it did not appear at all.

One thing these captures do not reproduce: SPY never sent type 10 across two runs, while it did send the type 2 and type 6 trade pairs. On that feed the running total was absent rather than mismapped. It does not bear on the identities, but the field is evidently not guaranteed to arrive.

userFRM:
Correction to the tables in my comment above: type 3 is the prior session's close and type 22 the current session's open, which is how this branch maps them. I had 3 down as the open.

The coherence argument is unchanged, since what it turns on is that 741.69 cannot be a session high while the last trade printed 744.69. Restated with the right labels, SPY read low 737.68, prior close 741.69, last 744.69, high 748.89, with no type 22 since that session had not closed. Front month ES after its close read low 7427.50, prior close 7472.50, open 7479.50, high 7541.00.

userFRM:

Issue #230, opened 2026-07-15
contract details: sec_type is a Debug derive ('Stock' not 'STK') so Contracts can't round-trip; Other=>STK; market_name dead; ContractDetails.contract clones on get


deepentropy:
## Summary

Four defects on the data `contract_details` hands back. The first makes a returned `Contract` un-reusable; the last makes a natural Python idiom a silent no-op.

## 1. `sec_type` is a `Debug` derive, so a returned Contract cannot be round-tripped

`src/python/compat/contract.rs:2032`:

```rust
c.sec_type = format!("{:?}", def.sec_type);
```

`SecurityType` derives `Debug` with variants `Stock, Option, Future, Forex, Index, Bond, Warrant, Other` (`control/contracts.rs:46-55`), so `contract_details` reports **`"Stock"`**, not `"STK"`.

Verified live today — every row for AAPL/QQQ/SPY came back `sec_type='Stock'`.

That gives **three conventions in one library** for the same concept:

| Convention | Where | Stock is |
|---|---|---|
| API | what `req_contract_details` *wants* | `"STK"` |
| FIX | `send_secdef_request_by_symbol` (STK→CS), `symbol_samples` via `to_fix()` | `"CS"` |
| Debug derive | the `contract_details` callback | `"Stock"` |

So the obvious thing — take the Contract you got back, feed it to another call — sends `167=Stock`, which `from_fix` maps to `SecurityType::Other`. **Silently the wrong instrument type.**

`sec_type_to_str` already exists in `ccp.rs:39-50` and produces the correct `"STK"`; it is used for the internal contract cache but not here.

## 2. `sec_type_to_str` maps `Other => "STK"`

`ccp.rs:48`:

```rust
SecurityType::Other => "STK",
```

An instrument ibx could not classify is reported as a **stock**. For a library whose order path is STK-only (`validate_order_contract`), calling an unknown thing a stock is the one wrong guess that will not be caught downstream. `""` or an error would both be safer.

## 3. `market_name` is parsed and then thrown away

`contract.rs:2044`:

```rust
market_name: String::new(),
```

Even though `def.market_name` **is** parsed from tag 58 (`control/contracts.rs:321-323`). The field is always `""`. Verified live: every row. Dead field — either wire it up or remove it.

## 4. `ContractDetails.contract` clones on get, so mutating it is a silent no-op

```python
>>> d = ContractDetails()
>>> c = Contract(); c.con_id = 265598
>>> d.contract = c
[...]

PR #306, opened 2026-07-29
historical data: send the contract's own security type and exchange (ibx#305)

### LINKED ISSUE #305: historical data: secType and exchange are hardcoded to CS/SMART, so a request for any non-stock contract is rejected
## Problem

Both historical query builders fill the contract descriptor with constants rather than the requested contract's own values:

https://github.com/deepentropy/ibx/blob/main/src/engine/hot_loop/hmds.rs#L671

```rust
let req = crate::control::historical::HistoricalRequest {
    query_id: query_id.clone(),
    con_id: con_id as u32,
    symbol: symbol.to_string(),
    sec_type: "CS",
    exchange: "SMART",
```

and again at the keep-up-to-date path:

https://github.com/deepentropy/ibx/blob/main/src/engine/hot_loop/hmds.rs#L758

```rust
    sec_type: "CS",
    exchange: "SMART",
```

`ControlCommand::FetchHistorical` carries neither field, so `req_historical_data`'s contract argument contributes only `con_id` and `symbol` — its `sec_type` and `exchange` are dropped before the request is built.

## Observed

Requesting a daily bar for a front-month CME future (`MNQU6`, conId 793356225, `secType FUT`, `exchange CME`) on a live account:

```
[error] req_id=3 code=162 msg=BEST queries are not supported for this contract
```

`historical_data_end` then fires with no bars. The subscription itself is fine — a `req_mkt_data` on the same contract in the same session streams normally, so this is the historical path specifically.

## Impact

Historical data appears to work only for US stocks routable through SMART. For futures, options, FX or anything venue-specific, the request is built describing a different contract than the caller asked for, and the server rejects it. Because the rejection arrives as an error callback and `historical_data_end` still fires, a caller that only waits for the end marker sees an empty result rather than a failure — the same silent-empty shape as several other issues here.

Worth noting the tick-by-tick query has the same shape, with `<exchange>BEST</exchange>` and `<secType>CS</secType>` hardcoded in its XML at `hmds.rs:562`, so tick-by-tick for a non-stock is likely rejected the same way. I have not confirmed that live.

## Fix

Carry `sec_type` and `exchange` on `FetchHistorical` from the caller's `Contract`, the way `FetchContractDetails` already carries its own fields, and pass them into `HistoricalRequest` rather than the constants. The values are in hand at `req_historical_data` — the caller supplies a full `Contract` and the two fields are simply not read.

Found while trying to use a daily bar as an independent check on the `35=P` OHLC field identities in #303; the request that was meant to be the reference could not be issued.

userFRM:
## Summary

- Both historical query builders filled the contract descriptor with constants — `sec_type: "CS"`, `exchange: "SMART"` — and `ControlCommand::FetchHistorical` carried neither field, so the caller's `Contract` contributed only its conId and symbol. Every request described a stock on SMART regardless of what was asked for, and SMART maps to BEST on this query.
- Requesting a daily bar for a front-month CME future on a live account returns `error 162: BEST queries are not supported for this contract`, and no bars. `req_mkt_data` on the same contract in the same session streams normally, so only the historical path was affected.
- The failure is quiet from the caller's side: the rejection arrives as an error callback but `historical_data_end` still fires, so anything waiting on the end marker sees an empty result rather than a failure.
- Both fields now travel on the command and reach the query. An empty security type still falls back to the stock encoding and an empty exchange to SMART, so a caller supplying neither gets exactly what it got before. A supplied type goes through the same `SecurityType` mapping the subscription path already uses, so `STK` still encodes as `CS`.

Closes #305.

## Test plan

- [x] `cargo test --offline --lib` — 809 passed. The 2 failures are `config::expiry_tests::{named_zone_converts_with_dst, instant_round_trips_to_wire}`, which fail on the base commit too: the host has no legacy timezone files (fixed separately in #336).
- [x] `cargo check --offline` on every target that builds offline — `--lib`, `--features python`, `--test control_plane`, `--test scenarios`, `--test hot_loop_lifecycle`: all clean. `control_plane` was broken by this branch and is the reason the list is now explicit.
- [x] Mutation check: reinstating `CS`/`SMART` in the batch builder fails `the_query_on_the_wire_carries_the_contract_s_own_type_and_venue`; in the keep-up-to-date builder alone, `the_streaming_query_carries_them_too`; dropping the unlisted-type passthrough fails `a_stated_security_type_reaches_the_wire_in_its_own_spelling`; dropping the client forwarding fails `req_historical_data_carries_the_contract_s_own_type_and_venue`. All by name.
- [ ] Not in this PR: the head-timestamp constructor still hardcodes `CS`/`SMART` — same family, same file, its own change.

### DIFF
--- a/src/api/client/reference.rs
+++ b/src/api/client/reference.rs
@@ -19,6 +19,8 @@ impl EClient {
             req_id: req_id as u32,
             con_id: contract.con_id,
             symbol: contract.symbol.clone(),
+            sec_type: contract.sec_type.clone(),
+            exchange: contract.exchange.clone(),
             end_date_time: end_date_time.into(),
             duration: duration.into(),
             bar_size: bar_size.into(),
--- a/src/api/client/tests.rs
+++ b/src/api/client/tests.rs
@@ -1190,13 +1190,41 @@ fn req_historical_data_sends_fetch_historical() {
     client.req_historical_data(5, &spy(), "20260101 16:00:00", "1 D", "1 hour", "TRADES", true, 1, false).unwrap();
     let cmd = rx.try_recv().unwrap();
     match cmd {
-        ControlCommand::FetchHistorical { req_id, con_id, duration, bar_size, what_to_show, use_rth, .. } => {
+        ControlCommand::FetchHistorical {
+            req_id, con_id, duration, bar_size, what_to_show, use_rth, sec_type, exchange, ..
+        } => {
             assert_eq!(req_id, 5);
             assert_eq!(con_id, 756733);
             assert_eq!(duration, "1 D");
             assert_eq!(bar_size, "1 hour");
             assert_eq!(what_to_show, "TRADES");
             assert!(use_rth);
+            // The contract's own fields have to leave the client, or the
+            // engine has nothing but the old constants to fall back on
+            // (ibx#305). `spy()` states neither, so both arrive empty and the
+            // engine substitutes — that substitution is tested at its source.
+            assert_eq!(sec_type, "");
+            assert_eq!(exchange, "");
+        }
+        _ => panic!("expected FetchHistorical"),
+    }
+}
+
+/// A contract that does state its own type and venue must carry both, which is
+/// the whole of what this fixes: every historical query described itself as a
+/// SMART-routed stock regardless of the contract asked for.
+#[test]
+fn req_historical_data_carries_the_contract_s_own_type_and_venue() {
+    let (client, rx, _shared) = test_client();
+    let es = Contract {
+        con_id: 495512563, symbol: "ES".into(),
+        sec_type: "FUT".into(), exchange: "CME".into(), ..Default::default()
+    };
+    client.req_historical_data(6, &es, "20260101 16:00:00", "1 D", "1 hour", "TRADES", true, 1, false).unwrap();
+    match rx.try_recv().unwrap() {
+        ControlCommand::FetchHistorical { sec_type, exchange, .. } => {
+            assert_eq!(sec_type, "FUT");
+            assert_eq!(exchange, "CME");
         }
         _ => panic!("expected FetchHistorical"),
     }
--- a/src/control/historical.rs
+++ b/src/control/historical.rs
@@ -166,8 +166,12 @@ pub struct HistoricalRequest {
     pub query_id: String,
     pub con_id: u32,
     pub symbol: String,
-    pub sec_type: &'static str,
-    pub exchange: &'static str,
+    /// Wire security type and exchange for the contract being requested.
+    /// Owned rather than static: they come from the caller's `Contract`, and
+    /// hardcoding them described a different contract than was asked for
+    /// (ibx#305).
+    pub sec_type: String,
+    pub exchange: String,
     pub data_type: BarDataType,
     pub end_time: String,
     pub duration: String,
@@ -200,7 +204,7 @@ pub struct HistoricalResponse {
 
 /// Build the XML query for a historical bar data request.
 pub fn build_query_xml(req: &HistoricalRequest) -> String {
-    let exchange = match req.exchange {
+    let exchange = match req.exchange.as_str() {
         "SMART" => "BEST",
         e => e,
     };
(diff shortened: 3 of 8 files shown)


userFRM:
Reviewed. The production change is correct and complete for what it claims — and **both tests it shipped with pass with the fix reverted.** Fixed in `f1ce38a`.

They drive `build_query_xml`, which honoured the request's fields before this change too. The constants were applied a layer above it, so replacing them changes nothing the encoder can see. Reinstating `CS`/`SMART` in the builder — the precise defect — left the whole suite green. The commit's test plan did not test the commit.

Three tests that do:

- The client test asserts the contract's type and venue leave `req_historical_data` at all, which is where they were being dropped. The existing test destructured with `..` and asserted neither.
- The two substitution helpers are pinned directly, including both spellings of a stock, the empty case that reproduces the previous behaviour exactly, and an unrecognised type staying empty so the gateway names it rather than the client silently describing it as something else.
- One drives `send_historical_request_ex` over a loopback pair and reads the bytes. This is the one that matters: helpers are worth nothing if the builder stops calling them, and my first attempt at closing this gap tested only the helpers — which passes the reverted build just as the originals did. Reinstating the constants now fails by name.

Independent verification worth recording: the spelling matches the reference mapping on both vocabularies, `CASH` still selects IDEALPRO, the empty-field fallbacks reproduce the old behaviour exactly so nothing regresses for a caller who states neither, and both engine call sites and both client constructors plumb the fields.

Known residual, disclosed in the commit already: the head-timestamp constructor still hardcodes `CS`/`SMART`. Same family, same file, and it deserves its own change rather than being folded in here.

808 pass plus the two `config::expiry_tests` timezone failures. `--features python` clean.

userFRM:
Updated in `b343c20`. Three changes.

**The passthrough is now bounded.** Letting an unlisted security type through means putting caller-supplied text into a query document, so `FOP&` produced a malformed query rather than a refused one. What passes through is restricted to the shape a security type has — short, alphanumeric — and anything else is blanked, which is what the enum already did for everything.

**The claim is narrowed to what the change delivers.** Two corrections:

- The subscribe path passes unlisted types through only on its descriptive branch. A subscription with a known conId sends `CS` regardless and lets the server resolve from the conId, so the comparison does not hold in general.
- More importantly, this does not survive the likelier path. A contract returned by `req_contract_details` for an unlisted type arrives with an **empty** security type, because the enum has no variant to carry it — and an empty type still means stock here. So this helps a caller who states `FOP` themselves and does nothing for one who passes back a contract the client handed them. Recovering that needs the enum to stop discarding what it cannot name, which is #230. The comment now says so.

**The test helper reads a complete frame.** It stopped at `</secType>` — a field in the middle — so a TCP split immediately after it would have satisfied the assertions on a partial read.

809 pass plus the two `config::expiry_tests` timezone failures; all five offline targets clean. Letting anything through, and blanking everything unlisted, each fail `a_stated_security_type_reaches_the_wire_in_its_own_spelling` by name.

userFRM:
It found that my own fix introduced a hole and oversold itself. Both corrected in `d89e59d`.

**The passthrough was unbounded.** Letting an unlisted security type through meant putting caller-supplied text into a query document unescaped, so `FOP&` produced a malformed query rather than a refused one. That is a defect I added while removing a different one. What passes through is now restricted to the shape a security type actually has — short, alphanumeric — and anything else is blanked, which is what the enum already did for everything.

**And the claim was wider than the change.** Two corrections:

- I wrote that the subscribe path passes unlisted types through. Only its descriptive branch does. A subscription with a known conId sends `CS` regardless and lets the server resolve from the conId, so the comparison did not hold in general.
- More seriously, the fix does not survive the likelier path. A contract returned by `req_contract_details` for an unlisted type already arrives with an **empty** security type, because the enum has no variant to carry it — and an empty type still means stock here. So this helps a caller who writes `FOP` themselves and does nothing for a caller who passes back a contract the client handed them. Recovering that needs the enum to stop discarding what it cannot name, which is #230. The comment says so now instead of leaving the impression it is covered.

**The test helper could return a truncated frame.** It stopped reading at `</secType>` — a field in the middle of the frame — so a TCP split immediately after it would have satisfied the assertions on a partial read. It stops on a complete frame now.

Mutation-verified both directions: letting anything through, and blanking everything unlisted, each fail `a_stated_security_type_reaches_the_wire_in_its_own_spelling` by name.

809 pass plus the two `config::expiry_tests` timezone failures. All five offline targets check clean.

userFRM:

PR #310, opened 2026-07-29
ccp: report the filled quantity, not the unfilled remainder (ibx#309)

### LINKED ISSUE #309: ccp: filled_quantity is populated from LeavesQty, so a completed order reports as entirely unfilled
## Problem

`Order.filled_quantity` is assigned the value parsed from tag 151:

https://github.com/deepentropy/ibx/blob/main/src/engine/hot_loop/ccp.rs#L1043

```rust
filled_quantity: leaves_qty as f64,
```

where `leaves_qty` comes from LeavesQty — the quantity still working:

https://github.com/deepentropy/ibx/blob/main/src/engine/hot_loop/ccp.rs#L799

```rust
let leaves_qty = parsed.get(&151).and_then(|s| s.parse::<i64>().ok()).unwrap_or(0);
```

The same variable is used correctly elsewhere in the same handler, as `remaining` and as `remaining_qty`. Only this assignment takes it for the opposite quantity.

The correct source is already parsed, in scope, and used for the order-status callback:

https://github.com/deepentropy/ibx/blob/main/src/engine/hot_loop/ccp.rs#L921

```rust
let cum_qty: f64 = parsed.get(&14).and_then(|s| s.parse().ok()).unwrap_or(0.0);
```

## Impact

LeavesQty and CumQty are complements, so the reported figure is wrong by exactly the amount that has not filled.

- A 100-lot with 30 done reports `filled_quantity = 70`.
- A **completed** order has LeavesQty 0, so it reports `filled_quantity = 0` — fully filled, reported as entirely unfilled.

That second case is the dangerous one. It is the state a reconciliation is most likely to act on, and zero-filled on a completed order looks like an order that never worked rather than like a bug. Anything sizing a follow-up order, computing a position from callbacks, or deciding whether to re-send will act on the inverse of the truth.

## Fix

Assign `cum_qty`. One line, and the value is already there.

Worth a test that a partial fill reports the filled side rather than the remainder — the two are easy to swap and nothing in the suite distinguished them.

userFRM:
## Summary

- `Order.filled_quantity` was assigned `leaves_qty`, parsed from tag 151 — **LeavesQty**, the part of the order still working. The filled amount is tag 14, **CumQty**, which is already parsed a hundred lines above and used for the order-status callback.
- The two are complements, so the reported figure is wrong by exactly the amount that has *not* filled. A 100-lot with 30 done reports 70. A **completed** order has LeavesQty 0, so it reports as entirely unfilled.
- That second case is the dangerous one: it is the state a reconciliation is most likely to act on, and zero-filled on a completed order reads like an order that never worked rather than like a bug. Anything sizing a follow-up, computing a position from callbacks, or deciding whether to re-send acts on the inverse of the truth.
- The same variable is correct where it is used as `remaining` and `remaining_qty`. Only this one assignment takes it for the opposite quantity.

Closes #309.

## Test plan

- [x] `cargo test --lib` — 807 pass. The two `config::expiry_tests` failures are pre-existing on main.
- [x] `filled_quantity_is_the_filled_amount_not_the_remainder` drives a partial fill — 100 ordered, 30 filled, 70 working — through `handle_exec_report` and asserts the reported figure is 30. Restoring the old assignment fails it with 70.

### DIFF
--- a/src/engine/hot_loop/ccp.rs
+++ b/src/engine/hot_loop/ccp.rs
@@ -918,7 +918,11 @@ impl CcpState {
             let exec_exchange = parsed.get(&30).cloned().unwrap_or_default();
             let transact_time = parsed.get(&60).cloned().unwrap_or_default();
             let avg_px: f64 = parsed.get(&6).and_then(|s| s.parse().ok()).unwrap_or(0.0);
-            let cum_qty: f64 = parsed.get(&14).and_then(|s| s.parse().ok()).unwrap_or(0.0);
+            // Absent is not zero. This value is written into a row that
+            // persists, so a later report that omits the tag — a pending
+            // cancel, say — would otherwise wipe a real filled quantity back
+            // to nothing, which is the symptom this is correcting.
+            let cum_qty: Option<f64> = parsed.get(&14).and_then(|s| s.parse().ok());
             let last_liq: i32 = parsed.get(&851).and_then(|s| s.parse().ok()).unwrap_or(0);
 
             let sec_type_str = match sec_type.as_str() {
@@ -1040,7 +1044,13 @@ impl CcpState {
                 tif: if tif_str.is_empty() { fb_tif.to_string() } else { tif_str.to_string() },
                 account: if account.is_empty() { account_id.to_string() } else { account.clone() },
                 perm_id,
-                filled_quantity: leaves_qty as f64,
+                // Tag 14 (CumQty), not tag 151 (LeavesQty). The two are
+                // complements, so reporting the remainder as the filled amount
+                // makes a completed order read as entirely unfilled (ibx#309).
+                filled_quantity: cum_qty.unwrap_or_else(|| {
+                    shared.orders.get_order_info(clord_id)
+                        .map_or(0.0, |info| info.order.filled_quantity)
+                }),
                 outside_rth,
                 clearing_intent,
                 auto_cancel_date,
@@ -1089,7 +1099,9 @@ impl CcpState {
                 shares: last_shares as f64,
                 price: last_px,
                 order_id: clord_id as i64,
-                cum_qty,
+                // The execution record describes this report, so an absent
+                // cumulative is zero here rather than the cached total.
+                cum_qty: cum_qty.unwrap_or(0.0),
                 avg_price: avg_px,
                 last_liquidity: last_liq,
                 ..Default::default()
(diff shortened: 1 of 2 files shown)


userFRM:
This PR was not what its title said. Rewritten to match.

The commit carried three unrelated things: the one-line ccp fix it describes, the **entire 108-line server-tag table redesign** — byte-identical to what #286 already proposes, so this would have duplicated that PR — and a three-line `IBX_WIRE_CAPTURE` debug hook left over from an unrelated capture session, evaluating `std::env::var` per `35=P` message on the hot loop.

The diff is now the ccp fix and its test, and nothing else. #286 keeps the server-tag redesign, where it has a message that tells its own story.

**One real gap closed along with it.** The Python client hand-copies the fields it forwards on `completed_order`, and did not include `filled_quantity` — so the corrected value stopped one hop short of the surface where the symptom actually shows, and a Python reconciliation still read `0.0` for every completed order. Both of its sites carry it now.

**The test could not tell the fix from its complement.** `total_quantity - leaves_qty` produces the same number on any frame carrying both tags, so replacing the fix with it passed. The test now also drives a frame with no tag 151, where the complement would report the whole order as filled.

That hook appearing here is the second time the same leftover has ridden into a commit in this batch — it was also caught in #316 and removed there.

userFRM:
The fallback could put the symptom straight back.

CumQty was optional at parse and defaulted to zero, then overwrote the cached field unconditionally — so a later report that omits tag 14 wiped what an earlier one had established. A pending cancel is exactly that shape: a fill sets 30, the cancel arrives without tag 14 and carries only LeavesQty, and the cache goes to 0. That is the same "completed order reads as entirely unfilled" this change exists to remove, reached from a different direction. An absent tag now keeps what is stored, and the test drives that second report.

This is the fourth place in this file where an absent value was read as a real zero — the position-feed quantity, the update-frame quantity, the midnight-seed quantity, and now this. Worth saying plainly, because it is a property of the original code rather than of any one fix.

Also pinned the other side: LeavesQty is still the remainder on the fill, which nothing asserted, so zeroing it passed the whole suite.

The rewrite check came back clean — the version of this commit that carried an unrelated 108-line redesign and a leftover debug hook is gone, and neither survives at the current tip. The Python `completed_order` copies are confirmed present on both branches of that handler; there is no behavioural test for them because `cargo test --features python` does not compile at base, on a pre-existing error in a file this branch does not touch.

userFRM:

PR #316, opened 2026-07-29
market data: describe the contract when subscribing by conId (ibx#308)

### LINKED ISSUE #308: market data: a futures subscription is acknowledged once instead of twice, so no bid or ask ever arrives
## Observed

A single `req_mkt_data` with default arguments, live account, logging every decoded `35=P` entry as `(server_tag, tick type, magnitude)` before any mapping. The realtime fan-out sends two entries in one `35=V` — `264=442` (bid/ask) and `264=443` (last) under `146=2` — for both instruments.

**Stock (AAPL, STK/SMART) — works:**

```
Sent 35=V subscribe: con_id=265598 sec_type=STK ids=1,2 seq=3
Farm 35=Q subscription ack received
Subscribed instrument 0 -> server_tag 149506, minTick 0.01
Farm 35=Q subscription ack received          <-- second ack
Subscribed instrument 0 -> server_tag 149506, minTick 0.01
Ticker setup: con_id 265598 -> server_tag 149508, minTick 0.01
```

and both halves of the feed arrive, on separate tags:

```
tag ...93:  types [0, 1, 4, 5, 7, 11, 16, 17]      <-- bid, ask, bid size, ask size
tag ...95:  types [2, 3, 6, 8, 9, 13, 20, 21, 27]  <-- last, close, high, low, sizes
```

**Future (MNQ front month, FUT/CME) — does not:**

```
Sent 35=V subscribe: con_id=793356225 sec_type=FUT ids=1,2 seq=3
Farm 35=Q subscription ack received           <-- only one
Subscribed instrument 0 -> server_tag 107684, minTick 0.25
Ticker setup: con_id 793356225 -> server_tag 107685, minTick 0.25
```

and only one tag ever carries anything:

```
tag 107685: types [2, 3, 6, 8, 9, 10, 12, 13, 20, 21, 22]
```

Tag 107684 — the one the `35=Q` ack registered — receives nothing at all. Across 733 entries in a three-minute capture, **types 0, 1, 4 and 5 never appear once**.

No rejection accompanies this: no error callback, no reject frame, nothing in the logs about the second request. The subscription reports success and trades stream normally.

## What that isolates

The same code path, the same two-entry `35=V`, the same account. A stock is acknowledged twice and delivers both halves; a future is acknowledged once and delivers only trades. So this is not the fan-out failing to send the second entry, and not an entitlement gap on the account — it is specific to how the futures subscription is formed or how its second acknowledgement is recognised.

## Impact

For futures there is no top of book. A caller gets last, volume and the session statistics but never a bid or an ask, so anything needing a spread — a marketable-limit decision, a mid, a liquidity check before sizing — has nothing to work from. `tick_price` fires for LAST and never for BID or ASK.

Because the subscription succeeds and trades do flow, it looks like a quiet instrument rather than a missing half of the feed.

## Related

On this account the tag that carries the futures data is the ticker-setup one, assigned around 107685 — above the 65536 bound in #281. Before that fix the cap discarded the only tag that delivers anything, which is why the first live attempt produced no ticks at all. #286 makes the trade stream visible; this issue is the half that is still missing.

Happy to capture anything specific that would narrow it further — a `35=V` byte dump for both instrument types side by side would probably settle it.

userFRM:
## Summary

- A 35=V subscribe built from a known conId described **every** contract as a SMART-routed common stock: SecurityType (167) was the literal `CS` and Exchange (207) the literal `BEST`. The comment said the server resolves the rest from the conId. It does not — both fields take part in routing the subscription.
- For an equity the assumption is invisible: `SMART` maps to `BEST` and `STK` to `CS` anyway, so the two literals happen to be the correct values. For anything else they are wrong, and the server neither errors nor rejects. It acknowledges the part it can serve and drops the rest in silence.
- The symptom is a futures subscription that never produces a quote. The realtime fan-out sends two entries, one for bid/ask and one for last, and only the last entry comes back acknowledged. Trades and volume arrive; bid, ask, bid size and ask size never do, and nothing on the wire explains the gap.

Closes #308.

## Evidence

Subscription acknowledgements captured on a live session. An equity gets both entries back:

```
AAPL  STK/SMART   192693,1,0.01,0,3,9c,,1,1     ← entry 1, bid/ask
                  192693,2,0.01,0,3,9c,,1,1     ← entry 2, last
MNQ   FUT/CME     152613,2,0.25,0,1,5,,1,1      ← entry 2 only
ES    FUT/CME     110927,2,0.25,0,1,5,,1,1      ← entry 2 only
```

Across 733 captured acknowledgements no futures subscription ever received the first entry. No error accompanies the one that is missing.

The account is entitled to the data — the same contracts on the same account, through the Java gateway, quote normally:

```
MNQ   bid=27897.5  ask=27898.25  bidSize=1   askSize=1
ES    bid=7454.25  ask=7454.5    bidSize=20  askSize=3
```

Sending the contract's own SecurityType and Exchange gets both entries acknowledged for the future as well, and bid/ask begins flowing immediately:

```
MNQ   ticks=48  bid=27798.25  ask=27799.00  last=27800.75
AAPL  ticks=27  bid=343.18    ask=343.21    last=343.04
```

## Change

- The conId path maps SecurityType and Exchange the way the descriptive path directly below it already did, so the two branches no longer disagree about what the contract is.
- `sec_type_to_fix` joins `exchange_to_fix`, which it sits next to and mirrors.
- The tag list moved into a free function so the wire shape can be asserted without a connection.

## Equities are unchanged

The mapping returns exactly the literals it replaces for a SMART-routed stock, so this path is byte-for-byte what it was. The stock test pins the whole ordered tag list — both the realtime fan-out and the single-entry mode — rather than only the two tags that changed, so a reordering or a dropped field fails it too.

## Tests

Four tests over the builder. Reverting the fix fails `conid_subscribe_describes_the_actual_contract` and leaves `conid_subscribe_is_unchanged_for_stocks` passing, which is the property being claimed.

## Test plan

- [x] `cargo test --offline --lib` — 808 passed. The 2 failures are `config::expiry_tests::{named_zone_converts_with_dst, instant_round_trips_to_wire}`, which fail on the base commit too: the host has no legacy timezone files (fixed separately in #336).

### DIFF
--- a/src/control/contracts.rs
+++ b/src/control/contracts.rs
@@ -192,6 +192,17 @@ impl Default for ContractDefinition {
 pub fn exchange_to_fix(exchange: &str) -> &str {
     match exchange {
         "SMART" => "BEST",
+        // Legacy spelling for NASDAQ. The depth path already translates it;
+        // routing a subscription under the old name reaches nothing.
+        "ISLAND" => "NASDAQ",
+        other => other,
+    }
+}
+
+/// Map a security type to its wire format.
+pub fn sec_type_to_fix(sec_type: &str) -> &str {
+    match sec_type {
+        "STK" => "CS",
         other => other,
     }
 }
@@ -824,8 +835,18 @@ mod tests {
     fn exchange_mapping() {
         assert_eq!(exchange_to_fix("SMART"), "BEST");
         assert_eq!(exchange_to_fix("NYSE"), "NYSE");
+        assert_eq!(exchange_to_fix("ISLAND"), "NASDAQ", "legacy spelling must route");
         assert_eq!(exchange_from_fix("BEST"), "SMART");
         assert_eq!(exchange_from_fix("ARCA"), "ARCA");
+    }
+
+    #[test]
+    fn sec_type_wire_mapping() {
+        // Only STK is renamed on the wire; everything else passes through.
+        assert_eq!(sec_type_to_fix("STK"), "CS");
+        assert_eq!(sec_type_to_fix("FUT"), "FUT");
+        assert_eq!(sec_type_to_fix("OPT"), "OPT");
+        assert_eq!(sec_type_to_fix("CASH"), "CASH");
     }
 
     #[test]
--- a/src/engine/hot_loop/farm.rs
+++ b/src/engine/hot_loop/farm.rs
@@ -12,6 +12,62 @@ use crossbeam_channel::Sender;
 
 use super::{HeartbeatState, emit, fast_extract_msg_type, find_body_after_tag};
 
+/// Build the 35=V subscribe tag list for a contract whose conId is known.
+///
+/// Kept pure so the wire shape stays unit-testable. SecurityType (167) and
+/// Exchange (207) must describe the actual contract: the server routes the
+/// subscription by them, and a mismatch is answered with a partial ack rather
+/// than an error.
+fn build_conid_subscribe_tags(
+    realtime: bool,
+    bid_ask_id: u32,
+    last_id: u32,
+    con_id: i64,
+    exchange: &str,
+    sec_type: &str,
+    mode_9887: i32,
+    ts: &str,
+) -> Vec<(u32, String)> {
+    let con_id_str = (con_id as u32).to_string();
+    // Subscribing by conId alone is a supported shape — `Contract` defaults
+    // both fields to empty, and the in-tree benchmark does exactly that. Keep
+    // the smart-routed stock those callers used to get rather than sending an
+    // empty SecurityType and Exchange, which is the silent partial-ack this
+    // change exists to remove.
+    let exchange = if exchange.is_empty() { "SMART" } else { exchange };
+    let sec_type = if sec_type.is_empty() { "STK" } else { sec_type };
+    let fix_exchange = crate::control::contracts::exchange_to_fix(exchange);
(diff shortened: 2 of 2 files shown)


userFRM:
Pushed two corrections after a closer read of the callers.

**Subscribing by conId alone kept working.** `Contract` defaults both descriptive fields to empty strings, and the in-tree benchmark subscribes that way, so replacing the literals with the mapped values sent `167=` and `207=` empty for those callers. That is the same silent partial acknowledgement this change exists to remove, arriving from the other direction. An undescribed contract now falls back to the smart-routed stock the literals were giving it, which keeps that wire shape byte-identical. Pinned by a test that asserts the undescribed and the explicitly-`SMART`/`STK` forms produce the same tag list.

**`ISLAND` maps to `NASDAQ`.** The depth path already translated it. While the conId path sent a literal exchange the difference could not surface; now that it routes the contract's own value, the legacy spelling would go out verbatim and reach nothing. Folded into `exchange_to_fix` next to `SMART`→`BEST` so the two paths cannot drift.

The claim about equities is narrowed accordingly — byte-identical holds for smart-routed equities. One on a direct exchange now routes to that exchange rather than to `BEST`, which is the intended direction of the fix.

The stock test pins the whole ordered tag list for both the realtime fan-out and the single-entry mode, so a reordering, a dropped field, or a changed constant fails it too — not only the two tags this touched.

userFRM:

Issue #321, opened 2026-07-29
ccp: the recovery insert defaults a missing Side to Sell, so later fills for that order move the position the wrong way


userFRM:
## Problem

The cross-session recovery insert derives the order's side from tag 54 with a catch-all:

```rust
let side = match parsed.get(&54).map(|s| s.as_str()) {
    Some("1") => Side::Buy,
    Some("5") => Side::ShortSell,
    _ => Side::Sell,
};
```

A report with no tag 54, or one carrying a value the arm does not name, is recorded as a sell.

## Impact

The guess is not confined to the recovered record. Once the order is in `context`, every later fill for it books through the tracked path and takes its side from the stored order — so a recovered buy that was recorded as a sell moves the position down by the filled quantity instead of up. The error is twice the fill size and it is silent: nothing distinguishes a guessed side from a stated one afterwards.

Compared with the alternative, a fill that could not be booked at all leaves the engine short of the truth and the position feed still disagrees visibly. A wrong sign puts a confident wrong number in its place.

## Solution

Require the side. A recovery record without a usable tag 54 should be refused rather than defaulted — the order stays untracked, which is the state the engine already handles, instead of being tracked with an attribute that was invented.

The same reasoning is why the untracked-fill path added in #314 refuses a report without tag 54 rather than defaulting it.

userFRM:
Closing this to clear the tracker. #409 carries the fork's main if the detail here is still useful.

Issue #320, opened 2026-07-29
ccp: on a fresh process the recovery insert tracks a replayed order, so its replayed execution books as a new fill


userFRM:
## Problem

At session start the gateway replays recent activity. Two things happen in order:

1. A New/New report (`150=0`, `39=0`) with a contract id and a quantity inserts the order into `context` — the cross-session recovery path.
2. The replayed execution for that same order arrives, carrying `97=Y` and its original ExecID.

By the time the execution is handled the order is tracked, so it books through the normal path: a `Fill` is emitted and the position moves. The ExecID window cannot stop it — a fresh process has no memory of the ExecID at all.

## Impact

Phantom fill events on every restart that has open, partially-filled orders. The position in `context` self-heals, because the position feed reconciles it, but the `Fill` events and the execution callbacks are already out — anything that acts on those rather than reconciling against the feed has acted on a fill that happened before the process started.

## Solution

Consult the resend marker on the fill path generally rather than only where the order is unknown. A replayed execution restates history in both cases; the only reason the tracked path gets away with it during a session is that the ExecID window happens to cover the reconnect burst, and that window does not exist on a fresh process.

Neither `97` nor `43` is read anywhere in the handler today.

Related: #317, where the same replay batch is delivered twice.

Found while adding the untracked-fill path in #314 — that change guards its own path against exactly this, and this is the one it cannot reach.

userFRM:
Closing this to clear the tracker. #409 carries the fork's main if the detail here is still useful.

PR #322, opened 2026-07-29
ccp: book a fill for an order this session does not track (ibx#314)

### LINKED ISSUE #314: ccp: a fill for an order this session does not track is dropped, position included, and its ExecID is consumed so the replay cannot recover it
## Problem

In the execution-report handler, the fill block is entered, the ExecID is recorded, and then the fill is booked only if the order is already in `context`:

```rust
if matches!(exec_type, "F" | "1" | "2") && last_shares > 0 {
        return;
    }
    if let Some(order) = context.order(clord_id).copied() {
        ...
        context.update_position(order.instrument, delta);
```

There is no `else`. If the order is not tracked, the fill evaporates: no `Fill`, no `exec_details`, no position update, and no log line saying anything happened.

The ExecID has already been consumed at that point. A later replay of the same execution — the one the gateway pushes after a reconnect — is then rejected as a duplicate, so the fill can never be recovered.

The order is absent in several ordinary ways:

- it reached a terminal status and was removed, which is the cancel/fill race: the cancel ack is processed first, the fill arrives second and finds nothing
- it was placed from another client, or in a previous session
- the what-if path removes it too

The recovery insert does not cover these — it only fires for `150=0 && 39=0`, and requires a contract id and a positive quantity.

## Impact

A position the account actually holds that the engine does not know about. Every subsequent sizing decision is wrong by that amount and any flatten-all logic leaves it behind. The gateway's own position feed shows the fill, so the two disagree, and that disagreement is the only signal — there is no error and no warning.

The cancel/fill race is the easiest way to hit it deliberately: place a marketable order and cancel it immediately.

## Solution

Book the fill from the execution report when the order is not tracked. The report carries what is needed — ContractID (6008), Side (54), LastShares (32), LastPx (31) — and `try_register_instrument` resolves the contract without panicking on a full table.

Both the contract and the side have to be present to book it. A guessed side moves the position the wrong way, which is worse than reporting that the fill could not be placed, so the fallback requires tag 54 rather than defaulting.

Record the ExecID only once the fill can actually be booked. An execution that was dropped must stay replayable, otherwise the reconnect replay that would have recovered it is rejected as a duplicate.

userFRM:
## Summary

- The fill block recorded the ExecID and then booked the fill **only if the order was already in `context`**. There was no other branch, so a fill for an untracked order was dropped in full: no `Fill`, no execution detail, no position update, and nothing in the log to say a fill had been seen.
- The ExecID had already been consumed by then, which made the loss permanent — the replay the gateway pushes after a reconnect, the thing that would have recovered it, is rejected as a duplicate.
- An order is missing from `context` in ordinary ways: it reached a terminal status and was removed (the cancel/fill race — cancel ack processed first, fill arrives to find nothing), or it was placed from another client or an earlier session. The recovery insert does not cover those; it only fires for a New/New report with a contract id and a quantity.
- What is left is a position the account holds and the engine does not, disagreeing with the gateway's own position feed, with no error either way.

Closes #314.

## Change

The execution report carries what booking needs, so the fill is placed from the report when the order is unknown.

- **Both the contract and the side must be on the report.** A guessed side moves the position the wrong way, which is worse than reporting that the fill could not be placed, so a report without tag 54 is refused rather than defaulted.
- **Registration is fallible.** A full instrument table is a condition to report, not one to abort an inbound message on.
- **The ExecID is recorded once the fill can be booked**, not before. An execution that could not be placed stays replayable.

## Replays

On a fresh process the gateway resends prior executions carrying `97=Y` and their original ExecIDs, for orders no session tracks. Booking those would build a position out of history on top of the one the position feed already reports, so this path refuses them. Captured on paper — session one:

```
35=8|34=000108|43=N|11=9001.0|17=00025b49.6a706c96.01.01|150=2|39=2|32=1|31=737.53
```

new process, same account, first messages on the socket:

```
35=8|34=000007|43=N|97=Y|11=9001.0|17=00025b49.6a706c96.01.01|150=2|20=0|39=2|32=1|31=737.53
```

Same ExecID, marked PossResend. Within a process the ExecID window covers the reconnect burst, so the marker is consulted only where the order is unknown.

It does not reach a replay whose order the recovery insert has already tracked — that books as it did before this change. Filed separately as #320, along with #321 for the side default on that same insert.

## Tests

Reverting the untracked booking fails four tests; reordering the ExecID record fails the one that targets it; removing either resend arm fails the replay test; flipping the side mapping or the delta sign fails the position assertions.

`cargo test --lib`: 807 pass. The two `config::expiry_tests` failures are pre-existing and unrelated (timezone data).

## Test plan

### DIFF
--- a/src/engine/context.rs
+++ b/src/engine/context.rs
@@ -894,6 +894,13 @@ impl Context {
         self.market.register(con_id)
     }
 
+    /// Register without panicking when the instrument table is full. Inbound
+    /// message handling must use this — a full table is a condition to report,
+    /// not one to abort the engine on.
+    pub fn try_register_instrument(&mut self, con_id: i64) -> Option<InstrumentId> {
+        self.market.try_register(con_id)
+    }
+
     pub fn set_symbol(&mut self, id: InstrumentId, symbol: String) {
         self.market.set_symbol(id, symbol);
     }
--- a/src/engine/hot_loop/ccp.rs
+++ b/src/engine/hot_loop/ccp.rs
@@ -16,6 +16,52 @@ use crossbeam_channel::Sender;
 
 use super::{HeartbeatState, emit, clone_for_event, parse_price_tag, decode_tif};
 
+/// Where to book a fill whose order this session does not track.
+///
+/// Both the contract and the side come off the report, and both are required:
+/// a guessed side would move the position the wrong way, which is worse than
+/// reporting that the fill could not be placed.
+fn untracked_fill_target(
+    context: &mut Context,
+    parsed: &std::collections::HashMap<u32, String>,
+) -> Option<(InstrumentId, Side)> {
+    // A replayed execution restates history rather than reporting something
+    // new. On a fresh process the gateway resends prior fills with 97=Y and
+    // their original ExecIDs, for orders no session tracks; booking those
+    // would build a position out of the past on top of the one the position
+    // feed already reports. Within a process the ExecID window catches the
+    // reconnect burst, so only the untracked case needs this.
+    let replayed = |tag| parsed.get(&tag).map(|v| v.eq_ignore_ascii_case("Y")).unwrap_or(false);
+    if replayed(97) || replayed(43) {
+        log::debug!("Untracked fill is a replay, leaving the position alone");
+        return None;
+    }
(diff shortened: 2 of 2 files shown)


userFRM:

PR #323, opened 2026-07-29
orders: report the order's cumulative fill on order_status, not the last print (ibx#315)

### LINKED ISSUE #315: orders: order_status reports the last print's size and price as filled and avgFillPrice, so a multi-print order never reports its true average
## Problem

The fill dispatch passes the individual print where the callback's contract asks for the order total:

```rust
wrapper.order_status(
    fill.order_id as i64, status, fill.qty as f64, fill.remaining as f64,
    price_f, perm_id, parent_id, price_f, 0, "", 0.0,
);
```

Argument 3 is `filled` and receives `fill.qty`, the size of this print. Argument 5 is `avgFillPrice` and receives `price_f`, the price of this print. Argument 8, `lastFillPrice`, receives the same value, which is the one place it belongs.

IB's contract is `filled` = cumulative, `avgFillPrice` = volume-weighted across every print of the order, `lastFillPrice` = this print.

## Impact

A 300-share order that fills in three 100-share prints should report

```
filled=100 avg=P1  →  filled=200 avg=(P1+P2)/2  →  filled=300 avg=VWAP
```

and instead reports

```
filled=100 avg=P1  →  filled=100 avg=P2  →  filled=100 avg=P3
```

Any average-cost or P&L calculation reading `avgFillPrice` is wrong whenever an order fills in more than one print, which above a round lot is most of the time. `filled` never reaches the order quantity, so completion detection on that field never fires.

This is every fill on the EClient and Python paths.

## Solution

The gateway already sends both values on each execution report — CumQty (14) and AvgPx (6) — and they are parsed, but only for the execution record, never for the status callback.

Carry them on the fill event so they travel with it, and pass them where the callback expects them. Reading them from the order cache at dispatch time instead would race a second fill arriving between the push and the drain, and report the newer print's cumulative against the older fill.

Where the report omits them, falling back to the print keeps the current behaviour rather than reporting zero.

The order-update path has the matching hole — it passes `0.0` for `avgFillPrice` unconditionally.

userFRM:
> Stacked on #322. The first commit is that PR; review this one from its second commit.

## Summary

- `order_status` was given the size and price of the print that triggered it, where the callback's contract asks for the order as a whole. Argument 3 (`filled`) received the print's quantity and argument 5 (`avgFillPrice`) its price. Argument 8, `lastFillPrice`, received the same price — the one place it belonged.
- IB's contract is `filled` cumulative, `avgFillPrice` volume-weighted across every print, `lastFillPrice` this print.
- A 300-share order filling in three 100-share prints should report `filled=100 avg=P1 → filled=200 avg=(P1+P2)/2 → filled=300 avg=VWAP`. It reported `filled=100` three times, at three different prices each presented as the average.
- Any average-cost or P&L calculation reading `avgFillPrice` is wrong whenever an order fills in more than one print, which above a round lot is most of the time. `filled` never reaches the order quantity, so completion detection on that field never fires.

Closes #315.

## Change

The gateway sends both numbers on every execution report — CumQty (14) and AvgPx (6). They were already parsed, but only into the execution record, never into the status callback.

They now travel on the fill itself. Reading them from the order cache at dispatch time would have been a smaller change, but a second print arriving between the push and the drain would report its cumulative against the earlier fill; carrying them with the event keeps each one self-consistent. A report that omits either field falls back to the print, which is what the callback reported before.

## Every consumer, not just the one the issue named

- **Python compat client** — dispatches the same drained fill through its own `order_status` and had the identical pair.
- **Open-order snapshot** — both clients stored the print as the order's filled quantity, so after a partial fill the push callback said `filled=200` while `reqOpenOrders` said `filled=100` for the same order.

## Tests

One end-to-end assertion pins the whole callback string, so it covers `filled` and `avgFillPrice` together:

```
order_status:42:PartiallyFilled:200:100:150.5
```

Reverting the dispatch produces `...:100:100:151`, which is the reported bug exactly. Two more tests drive an execution report through the handler and assert the drained fill's cumulative pair comes off tags 14 and 6 rather than 32 and 31, and that it falls back to the print when neither is present.

## Test plan

- [x] `cargo test --offline --lib` — 814 passed. The 2 failures are `config::expiry_tests::{named_zone_converts_with_dst, instant_round_trips_to_wire}`, which fail on the base commit too: the host has no legacy timezone files (fixed separately in #336).

### DIFF
--- a/src/api/client/dispatch.rs
+++ b/src/api/client/dispatch.rs
@@ -54,9 +54,12 @@ impl EClient {
             let (perm_id, parent_id) = self.shared.orders.get_order_info(fill.order_id)
                 .map(|info| (info.order.perm_id, info.order.parent_id))
                 .unwrap_or((0, 0));
+            // `filled` and `avgFillPrice` describe the order so far;
+            // `lastFillPrice` describes this print.
+            let avg_price_f = fill.avg_price as f64 / PRICE_SCALE_F;
             wrapper.order_status(
-                fill.order_id as i64, status, fill.qty as f64, fill.remaining as f64,
-                price_f, perm_id, parent_id, price_f, 0, "", 0.0,
+                fill.order_id as i64, status, fill.cum_qty as f64, fill.remaining as f64,
+                avg_price_f, perm_id, parent_id, price_f, 0, "", 0.0,
             );
 
             let side_str = match fill.side {
@@ -102,7 +105,7 @@ impl EClient {
             self.core.push_execution(req_id, c, exec, report);
 
             // Update open order tracking
-            self.core.update_order_fill(fill.order_id, status, fill.qty as f64, fill.remaining as f64);
+            self.core.update_order_fill(fill.order_id, status, fill.cum_qty as f64, fill.remaining as f64);
         }
 
         // Order updates → order_status
--- a/src/api/client/tests.rs
+++ b/src/api/client/tests.rs
@@ -1644,6 +1644,7 @@ fn process_msgs_dispatches_fill() {
         instrument: 0, order_id: 42, side: Side::Buy,
         price: 150 * PRICE_SCALE, qty: 100, remaining: 0,
         commission: PRICE_SCALE, timestamp_ns: 123456789,
+        cum_qty: 100, avg_price: 150 * PRICE_SCALE,
     });
     let mut w = RecordingWrapper::default();
     client.process_msgs(&mut w);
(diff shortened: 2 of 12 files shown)


userFRM:
Four corrections in `<this branch>`.

**Two integration targets stopped compiling.** Adding fields to `Fill` broke twenty struct literals across `tests/scenarios.rs` and `tests/hot_loop_lifecycle.rs`, which `cargo test --lib` never builds. Both compile now, and the gate for this branch covers every target rather than the library alone.

**A debug hook removed.** An `IBX_WIRE_CAPTURE` println had been left in the fill path from wire-capture work; it is out, and the branch is clean of it throughout its history.

**Scope trimmed to the message.** The commit had also carried an unrelated server-tag redesign, byte-identical to the one in #286, under a title about filled quantity. Stripped to what the message claims.

**A misindented field** corrected to rustfmt.

The functional change is unaltered: `order_status` reports cumulative filled quantity rather than the unfilled remainder, and LeavesQty remains correct at both of its production uses.

userFRM:

PR #326, opened 2026-07-29
orders: route an order carrying only conditions through the extended encoder (ibx#325)

### LINKED ISSUE #325: orders: an order whose only extended attribute is conditions has them silently dropped and routes immediately
## Problem

`has_extended_attrs()` decides whether an order routes through the encoder that emits the attribute block. It names thirteen fields, and `conditions` is not among them:

```rust
self.display_size > 0
    || self.min_qty > 0
    || self.hidden
    || self.outside_rth
    ...
    || self.cash_qty > 0.0
```

`attrs()` does copy the conditions into `OrderAttrs`. They are simply never reached: a DAY order carrying only conditions takes the plain submit branch, whose encoder emits no condition tags at all. They are copied and then thrown away.

The same gap covers `conditions_cancel_order`, `conditions_ignore_rth` and `oca_type` — all four are carried by `attrs()` and none are named by the predicate.

## Impact

A conditional order becomes an unconditional one. Place a marketable DAY limit whose only extra attribute is a price condition that is currently false — "trigger only if SPY >= 999" — and the `35=D` carries no condition tags. The gateway routes it immediately and it fills.

The correct behaviour is that the order sits inactive until the condition is met. Instead you take the position now, at the current price, in the direction you explicitly said you were not ready to trade. There is no error and nothing in the message to show a condition was ever requested.

Setting any one of the thirteen listed attributes, or a TIF other than DAY, makes it behave correctly — which is why a test using a realistic attribute bundle will not see it.

Reachable from `EClient::place_order` and the Python client. Not from `Context::submit_*_ex`, which takes `OrderAttrs` directly and always routes through the shared encoder.

## Solution

Name every field `attrs()` carries. The two have to agree: anything copied into `OrderAttrs` but not named by the predicate is dropped without a word.

A test with one entry per carried attribute keeps them agreeing — setting each field alone and asserting the order routes through the extended encoder. Adding a field to `attrs()` without adding it to the predicate is the failure this guards against, and it is silent every time.

userFRM:
## Summary

- `has_extended_attrs()` decides whether an order reaches the encoder that emits the attribute block. It named thirteen fields, and `conditions` was not one of them — nor were `conditions_cancel_order`, `conditions_ignore_rth` or `oca_type`.
- All four **are** copied into `OrderAttrs` by `attrs()`. They were simply never reached: the order had already been routed to the plain encoder, which emits no condition tags at all. Copied, then thrown away.
- So a conditional order went out unconditional. A DAY limit whose only extra attribute is a price condition that is currently false carries no condition tags, the gateway routes it immediately, and it fills — the position is taken now, at the current price, in the direction the caller said they were not ready to trade. No error, and nothing in the message showing a condition was ever requested.

Closes #325.

## Why it hides

Setting any one of the thirteen listed attributes, or a TIF other than DAY, makes it behave correctly. Any test using a realistic attribute bundle passes.

Reachable from `EClient::place_order` and the Python client. Not from `Context::submit_*_ex`, which takes `OrderAttrs` directly and always routes through the shared encoder.

## Change

The predicate names everything `attrs()` carries. The two have to agree field for field — anything copied into `OrderAttrs` and not named by the predicate is dropped in silence.

## Test

One entry per carried attribute, each set alone on an otherwise default order. Dropping any of the four added arms fails it. It doubles as the checklist: a field added to `attrs()` without being added to the predicate fails the test rather than shipping.

`cargo test --lib`: 803 pass. The two `config::expiry_tests` failures are pre-existing and unrelated (timezone data).

## Test plan

### DIFF
--- a/src/api/types.rs
+++ b/src/api/types.rs
@@ -479,6 +479,14 @@ impl Order {
             || self.all_or_none
             || self.trigger_method > 0
             || self.cash_qty > 0.0
+            // Everything `attrs()` carries has to be named here, or the order
+            // takes the plain encoder and the attribute is dropped without a
+            // word. Conditions are the costly one: the order goes out
+            // unconditional and routes immediately.
+            || !self.conditions.is_empty()
+            || self.conditions_cancel_order
+            || self.conditions_ignore_rth
+            || self.oca_type > 0
     }
 }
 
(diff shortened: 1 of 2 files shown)


userFRM:
Two corrections after a closer look at what the test actually guarantees.

**The forward guard is now real.** The commit said a field added to `attrs()` without being added to the predicate would fail the test. That was not true — the test is a hand-written list, and a new field would have left it compiling and passing. The test now destructures `OrderAttrs` without a rest pattern, so adding a field there stops it compiling until it is accounted for in both places. Adding a probe field to the struct produces:

```
error[E0063]: missing field `probe_field` in initializer of `types::OrderAttrs`
help: include the missing field in the pattern
```

**The Python client made the `oca_type` arm inert.** `Order::to_api()` copies the two condition flags across but omitted `oca_type`, so an OCA type set from Python fell to the default and was replaced by the gateway's before reaching any of this. One line, and the new arm is reachable from that client.

Also confirmed while checking: `good_after_time` is named by the predicate but never carried — `attrs()` hardcodes it to `0` pending a wire capture (#199). That is pre-existing and deliberate, and the only remaining asymmetry in either direction.

userFRM:
A second omission in the same Python function, found by the same comparison that turned up `oca_type`.

`Order::to_api()` also dropped `trail_stop_price`, so a Python `TRAIL` or `TRAIL LIMIT` order silently lost its initial stop trigger — the field is defined on the Python class and read by the request builder, and simply never crossed. One line, same shape as the `oca_type` fix, and an exhaustive field-by-field comparison confirms nothing else consumed on the outbound path is missing. `conditions` is the intentional exception: `to_api` leaves it empty and the placement call converts it separately.

On the test: `good_after_time` is named by the predicate but is not actually carried — `attrs()` hardcodes it to `0` pending a wire capture (#199) — so that entry pins the routing rather than an emitted tag. Marked as such in the list rather than left to read as though it were carried.

Two reviewers built plain and extended messages for all 21 encoder shapes and compared 210 pairs: every extended tag set is a superset of the plain one, with matching order type, TIF, quantity, price, routing, currency, side and account. The condition arm is the one that changes broker behaviour; the two flag arms and `oca_type` route without emitting anything extra, since the flags only appear inside a non-empty condition block and 6209 only alongside an OCA group. Wasted routes, byte-identical wire.

userFRM:

Issue #317, opened 2026-07-29
ccp: the session-start execution replay is delivered twice and only fills are deduplicated, so every other callback fires twice


userFRM:
## Problem

At session start the gateway replays recent executions. Each replayed message arrives twice — once inside a FIXCOMP envelope and once as plain FIX — carrying the same MsgSeqNum and the same ExecID:

```
WIRE< ccp/comp  8=FIX.4.1|9=000457|35=8|34=000007|43=N|97=Y|11=9001.0|17=00025b49.6a706c96.01.01|150=2|20=0|39=2|32=1|31=737.53
WIRE< ccp/fix   8=FIX.4.1|9=000457|35=8|34=000007|43=N|97=Y|11=9001.0|17=00025b49.6a706c96.01.01|150=2|20=0|39=2|32=1|31=737.53
```

Byte-identical, both reaching the handler. Counting inbound execution reports over one paper session, the replay batch is doubled and this session's own reports are not:

```
   2  35=8|34=000007      ← replay
   2  35=8|34=000009      ← replay
   2  35=8|34=000011      ← replay
   2  35=8|34=000013      ← replay
   1  35=8|34=000118      ← this session
   1  35=8|34=000119
   1  35=8|34=000120
```

The handler has no MsgSeqNum-based deduplication. The only thing preventing the second copy from being acted on is the ExecID window, and that covers the fill block alone — status transitions, the order and contract cache enrichment, and the execution-detail callbacks all run again on the second copy.

The ExecID window is also bounded at 1024 entries. A replay batch larger than the window, or one interleaved with enough live executions, evicts its own earlier entries and the fill block itself becomes reachable twice.

## Impact

Duplicated callbacks for the same execution. Anything counting callbacks rather than reconciling state sees each replayed execution twice.

The narrow escape is that these carry `97=Y`, so a handler that consulted the resend marker would reject the second copy — but neither `97` nor `43` is read anywhere in the handler today, so nothing does.

## Solution

Deduplicate on MsgSeqNum for inbound messages on this socket. FIX carries the field for exactly this, it is already parsed, and it covers every message type rather than only the ones that happen to have an ExecID.

Consulting the PossResend marker (97) would also cover this particular case, and is worth doing regardless — a replayed execution restates history rather than reporting something new, and a fresh process has no ExecID memory at all.

userFRM:
Closing this to clear the tracker. #409 carries the fork's main if the detail here is still useful.

deepentropy:
[...]

PR #331, opened 2026-07-29
ccp: let a duplicate execution suppress the fill, not the whole report (ibx#330)

### LINKED ISSUE #330: ccp: a duplicate ExecID returns out of the whole execution handler, so a replayed final fill leaves the order open forever
## Problem

Deduplication exists to stop a fill being counted twice. It is implemented as an early return from the entire handler:

```rust
    return;
}
```

That `return` skips everything the report still had to do — the order and contract cache update, and the terminal block that pushes the completed order and calls `context.remove_order`.

## Impact

A replayed execution carrying a terminal status never retires the order. It stays in `open_orders` indefinitely: `req_open_orders` keeps returning an order that is filled, and no completed-order record is ever produced for it.

This is reachable in ordinary operation, not only in theory. The gateway replays recent executions at session start, and #317 shows each replayed message is currently delivered twice within the same session — so the second copy of a final fill hits this path with the order still open.

Stale open-order state is what cancel-all loops and re-entry guards read. An order that shows as working forever is the input that makes those act.

## Solution

Suppress the fill, not the report. A duplicate should skip the booking — the position update, the `Fill` event, the execution record — and let the status transition and terminal bookkeeping run as they would for any other frame. The status has already been applied by the time the fill block is reached, so the state machine is the part that must not be short-circuited.

userFRM:
> Stacked on #323 (which is stacked on #322). Review this one from its third commit.

## Summary

- Deduplication exists to stop a fill being counted twice. It was implemented as an early `return` from the entire handler, so everything after the fill block went with it — the order and contract cache update, and the terminal block that pushes the completed order and calls `context.remove_order`.
- A replayed execution carrying a terminal status therefore never retired its order. It stayed in `open_orders` indefinitely: `req_open_orders` kept returning an order that was filled, and no completed-order record was produced.

Closes #330.

## Reachable in ordinary operation

The gateway replays recent executions at session start, and #317 shows each of those messages is currently delivered twice within the same session — so the second copy of a final fill lands here with the order still open.

Stale open-order state is what cancel-all loops and re-entry guards read. An order that shows as working forever is the input that makes those act.

## Change

A duplicate skips only the booking — the position update, the `Fill` event, the execution record. The rest of the report is processed as it would be for any other frame. The status has already been applied by the time the fill block is reached, so the state machine is the part that must not be short-circuited.

## Test

A partial fill is booked, then the same ExecID is replayed carrying the terminal status. The fill is not counted twice, and the order still reaches its terminal state and is reported completed. Restoring the early `return` fails it.

## Test plan

- [x] `cargo test --offline --lib` — 817 passed. The 2 failures are `config::expiry_tests::{named_zone_converts_with_dst, instant_round_trips_to_wire}`, which fail on the base commit too: the host has no legacy timezone files (fixed separately in #336).

### DIFF
--- a/src/api/client/dispatch.rs
+++ b/src/api/client/dispatch.rs
@@ -54,9 +54,12 @@ impl EClient {
             let (perm_id, parent_id) = self.shared.orders.get_order_info(fill.order_id)
                 .map(|info| (info.order.perm_id, info.order.parent_id))
                 .unwrap_or((0, 0));
+            // `filled` and `avgFillPrice` describe the order so far;
+            // `lastFillPrice` describes this print.
+            let avg_price_f = fill.avg_price as f64 / PRICE_SCALE_F;
             wrapper.order_status(
-                fill.order_id as i64, status, fill.qty as f64, fill.remaining as f64,
-                price_f, perm_id, parent_id, price_f, 0, "", 0.0,
+                fill.order_id as i64, status, fill.cum_qty as f64, fill.remaining as f64,
+                avg_price_f, perm_id, parent_id, price_f, 0, "", 0.0,
             );
 
             let side_str = match fill.side {
@@ -102,7 +105,7 @@ impl EClient {
             self.core.push_execution(req_id, c, exec, report);
 
             // Update open order tracking
-            self.core.update_order_fill(fill.order_id, status, fill.qty as f64, fill.remaining as f64);
+            self.core.update_order_fill(fill.order_id, status, fill.cum_qty as f64, fill.remaining as f64);
         }
 
         // Order updates → order_status
--- a/src/api/client/tests.rs
+++ b/src/api/client/tests.rs
@@ -1644,6 +1644,7 @@ fn process_msgs_dispatches_fill() {
         instrument: 0, order_id: 42, side: Side::Buy,
         price: 150 * PRICE_SCALE, qty: 100, remaining: 0,
         commission: PRICE_SCALE, timestamp_ns: 123456789,
+        cum_qty: 100, avg_price: 150 * PRICE_SCALE,
     });
     let mut w = RecordingWrapper::default();
     client.process_msgs(&mut w);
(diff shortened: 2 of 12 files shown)


userFRM:
Strengthened the test to assert the status notification, not just the removal.

The property this change turns on is that a duplicate still lets the terminal status reach the application. Treating the duplicate as though it had booked a fill would swallow that, because the status notification is suppressed whenever a fill was reported instead — and the original test could not tell the difference, since it only checked that the order had been removed.

It now pins exactly one `OrderUpdate` carrying `Filled` for the right order, plus the completed-order record, so suppressing the notification fails it.

The wider "and nothing else" surface — the rich-order cache write, terminal removal for `Cancelled` and `Rejected`, the `OrderUpdate` and `CompletedOrder` payloads — is untested, but all of it predates this change and none of it is what this commit alters, so I have left it alone rather than growing the diff.

userFRM:
Closed four mutations the tests were letting through, one of which was the whole point of the change.

**The position was never asserted.** Making the duplicate arm move the position again passed the entire suite — that is precisely the harm deduplication exists to prevent, and nothing caught it. The test now pins the position, the filled quantity on both the status update and the completed-order record, and the event channel, which every other test here passes `None` for. Emitting a second `Fill` to a user callback was invisible for the same reason.

**The empty-ExecID path had no coverage anywhere in the repo.** Dropping the `!exec_id.is_empty()` short-circuit records the empty string once and then silently suppresses every later fill that arrives without an id — real fills, dropped as duplicates. Two frames with no ExecID now have to book twice.

**A late duplicate could put a finished order back on the open list.** The enrichment block had no guard against a stale status, so a duplicate of an earlier partial arriving after the terminal rewrote the cache to `PartiallyFilled`, and `req_open_orders` reported a completed order as working until the terminal's own duplicate restored it. Self-healing given ordinary delivery shapes, but visible to a polling thread in between. The block now refuses to overwrite a terminal status with an open one.

Filed #344 for the one finding that is not this change's to fix: the ExecID window is 1024 entries, so a replay batch deeper than that double-books its head. The same probe behaves identically before this commit, so it is the bound rather than anything here — but #317's double delivery makes it easier to reach than it looks.

userFRM:

Issue #218, opened 2026-07-10
connection: no reconnect/redirect backoff on CCP and farm (immediate re-dial, 3-attempt cap)


deepentropy:
### Current behavior
CCP and farm reconnects spawn an immediate re-dial with no delay and a 3-attempt cap: `src/engine/hot_loop/mod.rs:779-799` (CCP) and `:720-746` (farm). Redirects re-dial immediately with `depth+1` and no backoff (`src/gateway.rs:673-678`). HMDS uses a different schedule `min(64, 3*2^(n-1))` s with no floor or jitter (`src/engine/hot_loop/mod.rs:1081`).

### Expected behavior
The gateway uses a jittered backoff ladder `min(2000 + ladder + jitter, 82000)` ms with ladder `{0,5000,15000,30000,50000,60000}`, jitter growing `5000 -> 20000` (+5000 per failure), ladder index incremented per failure and capped, and an adaptive socket read timeout clamped to 20-30 s.

### Impact
Immediate rapid-fire reconnects against live IB servers risk server-side rate-limiting or connection eviction, and the tight cap gives up sooner than the gateway would.

### Suggested fix
Implement the jittered backoff ladder for CCP/farm/redirect paths with the documented floor, ceiling, and per-failure growth.

deepentropy:
Fixed in 42251f9, implementing the ladder you documented: CCP and farm reconnects are scheduled on delay = min(2s floor + ladder + jitter, 82s), with the ladder stepping 0/5/15/30/50/60s per consecutive failure and the jitter range growing 5s to 20s (+5s per failure). Attempts are scheduled from the hot loop and the counter resets on success.

The 3-attempt hard cap is gone: a Disconnected notification still fires after three straight failures so consumers learn about the outage, but retries continue on the ladder indefinitely instead of giving up sooner than the gateway would. The reconnect-path redirect gets the same 2s floor before following (the initial-connect redirect stays immediate — that is the server directing the login, not a failure loop).

Two scope notes: the historical connection keeps its existing doubling schedule, which was matched to a capture rather than the documented ladder; and the adaptive 20-30s socket read timeout from your expected-behavior section was not taken on — the sockets are non-blocking in the hot loop, so there is no blocking read to clamp. Shout if either matters in practice.

Validated live: forced farm disconnect reconnects under the new scheduler; ladder bounds are unit-tested per failure count.

PR #333, opened 2026-07-29
orders: keep the sign when a price has no whole part (ibx#332)

### LINKED ISSUE #332: orders: format_price drops the minus sign for any price between -1 and 0, turning a credit into a debit
## Problem

`format_price` builds the integer and fractional parts separately:

```rust
let whole = price / PRICE_SCALE;
let frac = (price % PRICE_SCALE).unsigned_abs();
let mut s = StackStr::new();
s.write_i64(whole);
```

Integer division truncates toward zero, so any price strictly between -1 and 0 has `whole == 0`. The fraction has already had its sign stripped by `unsigned_abs`. Nothing writes the minus, and `-0.35` is emitted as `0.35`.

Outside that band the sign survives, because it lives in the whole part: `-1.5` formats correctly. It is only the interval where the whole part is zero that loses it.

## Impact

The emitted price is the correct magnitude with the wrong sign — a credit stated as a debit, or the reverse. That is not a malformed value the gateway can reject; it is a valid price for the opposite side of the trade.

`format_price` is the single price formatter on the order path, used at 67 call sites — limit and stop prices, bracket entry and exit legs, trailing amounts and limit offsets, adjusted stop and stop-limit prices, discretionary amount and cash quantity.

Prices in that band are ordinary wherever a net price can be negative: a spread or combination quoted as a net credit under a dollar sits exactly there, and small negative offsets and adjustments do too. Order submission is currently restricted to `STK`, which keeps most of those out of reach for now, but the formatter is shared and the restriction is not what makes it correct.

## Solution

Write the sign when the whole part cannot carry it:

```rust
if price < 0 && whole == 0 {
    s.push(b'-');
}
```

The band either side of the boundary is worth pinning in a test — `-1.5`, `-1.0`, `-0.35`, `-0.01`, `0`, and the positive cases — since the failure is confined to values whose integer part rounds away.

userFRM:
## Summary

- `format_price` writes the integer and fractional parts separately. Integer division truncates toward zero, so any price strictly between -1 and 0 has a whole part of `0`, and `unsigned_abs` has already stripped the sign from the fraction.
- Nothing wrote the minus. `-0.35` went out as `0.35`.
- Outside that band the sign survives, because it lives in the whole part — `-1.5` was always correct. Only the interval where the integer rounds away was wrong.

Closes #332.

## Why it matters

The result is the right magnitude with the wrong sign — a credit stated as a debit. There is nothing malformed for the gateway to reject; it is a valid price for the opposite side of the trade.

This is the single price formatter on the order path, reached from 67 call sites: limit and stop prices, bracket entry and exit legs, trailing amounts and limit offsets, adjusted stop and stop-limit prices, the discretionary amount and the cash quantity.

Prices in that band are ordinary wherever a net price can be negative — a spread quoted as a net credit under a dollar sits exactly there. Submission is restricted to `STK` today, which keeps most of them out of reach, but the formatter is shared and that restriction is not what makes it correct.

## Test

Pins both sides of the boundary: `-1.5`, `-1.0`, `-0.35`, `-0.01`, `0`, and the positive cases. Dropping the guard fails it with `0.35` where `-0.35` was expected.

## Test plan

### DIFF
--- a/src/engine/hot_loop/mod.rs
+++ b/src/engine/hot_loop/mod.rs
@@ -1299,6 +1299,12 @@ pub(crate) fn format_price(price: Price) -> StackStr {
     let whole = price / PRICE_SCALE;
     let frac = (price % PRICE_SCALE).unsigned_abs();
     let mut s = StackStr::new();
+    // Division truncates toward zero, so a price in (-1, 0) has a whole part
+    // of 0 and carries its sign only in the fraction. Writing the integer
+    // alone would turn -0.35 into 0.35.
+    if price < 0 && whole == 0 {
+        s.push(b'-');
+    }
     s.write_i64(whole);
     if frac != 0 {
         s.push(b'.');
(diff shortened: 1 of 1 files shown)


userFRM:
Fixed a placement problem this change had exposed.

The file already carried a stray duplicate `#[test]` stacked above `reconnect_backoff_ladder_bounds`, so that test was registered twice on `main`. My test landed between the stray attribute and the function it belonged to, which moved the double registration onto the new test — it ran twice — and left the ladder's `#218` comment sitting over the wrong function.

Removed the stray attribute and put the comment back with its function. The new test is registered once, `reconnect_backoff_ladder_bounds` is no longer duplicated either, and the `duplicated attribute` warning that was present on `main` is gone.

userFRM:
The test could not see the smallest tick.

Every case I wrote went through an `f64` — `-0.35`, `-0.5`, `-0.01` — so none of them reached one raw fixed-point unit. A guard that fires slightly too late still formats `-0.00000001` as positive and passed the whole suite. Raw-integer cases now pin the boundary itself: one tick either side of `-1`, the boundary, and zero. Three mutations fail on them — a guard that starts one tick late, one that skips the smallest fraction, and one that also fires at zero.

Independent verification worth recording, since it is more than I ran: formatted values were exact-parsed back to their original integer for a million deterministic inputs, exhaustively around `-1`, `0` and `+1`, and **all 99,999,999 values in the affected interval** were checked for sign. The extremes format correctly too — `i64::MIN` gives `-92233720368.54775808`.

A bare `-0` is unreachable: the price type is an integer with no signed zero, and anything below one tick truncates to `0` and formats as `"0"`.

userFRM:

Issue #334, opened 2026-07-29
orders: a replace restates almost none of the order — modifying a trailing stop drops its trail and the order goes Inactive


userFRM:
## Problem

The `35=G` field list is `35, 52, 11, 41, 44, 1, 6122, 6433, 38, 54, 40, 55, 167, 6035, 59, 6008, 6088, 6211, 6238`, plus `99` when the tracked order has a stop price.

There is no attribute block on it at all — no display size, no hidden flag, no OCA group or type, no parent link, no conditions, no algo, no trailing tags, no GTD expiry. It also omits TransactTime (60), which every `Submit*` sends.

Those attributes are not merely unsent. `OrderRequest::Modify` carries `{ new_order_id, order_id, price, qty }`, so they are not reachable at encode time.

## Impact

Observed on paper. A trailing stop, submitted:

```
35=D  ...|55=SPY|54=2|38=1|40=P|99=100|211=100|18=a|59=0|60=20260729-15:16:08|167=STK
```

`40=P` with the trail amount in both 99 and 211, and `18=a` marking it a trailing stop. Then re-placing the same id with only the quantity changed:

```
35=G  ...|11=9501.1|41=9501.0|44=0|6122=c|6433=1|38=2|54=2|40=P|55=SPY|167=STK|6035=SPY|59=0|6008=756733
```

The order type survives as `40=P`, and everything that made it a trailing stop does not: no 99, no 211, no `18=a`. What is left describes a pegged order with no offset. The gateway's answer:

```
[status] oid=9501 Inactive
```

So changing the size of a trailing stop leaves you without one. The order is not adjusted and not left alone — it is knocked out, and the only signal is a status the caller has to be watching for. For a protective trailing stop that is the entire purpose of the order.

The same shape applies to every other attribute a replace drops: an iceberg stops being hidden, an order in an OCA group stops being linked, a bracket child stops being parented. Where the gateway accepts the reduced message instead of rejecting it, the order stays on the book having quietly become something else, which is the worse outcome of the two.

## Solution

A replace has to restate the order. That means carrying the attributes on the modify request so they are reachable at encode time, and emitting the same attribute block the submit path already builds, rather than a hand-listed subset.

TransactTime is worth adding in the same pass — FIX requires it on a replace, and a session-level reject naming tag 60 is currently log-only.

[...]

PR #337, opened 2026-07-29
orders: move the trigger when a stop order is modified (ibx#324)

### LINKED ISSUE #324: orders: modifying a stop order sends LmtPrice 0 and the original stop, so the gateway rejects it and the order goes Inactive
## Problem

`place_order` on an already-tracked id emits a Modify, and takes the price from one field:

```rust
let price = (order.lmt_price * PRICE_SCALE_F) as i64;
```

For a stop order the price lives in `aux_price`. `lmt_price` is zero and `aux_price` is discarded, so the modify carries `44=0` and the stop price it sends is the one the order already had.

The builder compounds it: the modify price is always written to tag 44, and tag 99 is filled from the *original* order's stop. There is no path by which a stop's trigger can move.

## Impact

Observed on paper against the live gateway. A resting stop:

```
35=D|11=9202.0|55=SPY|54=2|38=1|40=3|99=600|59=0|167=STK|100=BEST
```

then `place_order` on the same id with `aux_price: 610.00`:

```
35=G|11=9202.1|41=9202.0|44=0|6122=c|6433=1|38=1|54=2|40=3|55=SPY|167=STK|6035=SPY|59=0|6008=756733|6088=Socket|99=600
```

`44=0` where the caller asked for 610, and `99=600` — the original trigger, unchanged. The gateway rejects it:

```
[status] oid=9202 Inactive
[error] code=202  Order 9202 cancel/modify rejected (reason: 1)
```

So the order does not merely fail to move. It ends up **Inactive** — a stop the caller was trying to adjust is gone, and the only signal is a rejection code on an order they believe they just repositioned. For a protective stop that is the whole point of the order.

`Context::modify` has the same shape from the other direction: it takes a single price, which is written to 44 regardless of order type. For a plain stop there is no limit leg, so that price lands on a field the order does not use while the trigger stays where it was.

The same capture shows `6433=1` on an order submitted without it, which is #312.

## Solution

Two parts, both needed:

- **Carry the stop price on the modify request.** `OrderRequest::Modify` has one price field; it needs the trigger as well, and `place_order` should populate it from `aux_price` the way the submit path does.
- **Route each price to the tag its order type uses.** The builder already reads `ord_type` for tag 40. A trigger-only type — plain stop, stop with protection, market-if-touched — has no limit leg, so its price belongs in 99 and 44 should not be sent at all. Types with both legs keep 44 for the limit and take the new trigger in 99.

For `Context::modify`, routing by the tracked order's type gives the caller what the single price argument can only have meant: on a stop order, the stop.

userFRM:
> Stacked on #319. Review this one from its second commit.

## Summary

- A modify carried one price and the builder always wrote it to tag 44. Tag 99 was filled from the **original** order's stop, so no sequence of calls could move a trigger.
- The public clients made it worse by reading the new price from `lmt_price` alone. A stop order carries its price in `aux_price`, so the field the caller set was discarded and the replace asked for a limit price of zero on an order with no limit leg.

Closes #324.

## The gateway rejects it

Placing a SELL STP at 600, then re-placing the same id with `aux_price: 610.00`:

```
35=D  ...|55=SPY|54=2|38=1|40=3|99=600|59=0|167=STK|100=BEST
35=G  ...|41=9202.0|44=0|6122=c|38=1|54=2|40=3|55=SPY|167=STK|59=0|6008=756733|99=600
```

`44=0` where the caller asked for 610, and `99=600` — the original trigger, unchanged. The answer:

```
[status] oid=9202 Inactive
[error] code=202  Order 9202 cancel/modify rejected (reason: 1)
```

So the trigger does not merely stay put. The stop the caller was repositioning ends up **Inactive** — gone — and the only signal is a rejection code on an order they believe they just moved. For a protective stop that is the entire purpose of the order.

## Change

`Modify` carries the trigger as well as the price, and both clients populate it from `aux_price` the way their submit paths already do.

Each price then goes to the tag its order type uses:

| type | 44 | 99 |
|---|---|---|
| trigger-only — STP, STP PRT, MIT | omitted | the new trigger |
| two-legged — STP LMT, LIT | the limit | the new trigger, or the one it had |
| everything else | the limit | untouched |

Trigger-only types send no 44 at all, which is the shape their own submit has. The two-legged bucket applies only where the order actually has a trigger: `b'K'` is Limit-if-Touched **and** Market-to-Limit, and the tracked trigger is what separates them. Every other type keeps the shape it had, so a stray aux price on a limit, or the offset on a pegged order, cannot arrive as a tag 99.

The replacement records the trigger the modify just asked for rather than the one it replaced, so a second modify restates the current price instead of resurrecting the previous one.

## Limit orders are unchanged

A plain limit modify emits exactly what it did before — same tags, same order, same bytes.

## Tests

Fourteen over the builder, driven through the real send path and reading back what was written. Each of these fails them: dropping MIT or STP PRT from the trigger-only set, ignoring a supplied trigger, shifting a supplied trigger, erasing the replacement's trigger, reverting the client's aux-price propagation, and reverting the routing.

## Not fixed here

A replace still restates almost nothing else about the order — no attribute block at all. Modifying a **trailing** stop drops its trail and the order goes Inactive the same way; that is #334, and it is not addressed by this change, since a trailing stop is a pegged type rather than a trigger type.

---

**Stacked on #249.** The first of the two commits here is that PR; this branch needs its `outside_rth` field on the replace request, and the two edit the same tag list. Review the second commit for this change, and merge #249 first.

## Test plan

- [x] `cargo test --offline --lib` — 811 passed. The 2 failures are `config::expiry_tests::{named_zone_converts_with_dst, instant_round_trips_to_wire}`, which fail on the base commit too: the host has no legacy timezone files (fixed separately in #336).
- [x] Mutation check: reverting the trigger routing fails `modifying_a_stop_moves_its_trigger` and `every_trigger_only_type_moves_its_trigger`; reverting the outside-RTH conditional fails `modify_emits_outside_rth_only_when_the_caller_set_it`.

### DIFF
--- a/src/api/client/orders.rs
+++ b/src/api/client/orders.rs
@@ -38,6 +38,7 @@ impl EClient {
                 order_id: oid,
                 price,
                 qty,
+                outside_rth: order.outside_rth,
             })
         } else {
             ClientCore::build_order_request(order, oid, instrument)?
--- a/src/api/client/orders.rs
+++ b/src/api/client/orders.rs
@@ -33,12 +33,17 @@ impl EClient {
         let cmd = if self.core.is_order_tracked(oid) {
             let price = (order.lmt_price * PRICE_SCALE_F) as i64;
             let qty = order.total_quantity as u32;
+            // A stop's trigger rides on aux_price, exactly as it does on the
+            // submit path. Reading only lmt_price left a stop order modifying
+            // itself to a limit price of zero.
+            let stop_price = (order.aux_price * PRICE_SCALE_F) as i64;
             ControlCommand::Order(OrderRequest::Modify {
                 new_order_id: oid,
                 order_id: oid,
                 price,
                 qty,
                 outside_rth: order.outside_rth,
+                stop_price,
             })
         } else {
             ClientCore::build_order_request(order, oid, instrument)?
--- a/src/api/client/tests.rs
+++ b/src/api/client/tests.rs
@@ -31,6 +31,31 @@ fn spy() -> Contract {
 //  Algo parsing
 // ═══════════════════════════════════════════════════════════════════
 
+/// Re-placing a tracked id is a modify, and a stop order's price lives in
+/// `aux_price`. Reading only `lmt_price` sent a limit price of zero for an
+/// order that has no limit leg, which the gateway rejects outright.
+#[test]
+fn modifying_a_stop_carries_the_new_trigger() {
+    let (client, rx, _shared) = test_client();
+    let stop = Order {
+        action: "SELL".into(), total_quantity: 1.0, order_type: "STP".into(),
+        aux_price: 600.0, tif: "DAY".into(), ..Default::default()
+    };
+    client.place_order(9201, &spy(), &stop).unwrap();
+    rx.try_recv().expect("the submit");
+
+    let moved = Order { aux_price: 610.0, ..stop };
+    client.place_order(9201, &spy(), &moved).unwrap();
+
+    match rx.try_recv().expect("the modify") {
+        ControlCommand::Order(OrderRequest::Modify { stop_price, .. }) => assert_eq!(
+            stop_price, (610.0 * PRICE_SCALE_F) as i64,
+            "the new trigger must reach the request",
+        ),
+        other => panic!("expected a Modify, got {other:?}"),
+    }
+}
+
 #[test]
 fn parse_algo_vwap() {
     let params = vec![
--- a/src/api/client/tests.rs
+++ b/src/api/client/tests.rs
@@ -393,6 +393,42 @@ fn place_order_adjustable_trail_carries_trailing_amount_and_unit() {
 }
 
 #[test]
+fn modify_carries_outside_rth_from_the_resubmitted_order() {
+    // ibx#247: the replace asserted 6433=1 unconditionally, so an order placed
+    // with outside_rth=false came back outside-RTH after any modify. The flag
+    // has to travel with the modify, since the tracked record has no field for
+    // it.
+    let (client, rx, shared) = test_client();
+    shared.market.set_instrument_count(1);
+    let order = Order {
+        action: "BUY".into(), total_quantity: 1.0, order_type: "LMT".into(),
+        lmt_price: 100.0, outside_rth: false, ..Default::default()
+    };
+    client.place_order(70, &spy(), &order).unwrap();
+    let _submit = rx.try_recv().unwrap();
+
+    // Same id -> modify. Caller still says outside_rth=false.
+    let reprice = Order { lmt_price: 101.0, ..order.clone() };
+    client.place_order(70, &spy(), &reprice).unwrap();
+    match rx.try_recv().unwrap() {
+        ControlCommand::Order(OrderRequest::Modify { outside_rth, .. }) => {
+            assert!(!outside_rth, "a modify must not opt the order into the extended session");
+        }
+        cmd => panic!("expected Modify, got {:?}", cmd),
+    }
+
+    // And it survives when the caller does want it.
+    let rth_out = Order { lmt_price: 102.0, outside_rth: true, ..order.clone() };
+    client.place_order(70, &spy(), &rth_out).unwrap();
+    match rx.try_recv().unwrap() {
+        ControlCommand::Order(OrderRequest::Modify { outside_rth, .. }) => {
+            assert!(outside_rth, "an explicit outside_rth=true must reach the replace");
+        }
+        cmd => panic!("expected Modify, got {:?}", cmd),
+    }
+}
+
+#[test]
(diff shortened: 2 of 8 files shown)


userFRM:
Two corrections, both about claims rather than behaviour.

**The comment overstated what the guard does.** It said the offset on a pegged order cannot arrive as a tag 99. That holds for a trigger supplied on the *request*, which is what the guard covers — but a pegged or relative order tracks its offset in `stop_price`, so its replace does restate that on 99, exactly as it did before this change. Reworded, along with the test comment that claimed those types are submitted with no trigger. The wire behaviour is unchanged either way; it is one of the reasons those types are refused a modify outright in #342.

**The bucket was pinned from below but not above.** Adding `b'U'` to the trigger-only set passed the whole suite — nothing asserted that a type *outside* the set keeps its limit leg. Now covered, and that mutation fails.

Verification from the review worth recording: every `ord_type` byte the codebase can produce was enumerated and placed in a bucket, and the `b'K'` split is correct for the right reason — Market-to-Limit always tracks a zero trigger and Limit-if-Touched tracks a real one, so the tracked trigger is what separates the two kinds sharing that byte. Ten before-and-after scenarios were byte-compared, with no regression outside the intended deltas.

One gap this does not close, noted for a follow-up: the CCP recovery insert stores only the first byte of tag 40, so a recovered multi-character type — `LT`, `SP`, `TSL`, `MIDPX` — lands outside every bucket and renders as `40=2`. A recovered stop-with-protection modify therefore still produces the invented tag 44 that this PR removes for the natively-tracked case. Pre-existing and byte-identical at base, but it is the same failure reached through the recovery door.

userFRM:
Rebased. This branch was stacked on the OutsideRTH commit from #319, and I have closed that PR in favour of #249 — the two took opposite positions on where the flag comes from, and #249's is the one that matches how a modify works in this API. Leaving this branch where it was would have shipped the rejected approach underneath a stop-trigger fix.

So the stack is now #249 → this. The trigger change itself is unaltered in behaviour; what changed is what it sits on:

- The fourteen-argument `build_replace_fields` extraction is gone. The replace tag list is back to an inline vector with two conditional pushes, which is what it needs to express "no limit leg" and "no extended-hours flag". I was not going to upstream a function that wide behind a `too_many_arguments` allow.
- `is_trigger_only` stays — that is the actual content of the fix.
- The replacement records the moved trigger, but no longer carries an `outside_rth` field on the tracked order, because #249 takes the flag from the resubmitted order instead of storing it.
- The tests that exercised the stored field are gone with it. The six trigger tests remain and now share one socket helper with the outside-RTH test rather than each rebuilding the loopback pair.

Mutation-verified after the rebase, each guard reverted on its own:

| Reverted | Fails |
|---|---|
| always emit tag 44 | `modifying_a_stop_moves_its_trigger`, `every_trigger_only_type_moves_its_trigger` |
| restate the old trigger | those two plus `a_supplied_trigger_moves_a_two_legged_order`, `the_replacement_keeps_the_trigger` |
| assert 6433 unconditionally | `modify_emits_outside_rth_only_when_the_caller_set_it` |

811 passed, plus the two `config::expiry_tests` failures that are missing legacy timezone files rather than anything on this branch.

One thing found while mutating, which belongs to neither this PR nor #249: making the **submit** paths emit `6433` unconditionally fails no test at all. The submit-side assertions check that the flag is present when it was asked for and never that it is absent when it was not, so they cannot catch a regression in that direction — the same one-sided shape that let this bug live on the replace path. Worth fixing where those tests live rather than widening this branch.

userFRM:
The rebase onto #249 was verified hunk for hunk against the previous tip: same eight files, five byte-identical, and the two that differ do so mechanically because the new base already carries `outside_rth` on the modify literals. The classification block and `is_trigger_only` are character-identical.

The dropped fourteen-argument extraction leaves no trace — the helper is absent tree-wide with no dangling callers, and the inline version preserves the exact tag order including both conditional positions. Dropping `.with_outside_rth(...)` is the correct adaptation rather than a loss: the new base's order record has no such field, because #249 takes the flag from the resubmitted request instead of storing it, and the base's own polarity test survived the test-module rewrite and passes.

All seven new or changed tests fail by name under their targeted reversions, including the two negative guards: dropping the `orig_stop != 0` condition fails `a_type_without_a_trigger_never_gains_one` with a trigger leaking onto a market-to-limit order, and killing the else-restatement fails `modifying_a_stop_limit_moves_the_limit_and_keeps_the_trigger`.

The claim that a trigger-only replace matches "the shape its own submit has" traces to both stop submit paths, which emit the order type and trigger with no tag 44 — and `LIT` versus `MTL` are distinguished exactly by the tracked trigger, which is what the two-legged branch relies on.

811 pass plus the two `config::expiry_tests` timezone failures. All five offline targets clean.

userFRM:

PR #340, opened 2026-07-29
orders: put the extended attributes on the tags the gateway reads (ibx#338, ibx#339, ibx#327)

### LINKED ISSUE #327: orders: the two condition flags are not symmetric on the wire — one sets a third field the other does not, suggesting they are crossed
## Problem

The two condition flags are emitted as:

```rust
if attrs.conditions_cancel_order {
    fields.push((6128, "1".to_string()));
}
if attrs.conditions_ignore_rth {
    fields.push((6151, "1".to_string()));
}
```

Setting each flag alone and reading back what the gateway reports for the order does not give a symmetric result. Four conditional orders, identical apart from the two flags, on paper:

| flags set | sent | gateway reports 6128 / 6151 / 6579 |
|---|---|---|
| neither (control) | — | `0 / 0 / 0` |
| `conditions_cancel_order` | `6128=1` | `1 / 0 / 0` |
| `conditions_ignore_rth` | `6151=1` | `0 / 1 / **1**` |
| both | `6128=1`, `6151=1` | `1 / 1 / 1` |

Tag **6579 is never sent by ibx and never decoded by it**, yet the gateway reports it set whenever `conditions_ignore_rth` is used, and clear whenever only `conditions_cancel_order` is. It tracks 6151 exactly and is untouched by 6128.

So one of the two flags has a side effect on a third field the other does not have. Two flags that are supposed to be independent booleans on the same condition block should not behave differently that way.

## Why it matters

The asymmetry is what you would expect if the two were crossed — if the field ibx labels `conditions_ignore_rth` is the one the gateway treats as the submit/cancel behaviour, and the field ibx labels `conditions_cancel_order` is something else.

If that is what is happening, both directions cost:

- an order meant to **cancel** when its condition lapses stays live instead, and fires later at a price the caller no longer wants
- a condition meant to be regular-hours-only evaluates on overnight prints, triggering on a thin quote that reverts by the open

Neither surfaces an error. The order is accepted and rests normally in both cases.

## What would settle it

Placing one conditional order with `conditions_cancel_order = true` and `conditions_ignore_rth = false`, then reading it back in TWS: the two checkboxes say directly which flag was actually set. The table above is everything the wire alone shows.

I have not changed the mapping, since the observation narrows it but does not by itself name the correct tag for each flag.

Reachable from `EClient::place_order` with conditions plus at least one other extended attribute, or `Context::submit_limit_ex`. With conditions as the *only* extended attribute the whole block is dropped first — #325.

### LINKED ISSUE #338: orders: setting displaySize or minQty kills the order — the gateway reports quantity 0 and returns Inactive
## Problem

The extended attribute block emits:

```rust
fields.push((111, format_uint(attrs.display_size as u64).to_string()));
...
fields.push((110, format_uint(attrs.min_qty as u64).to_string()));
```

Neither tag does what the field name says on this dialect. Sending either one makes the gateway reject the order.

## Evidence

Three orders on paper, identical apart from the attribute under test — same contract, side, quantity, price, TIF, all resting far from the market so none can fill:

| order | extra attribute | tag sent | gateway echoes | status |
|---|---|---|---|---|
| 9601 | none (control) | — | `38=200` | **PreSubmitted** |
| 9602 | `display_size: 100` | `111=100` | `38=0` | **Inactive** |
| 9603 | `min_qty: 50` | `110=50` | `38=0` | **Inactive** |

The control rests normally. Both of the others come back with the order quantity reported as **zero** — the quantity was sent correctly as `38=200` in all three — and go Inactive immediately.

So whatever tags 110 and 111 mean here, they are not MinQty and DisplaySize, and including them does not merely fail to hide the order: it corrupts the message enough that the gateway stops reading the quantity and rejects it.

## Impact

An iceberg order does not work. `displaySize` and `minQty` are on the public `Order` struct, reachable from `EClient::place_order`, the Python client, and `OrderAttrs` — and any order that sets either is dead on arrival.

The failure is at least loud: the order goes Inactive rather than resting with its size fully visible. A silently-visible iceberg would be worse, and is what would happen if these tags landed on a field the gateway tolerates.

Note that the whole attribute block is only emitted for an order that routes through the extended encoder, so before #325 an order whose only attribute was one of these was dropped before reaching this — which is why it can have gone unnoticed.

## Solution

Establish what carries display size and minimum quantity in this dialect, and move both. Until then, rejecting these two attributes at the API boundary would be better than emitting them: an order the caller knows was refused beats one that is silently Inactive.

The capture above is repeatable with any resting limit order; the control is the important part, since it shows the rejection tracks the attribute and not the order.

### LINKED ISSUE #339: orders: a trailing-percent order is rejected at every percentage except exactly 1%, which works by coincidence
## Problem

The trailing-percent encoder puts the percentage, in basis points, on tag 6268:

```rust
let pct_str = trail_pct.to_string(); // basis points: 100 = 1%
...
(6268, &pct_str),       // TrailingPercent (basis points)
```

6268 is not a value field. The gateway rejects any percentage other than 1%, and names the tag when it does:

```
35=8|39=8|103=0|11=9701.0|58=Invalid value in field # 6268|40=P
```

## Evidence

Four orders on paper, identical apart from the percentage:

| `trailing_percent` | sent | status |
|---|---|---|
| 0.5 | `6268=50` | **Inactive** |
| **1.0** | `6268=100` | **PreSubmitted** |
| 2.0 | `6268=200` | **Inactive** |
| 10.0 | `6268=1000` | **Inactive** |

Only 1% is accepted, and the reason it is accepted looks like a coincidence: 1% in basis points is `100`, and `100` is the code for "percent" in the unit enum ibx already documents a few hundred lines away, for the adjustable stop:

```rust
// amount (6260) and its unit (6269: 0=amount, 100=percent).
```

So `6268` appears to be a unit selector taking the same enum, and the only percentage that survives is the one whose basis-point value happens to equal the "percent" code. Every other value is an invalid unit.

## Impact

`trailing_percent` is unusable except at exactly 1%. Any other value produces an order that goes Inactive immediately, and a caller who tested at 1% — a natural first thing to try — would see it work and conclude the feature is fine.

That is likely what happened: the comment above the encoder cites a capture as the source for this layout, and a capture taken at 1% would confirm every part of it, including the one field that is wrong.

## Solution

Send the percentage where the percentage belongs, and 6268 as the unit it appears to be. The other trailing tags in the same message are already right — `99` and `211` carry the percent in decimal form and `18=a` marks it trailing — and those are what the accepted 1% order is actually working from.

Worth checking the adjustable-stop path in the same pass: it writes the trailing unit to 6269 and the amount to 6260, which is the mirror-image assignment, and only one of the two can be right.

userFRM:
> Stacked on #319. Review this one from its second commit.

## Summary

Four attributes were emitted on tags that do not carry them here. Each fix below was verified against a live session, with a control order differing only in the attribute under test.

Closes #338.
Closes #339.
Closes #327.

## Display size — 111 → 6103

| order | attribute | gateway echoes | status |
|---|---|---|---|
| control | none | `38=200` | PreSubmitted |
| A | `display_size: 100` on `111` | `38=0` | **Inactive** |

The quantity was sent correctly as `38=200` in both. On 111 the gateway reports it as zero and rejects.

On 6103 the same order is read as an iceberg — the gateway validates the value and says so:

```
REJECT: Display size should be a multiple of lot size
```

and a valid one rests:

```
qty=1000 display_size=200  →  PreSubmitted
```

## Minimum quantity — 110 → refused

Same failure on 110, and nothing in this dialect appears to carry it. It is now refused at the API boundary rather than emitted. That is the outcome the gateway was already producing, minus the round trip, and it says why — reusing the existing `all_or_none is not supported with TRAIL orders` pattern.

## Trailing percent — the value was on the unit tag

`6268` states the unit the trail is expressed in. The percentage, in basis points, was being written to it. Every value was rejected — the gateway names the field:

```
58=Invalid value in field # 6268
```

except exactly **1%**, which survived because 100 basis points collides with the code for "percent". Before and after:

| `trailing_percent` | before | after |
|---|---|---|
| 0.5% | Inactive | **PreSubmitted** |
| 1.0% | PreSubmitted | PreSubmitted |
| 2.0% | Inactive | **PreSubmitted** |
| 10.0% | Inactive | **PreSubmitted** |

The percentage now rides on `9822` as a fraction and `6268` carries the unit. Both encoders are fixed — the shared one for extended submissions had the same assignment.

The comment above that encoder cited a capture as its source. A capture taken at 1% confirms every part of the layout **including** the one field that was wrong, which is how it survived.

## Condition flags — they were on each other's tags

`conditions_ignore_rth` was landing on a field that is not a condition flag at all, which showed as an asymmetry: setting it dragged a third tag along, while `conditions_cancel_order` did not.

| flags set | before → echoed 6128 / 6151 / 6579 | after |
|---|---|---|
| neither | `0 / 0 / 0` | `0 / 0 / 0` |
| `cancel_order` | `1 / 0 / 0` | `0 / 1 / 1` |
| `ignore_rth` | `0 / 1 / 1` | `1 / 0 / 0` |

Each flag now sets exactly the field it names, and neither lands on the price field.

## Tests

Two tests drive the real send path over a loopback pair and read back what was written. Reverting display size to 111, re-crossing the condition flags, or putting the percentage back on 6268 each fail them.

## Test plan

- [x] `cargo test --offline --lib` — 815 passed. The 2 failures are `config::expiry_tests::{named_zone_converts_with_dst, instant_round_trips_to_wire}`, which fail on the base commit too: the host has no legacy timezone files (fixed separately in #336).

### DIFF
--- a/src/api/client/tests.rs
+++ b/src/api/client/tests.rs
@@ -31,6 +31,71 @@ fn spy() -> Contract {
 //  Algo parsing
 // ═══════════════════════════════════════════════════════════════════
 
+/// No tag in this dialect carries a minimum quantity. It used to go out on one
+/// the gateway rejects, which left the order Inactive with its quantity
+/// reported as zero — refusing says why, without the round trip.
+#[test]
+fn min_qty_is_refused_rather_than_sent() {
+    let (client, rx, _shared) = test_client();
+    let order = Order {
+        action: "BUY".into(), total_quantity: 200.0, order_type: "LMT".into(),
+        lmt_price: 100.0, tif: "DAY".into(), min_qty: 50, ..Default::default()
+    };
+    let err = client.place_order(9001, &spy(), &order).expect_err("must be refused");
+    assert!(format!("{err}").contains("min_qty"), "the error names the field: {err}");
+    assert!(rx.try_recv().is_err(), "and nothing reaches the wire");
+}
+
+/// A percentage that quantises to zero basis points would rest as a trailing
+/// stop that never trails.
+#[test]
+fn a_trailing_percent_below_the_wire_resolution_is_refused() {
+    let (client, _rx, _shared) = test_client();
+    let order = Order {
+        action: "SELL".into(), total_quantity: 1.0, order_type: "TRAIL".into(),
+        trailing_percent: 0.004, tif: "DAY".into(), ..Default::default()
+    };
+    let err = client.place_order(9002, &spy(), &order).expect_err("must be refused");
+    assert!(format!("{err}").contains("0.01%"), "the error names the limit: {err}");
+}
+
+/// A percentage that is not a number cannot become one by casting. `as u32`
+/// turns NaN into 0 and infinity into `u32::MAX`, either of which would reach
+/// the wire as a trail nobody asked for.
+#[test]
+fn a_non_finite_trailing_percent_is_refused() {
+    for pct in [f64::NAN, f64::INFINITY] {
+        let (client, _rx, _shared) = test_client();
+        let order = Order {
+            action: "SELL".into(), total_quantity: 1.0, order_type: "TRAIL".into(),
+            trailing_percent: pct, tif: "DAY".into(), ..Default::default()
+        };
+        let err = client.place_order(9004, &spy(), &order)
+            .expect_err("a non-finite percentage must be refused");
+        assert!(
+            format!("{err}").contains("finite") || format!("{err}").contains("0.01%"),
+            "the error explains itself: {err}",
+        );
+    }
+}
+
+/// Basis points are rounded, not truncated: 0.29% is 28.999... in floating
+/// point and would otherwise reach the gateway as 0.28%.
+#[test]
+fn a_trailing_percent_rounds_to_the_nearest_basis_point() {
+    let (client, rx, _shared) = test_client();
+    let order = Order {
+        action: "SELL".into(), total_quantity: 1.0, order_type: "TRAIL".into(),
+        trailing_percent: 0.29, tif: "DAY".into(), ..Default::default()
+    };
+    client.place_order(9003, &spy(), &order).unwrap();
+    match rx.try_recv().expect("the submit") {
+        ControlCommand::Order(OrderRequest::SubmitTrailingStopPct { trail_pct, .. }) =>
+            assert_eq!(trail_pct, 29, "0.29% is 29 basis points, not 28"),
+        other => panic!("expected a trailing-percent submit, got {other:?}"),
+    }
+}
+
 #[test]
 fn parse_algo_vwap() {
     let params = vec![
--- a/src/client_core.rs
+++ b/src/client_core.rs
@@ -1304,6 +1304,13 @@ impl ClientCore {
             ));
         }
 
+        // No tag in this dialect carries a minimum quantity. It used to go out
+        // on one the gateway rejects, which left the order Inactive; refusing
+        // it is the same outcome without the round trip, and says why.
+        if order.min_qty > 0 {
+            return Err("min_qty is not supported".to_string());
+        }
+
         if order.algo_strategy.eq_ignore_ascii_case("Adaptive") {
             return Ok(());
         }
@@ -1336,6 +1343,17 @@ impl ClientCore {
             "STP LMT" | "LIT" if order.aux_price == 0.0 => {
                 return Err(format!(
                     "{} order requires aux_price (stop/trigger price) but got 0.0",
+                    order.order_type
+                ));
+            }
+            // NaN compares false against everything, so it slips past both the
+            // zero check below and the `> 0.0` branch that picks the percent
+            // path — it has to be rejected on its own terms.
+            "TRAIL" | "TRAIL LIMIT"
+                if !order.trailing_percent.is_finite() || !order.aux_price.is_finite() =>
+            {
+                return Err(format!(
+                    "{} order requires finite trailing_percent and aux_price",
                     order.order_type
                 ));
             }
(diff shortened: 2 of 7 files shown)


userFRM:
Both reviews landed on the same two things, and both were right. The retags themselves were confirmed correct and complete — two encoders in agreement, no third one, no other reader or writer of any of the tags, no collisions on 6128 or 6579 — but the surrounding code was not ready for them.

**The percentage was being truncated before it reached the corrected formatter.** `(trailing_percent * 100.0) as u32` drops the fraction, so 0.29% became 28 basis points and went out as `9822=0.0028`. 573 of the 10,000 two-decimal percentages a caller can type were affected. It was harmless before this change only because the gateway rejected every non-1% value outright; correcting the tag is what would have made the wrong trail rest. Rounded now.

**Non-finite input slipped past everything.** `NaN` compares false against every bound, so it passed the zero check, missed the `> 0.0` branch that selects the percent path, and `as u32` would have turned it into 0 — or infinity into `u32::MAX`, a 42,949,672% trail. My first attempt at this guard was in the wrong place and my own test caught it: the check has to sit in validation, not inside the branch NaN never enters. Both are refused, along with a percentage that quantises below the wire's 0.01% resolution.

**Four mutations survived the tests I shipped**, all now closed:

- reverting *only* the shared extended encoder — the exact missed-second-encoder failure I had already been caught by once in this change
- swapping the two condition flags with each other, which the old test could not detect because it set both at once
- deleting the `min_qty` refusal entirely, which had no coverage at all
- truncating instead of rounding

The five stale rustdoc comments naming the old tags are corrected, including `min_qty`, which read as supported. And a caller reaching the engine directly still cannot be refused — `OrderAttrs.min_qty` is silently unused there — so that path now says so in the log rather than doing nothing quietly.

Note the API road to the retagged condition flags is opened by #326, which adds `conditions` to the routing predicate; without it a conditions-only order never reaches this encoder at all.

userFRM:

PR #341, opened 2026-07-29
orders: refuse a quantity the wire cannot carry, rather than truncating it (ibx#313)

### LINKED ISSUE #313: orders: a fractional totalQuantity is truncated to a whole number and sent without warning, so 1.5 shares becomes an order for 1
## Problem

`build_order_request` converts the requested quantity with a plain cast:

```rust
let qty = order.total_quantity as u32;
```

`total_quantity` is an `f64` on the public API, so any fractional part is discarded. Nothing validates it and nothing reports it — the truncated value goes straight onto the wire as tag 38.

## Impact

Observed on paper against the live gateway, market open, SPY:

Requesting `total_quantity: 1.5` sends

```
35=D ... 55=SPY|54=1|38=1|40=1 ...
```

and the order fills for **1 share**. The caller asked for 1.5 and got 1, with no error, no warning, and an `order_status` that reports `filled=1 remaining=0` as though that were the whole order. Position sizing built on the requested quantity is then wrong by the discarded fraction, and every downstream calculation inherits it.

The rounding is toward zero, so a quantity below one share disappears entirely. Requesting `total_quantity: 0.5` sends `38=0`, and the order comes back Inactive:

```
35=D ... 55=SPY|54=1|38=0|40=1 ...
[status] Inactive filled=0 remaining=0
```

That case at least fails loudly. The 1.5 case does not, which is the one that costs money.

This is reachable from the whole public surface: `EClient::place_order` and the Python bindings both route through `build_order_request`.

## Solution

Two parts, and the first is worth doing on its own:

- **Do not silently discard the fraction.** Either reject a quantity that is not a whole number with a clear error, or carry it through. Truncating and continuing is the one behaviour that should not survive.
- **Carry it through where the engine already can.** The encode side has the pieces: `QTY_SCALE` (10^4), the `format_qty` formatter, and a fractional order variant that sends `38=0.5` correctly. What is missing is a path from the public API to it — `build_order_request` has no fractional branch.

Worth noting the decode half has the mirror-image gap. `LastShares` (32) and `LeavesQty` (151) are parsed with `parse::<i64>()`, so a fractional value fails to parse, falls to `0`, and the fill is gated out entirely — no fill callback, no position update, no execution record. On this account the gateway formats whole quantities as plain integers (`32=1`, `151=0`, confirmed on the captures above), so that path is only reachable with a genuinely fractional fill, which the truncation above currently prevents anyone from producing through the public API. Fixing the encode side without the decode side would turn a wrong-size order into a dropped fill, so they want to land together.

userFRM:
## Summary

- The order quantity reaches the wire through `as u32`, which truncates. A caller asking for **1.5 shares sent an order for 1**, and was told nothing.
- Observed on paper: `total_quantity: 1.5` goes out as `38=1` and fills one share. The fill, the status and the position are all internally consistent — with an order that was never placed.

```
35=D ... 55=SPY|54=1|38=1|40=1 ...
[status] oid=9021 Filled filled=1 remaining=0
```

Closes #313.

## Why the dangerous case is the quiet one

Below one share it fails loudly — `0.5` goes out as `38=0` and the order comes back Inactive. It is the values **above** one that cost money, because they look like they worked. Position sizing built on the requested quantity is wrong by the discarded fraction, and everything downstream inherits it.

The same cast also turns a non-finite quantity into zero or the maximum, and a negative one into a very large positive.

## Change

Refuse a quantity this path cannot carry, and say which one it is: not a whole number, not finite, negative, or larger than the wire type.

The encoder does have the pieces for fractional orders — a fixed-point scale, a quantity formatter, and a variant that puts `38=0.5` on the wire correctly — but nothing routes from the public API to them. Until something does, refusing is the honest answer; rounding a caller's size down in silence is not.

Whole quantities are unaffected, and there is a test pinning that.

## Tests

Four cases, each mutation-checked: dropping any one of the four guards fails them.

## Test plan

### DIFF
--- a/src/api/client/tests.rs
+++ b/src/api/client/tests.rs
@@ -31,6 +31,98 @@ fn spy() -> Contract {
 //  Algo parsing
 // ═══════════════════════════════════════════════════════════════════
 
+/// The quantity reaches the wire through `as u32`, which truncates. Asking for
+/// 1.5 shares sent an order for 1 and reported nothing — the fill, the status
+/// and the position were all consistent with an order that was never placed.
+#[test]
+fn a_fractional_quantity_is_refused_rather_than_truncated() {
+    let (client, rx, _shared) = test_client();
+    let order = Order {
+        action: "BUY".into(), total_quantity: 1.5, order_type: "MKT".into(),
+        tif: "DAY".into(), ..Default::default()
+    };
+    let err = client.place_order(9101, &spy(), &order).expect_err("must be refused");
+    assert!(format!("{err}").contains("whole number"), "the error says why: {err}");
+    assert!(rx.try_recv().is_err(), "and nothing reaches the wire");
+}
+
+/// A quantity that is not a number, or is negative, or overflows the wire
+/// type, all reach `as u32` and become something the caller did not ask for.
+#[test]
+fn an_unusable_quantity_is_refused() {
+    for (qty, expect) in [
+        (f64::NAN, "finite"),
+        (f64::INFINITY, "finite"),
+        (-5.0, "negative"),
+        (5e9, "too large"),
+    ] {
+        let (client, _rx, _shared) = test_client();
+        let order = Order {
+            action: "BUY".into(), total_quantity: qty, order_type: "MKT".into(),
+            tif: "DAY".into(), ..Default::default()
+        };
+        let err = client.place_order(9102, &spy(), &order)
+            .expect_err("must be refused").to_string();
+        assert!(err.contains(expect), "quantity {qty}: expected {expect:?}, got: {err}");
+    }
+}
+
+/// The boundaries, exactly. Each of these was reachable by a mutation that the
+/// round-number cases above could not see: rejecting the valid maximum,
+/// admitting one past it, admitting a small negative, admitting a fraction
+/// near a half, and misclassifying negative infinity.
+#[test]
+fn the_quantity_boundaries_are_exact() {
+    let place = |qty: f64| {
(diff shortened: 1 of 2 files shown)


userFRM:
Three corrections, and the first one is a factual error in the commit message that is now fixed.

**I claimed a negative quantity became a large positive.** It does not. Rust's float-to-integer `as` cast saturates, so `-5.0_f64 as u32` is `0` — verified on this toolchain both directly and through `place_order` at the base commit. An order to sell five short became an order for nothing, which is still worth refusing, but it is not what I wrote. The message says the right thing now.

**Zero was still going out.** The guard accepted `0.0` and the cast passed it straight through. A blanket rejection would have been wrong in the other direction: a cash-quantity order legitimately carries no shares, because its size is stated in currency instead. So zero is refused only when no cash quantity is supplied, and both halves are tested.

**The boundaries were only tested at round numbers**, and six harmful mutations passed as a result — rejecting the valid maximum, admitting one past it, admitting `-1.0`, admitting `1.25`, misclassifying negative infinity, and breaking cash-quantity orders. Each is now pinned exactly and each mutation fails.

On the risk this change actually carries, which is a refusal that is too broad: every in-tree caller was traced and none breaks. The three that compute a quantity rather than taking it from the user are safe — one rounds up explicitly, one reuses a whole-share fill quantity, one is a command-line argument. The guard sits at a single choke point ahead of both the submit and the modify paths, so the what-if and algo submissions that return early further down are still covered.

userFRM:

PR #342, opened 2026-07-29
orders: refuse a modify the replace cannot express, rather than cancelling the order (ibx#334)

### LINKED ISSUE #334: orders: a replace restates almost none of the order — modifying a trailing stop drops its trail and the order goes Inactive

userFRM:
## Summary

- The replace message states the order type, the limit price and the trigger. It carries **no** peg offset, trailing amount or execution instruction — so for an order defined by one of those, it describes something other than the order being replaced.

Closes #334.

## Observed

A trailing stop, submitted:

```
35=D  ...|54=2|38=1|40=P|99=100|211=100|18=a|59=0|167=STK
```

then re-placed with only the quantity changed:

```
35=G  ...|41=9501.0|44=0|6122=c|38=2|54=2|40=P|55=SPY|167=STK|59=0|6008=756733
```

`40=P` survives; everything that made it a trailing stop does not — no 99, no 211, no `18=a`. What is left is a pegged order with no offset, and the gateway answers:

```
[status] oid=9501 Inactive
```

So changing the size of a trailing stop leaves you without one. The order is not adjusted and not left alone — it is knocked out, and the only signal is a status the caller has to be watching for. For a protective trailing stop that is the entire purpose of the order.

## Change

Refuse the modify. The order keeps working, which is strictly better than a replace that cancels it; the caller can cancel and re-place if they want a different order, and the difference is that they choose it.

The types a replace can restate are the ones defined by the fields it carries — MKT, LMT, STP, STP LMT, MOC, LOC, MIT, LIT, STP PRT. Everything else is refused, on both the Rust and Python clients.

## What this is not

It does not make the replace able to carry those orders. Doing that means restating the order's own type-specific fields, which `OrderRequest::Modify` does not carry — that is the rest of #334, and a larger change. Until then, saying so is the honest behaviour.

## Tests

Refusal across the trailing, pegged, snap, mid-price and relative types, plus a test that an ordinary limit still modifies. Mutation-checked in both directions: allowing every type through fails, and refusing plain limits fails too.

## Test plan

- [x] `cargo test --offline --lib` — 808 passed. The 2 failures are `config::expiry_tests::{named_zone_converts_with_dst, instant_round_trips_to_wire}`, which fail on the base commit too: the host has no legacy timezone files (fixed separately in #336).

### DIFF
--- a/src/api/client/orders.rs
+++ b/src/api/client/orders.rs
@@ -31,6 +31,17 @@ impl EClient {
 
         // If orderId is already tracked, this is a modification — emit Modify instead of Submit.
         let cmd = if self.core.is_order_tracked(oid) {
+            // A replace states the order type, the limit price and the trigger.
+            // An order defined by anything else cannot survive one, so refuse
+            // rather than send a message that destroys it.
+            if let Some(tracked) = self.core.tracked_order(oid) {
+                if let Some(why) = ClientCore::replace_cannot_restate(&tracked) {
+                    return Err(format!(
+                        "{why} cannot be modified: the replace does not carry the fields that \
+                         define it, and sending one would cancel the order"
+                    ));
+                }
+            }
             let price = (order.lmt_price * PRICE_SCALE_F) as i64;
             let qty = order.total_quantity as u32;
             ControlCommand::Order(OrderRequest::Modify {
--- a/src/api/client/tests.rs
+++ b/src/api/client/tests.rs
@@ -31,6 +31,167 @@ fn spy() -> Contract {
 //  Algo parsing
 // ═══════════════════════════════════════════════════════════════════
 
+/// A replace carries the order type, the limit price and the trigger — not the
+/// peg offset, the trailing amount or the execution instruction. Sent for a
+/// trailing stop it describes a pegged order with no offset, the gateway
+/// rejects it, and the caller is left with no stop at all. Refusing keeps the
+/// order they already have.
+#[test]
+fn a_type_the_replace_cannot_restate_is_not_modified() {
+    for order_type in ["TRAIL", "TRAIL LIMIT", "REL", "PEG MID", "MIDPX", "SNAP MID"] {
+        let (client, rx, _shared) = test_client();
+        let submit = Order {
+            action: "SELL".into(), total_quantity: 1.0, order_type: order_type.into(),
+            aux_price: 1.0, trailing_percent: 0.0, tif: "DAY".into(), ..Default::default()
+        };
+        // Submitting is fine; it is the replace that cannot express it.
+        let _ = client.place_order(9201, &spy(), &submit);
+        while rx.try_recv().is_ok() {}
+
+        // Skipping when tracking did not happen would let this pass without
+        // testing anything, which is how the modify gate went unnoticed.
+        assert!(
+            client.core.is_order_tracked(9201),
+            "{order_type} must submit and be tracked, or the refusal below proves nothing",
+        );
+        let err = client.place_order(9201, &spy(), &submit)
+            .expect_err("modifying it must be refused");
+        assert!(err.contains("cannot be modified"), "{order_type}: {err}");
+        assert!(rx.try_recv().is_err(), "{order_type}: nothing reaches the wire");
+    }
+}
+
+/// The order type alone does not decide this. An adaptive or algo order is an
+/// ordinary LMT defined by its algo tags; an adjustable stop is an ordinary STP
+/// defined by its conversion; a conditional order rides submit-only tags. A
+/// replace states none of those, so each is destroyed by one just as surely as
+/// a trailing stop is — and each would have passed a gate that looked only at
+/// the type.
+#[test]
+fn an_order_defined_by_more_than_its_type_is_not_modified() {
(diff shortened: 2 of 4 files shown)


userFRM:
The gate was too narrow in two provable ways — the refusal itself was correctly scoped, but the claim that everything else is refused was false.

**The order type alone does not decide this.** An adaptive or algo order is an ordinary `LMT` defined by its algo tags; an adjustable stop is an ordinary `STP` defined by its conversion; a conditional order rides tags the replace never carries. All three passed a gate that read only the type, and all three are destroyed by a replace exactly as a trailing stop is — proven by driving them through: the modify was accepted and emitted, and the replace states a bare `40=2` or `40=3` with the defining tags gone. The gate now reads the tracked order, not just its type string.

**Limit-if-touched was wrongly on the allowed list.** It is submitted as `40=LT` but tracked under a byte the replace renders as `K` — which is market-to-limit in this dialect. So a LIT modify sends a replace describing a different order type outright: either Inactive, or silently converted. It comes back on the list once the engine carries a discriminant that renders as `LT`.

The empty order type went with them, since it is reachable through the algo path and through a hydrated order missing tag 40, where the replace would state a type inferred from engine defaults.

Each fix fails a test under mutation. Verified in the other direction too: **no flow that worked before is now refused** — every excluded type was already being destroyed by the replace at base.

One thing worth separating out, not introduced here: for the types that *are* allowed, the type survives a modify but the trigger does not move — `Modify` carries only price and quantity, and the replace restates the original trigger. So a caller moving a stop through `place_order` on the same id gets an acked modify with the old trigger and no error. Same silent-divergence family as this issue; it deserves its own.

userFRM:
The gate was still too narrow, and two of the tests could pass without testing anything.

**Four more attributes ride tags a replace does not carry.** Hidden orders on 6135, all-or-none as an execution instruction, and the cash quantity on 5920 are each dropped by a replace exactly as a peg offset is — so a hidden iceberg becomes a visible order and a cash-sized order loses its size. What-if is refused for a different reason: it is a margin preview, so there is no resting order for a replace to act on at all. All four are refused now, and each fails a test under mutation.

**Two refusal tests skipped silently when the submit did not track.** A `continue` there means the loop can complete having asserted nothing — which is how the gate's original narrowness went unnoticed in the first place. They now require the order to be tracked before the refusal is meaningful. Both were in fact passing genuinely; the point is that they no longer *can* pass otherwise.

**Only `LMT` was covered on the positive side**, so wrongly excluding any other allowed type would have gone unseen — dropping `STP PRT` passed the whole suite. Every restatable type is now exercised end to end, and that mutation fails.

**The tracked-versus-incoming decision was unpinned.** A caller could have made a trailing stop look modifiable by describing it as a limit on the way in. There is a test for that now: submit a trailing stop, then modify it while claiming it is a limit, and the refusal still stands.

Both clients share the same predicate, so the Python path gets all of this without a second copy to drift.

userFRM:

PR #343, opened 2026-07-29
ccp: stop reporting the OCA group as an order's parent (ibx#329)

### LINKED ISSUE #329: orders: parentId is derived from tag 583, which ibx itself sends as the OCA group, so OCA siblings all report the same fabricated parent
## Problem

ibx contradicts itself about tag 583.

Outbound, it writes the OCA group there, in three places, labelled as such:

```rust
(583, &oca_group),         // OCAGroup
```

Inbound, it reads the same tag as the parent order id and hashes it:

```rust
let parent_id: i64 = parsed.get(&583).map(|s| perm_id_from_fix_order_id(s)).unwrap_or(0);
```

Both cannot be right.

## Evidence

Tag 583 is the OCA group. Two orders placed in group `PROBE-OCA-1`, then one cancelled — the gateway cancelled the sibling, which is what an OCA group is for:

```
== placing BUY 1 SPY LMT 100 oca=PROBE-OCA-1 (oid=9401)
== placing BUY 1 SPY LMT 101 oca=PROBE-OCA-1 (oid=9402)
== cancelling oid=9401
[status] oid=9402 Cancelled
OCA VERDICT: GROUPED (sibling cancelled)
```

And the `parentId` reported for those same two orders:

```
[status] oid=9401 PreSubmitted perm_id=4480919197864673551 parent_id=9132926678952268133
[status] oid=9402 PreSubmitted perm_id=4480920297376301762 parent_id=9132926678952268133
```

Both report the **same** parent, and neither order has one — they are siblings in a group, not children of anything. The number is the hash of the string `PROBE-OCA-1`.

## Impact

Every OCA order reports a fabricated `parentId` that is stable, non-zero, shared by the whole group, and corresponds to no order that exists. Anything reconstructing order relationships from callbacks — a bracket tracker, a position attributor, a UI grouping children under parents — reads it as a real link and builds the wrong tree. It never reads as missing data, which is what would make it easy to notice.

A non-OCA order has no tag 583, so `parentId` comes out 0 and looks correct. The wrong value appears exactly when a group is in use.

## Solution

Read the parent from the tag that carries it, and stop deriving it from the OCA group. Tag 583 is already being written as the group on the way out; reading it as something else on the way in is the whole bug.

While that is being settled, worth checking `permId` in the same handler: it is a hash of tag 37 rather than the integer the gateway assigns, so it is stable but is not the gateway's permId, and it will not match one quoted anywhere else.

userFRM:
## Summary

The engine writes the OCA group to tag 583 on the way out, in three places, and read the same tag back in as the parent order id — hashed into the `i64` the callback exposes. Both cannot be right.

```rust
(583, &oca_group),         // OCAGroup                          ← outbound
let parent_id = parsed.get(&583).map(perm_id_from_fix_order_id) // ← inbound
```

Closes #329.

## Evidence

583 is the group. Two orders placed in group `PROBE-OCA-1`, then one cancelled — the gateway cancelled the sibling, which is what an OCA group is for:

```
== placing BUY 1 SPY LMT 100 oca=PROBE-OCA-1 (oid=9401)
== placing BUY 1 SPY LMT 101 oca=PROBE-OCA-1 (oid=9402)
== cancelling oid=9401
[status] oid=9402 Cancelled          ← sibling went with it
```

The inbound reports carry the group on 583 for both, with distinct order ids:

```
oid=9401.0   583=PROBE-OCA-1   37=0256d0f1.0001417e.6a698306.0001
oid=9402.0   583=PROBE-OCA-1   37=0256d0f1.0001417e.6a698308.0001
```

and the parent reported for them was:

```
[status] oid=9401 PreSubmitted parent_id=9132926678952268133
[status] oid=9402 PreSubmitted parent_id=9132926678952268133
```

The same non-zero number for two orders that have no parent — the hash of the group name.

## Impact

Anything reconstructing order relationships from the callbacks — a bracket tracker, a position attributor, a UI grouping children under parents — reads that as a real link and builds the wrong tree. It never reads as missing data, which is what would make it easy to notice: a non-OCA order has no 583, reports zero, and looks correct. The wrong value appears exactly when a group is in use.

## Change

Nothing on the execution report carries a parent order id, so none is reported. An unknown parent is a worse answer than a correct one and a much better answer than a confident wrong one.

## Tests

An order carrying an OCA group reports no parent, and its status is still delivered. Restoring the hash fails it.

## Test plan

- [x] `cargo test --offline --lib` — 810 passed. The 2 failures are `config::expiry_tests::{named_zone_converts_with_dst, instant_round_trips_to_wire}`, which fail on the base commit too: the host has no legacy timezone files (fixed separately in #336).
- [x] Mutation check: reverting the fill-path parent fails `a_fill_reports_the_parent_the_child_was_given`; reverting the preview path fails `a_what_if_preview_reports_the_parent_the_child_was_given`.
- [ ] Not covered: reverting the type-4 dispatch itself fails no test — the route-to-gate step runs inside `connect`.

### DIFF
--- a/src/api/client/dispatch.rs
+++ b/src/api/client/dispatch.rs
@@ -108,9 +108,15 @@ impl EClient {
         // Order updates → order_status
         for update in self.shared.orders.drain_order_updates() {
             let status = order_status_str(update.status);
+            // The engine cannot know the parent — nothing on the execution
+            // report carries one — but this client placed the order and was
+            // told. Prefer what it recorded; an order it did not place keeps
+            // the engine's answer of none.
+            let parent_id = self.core.tracked_parent_id(update.order_id)
+                .unwrap_or(update.parent_id);
             wrapper.order_status(
                 update.order_id as i64, status, update.filled_qty as f64,
-                update.remaining_qty as f64, 0.0, update.perm_id, update.parent_id, 0.0, 0, "", 0.0,
+                update.remaining_qty as f64, 0.0, update.perm_id, parent_id, 0.0, 0, "", 0.0,
             );
             self.core.update_order_status(update.order_id, status, update.filled_qty as f64, update.remaining_qty as f64);
         }
--- a/src/api/client/dispatch.rs
+++ b/src/api/client/dispatch.rs
@@ -51,9 +51,13 @@ impl EClient {
             let price_f = fill.price as f64 / PRICE_SCALE_F;
             let commission_and_fees_f = fill.commission as f64 / PRICE_SCALE_F;
             let status = if fill.remaining == 0 { "Filled" } else { "PartiallyFilled" };
-            let (perm_id, parent_id) = self.shared.orders.get_order_info(fill.order_id)
+            // A fill emits its own order_status and never reaches the branch
+            // below, so the client's record has to be preferred here too.
+            let (perm_id, engine_parent) = self.shared.orders.get_order_info(fill.order_id)
                 .map(|info| (info.order.perm_id, info.order.parent_id))
                 .unwrap_or((0, 0));
+            let parent_id = self.core.tracked_parent_id(fill.order_id)
+                .unwrap_or(engine_parent);
             wrapper.order_status(
                 fill.order_id as i64, status, fill.qty as f64, fill.remaining as f64,
                 price_f, perm_id, parent_id, price_f, 0, "", 0.0,
@@ -108,10 +112,9 @@ impl EClient {
         // Order updates → order_status
         for update in self.shared.orders.drain_order_updates() {
             let status = order_status_str(update.status);
-            // The engine cannot know the parent — nothing on the execution
-            // report carries one — but this client placed the order and was
-            // told. Prefer what it recorded; an order it did not place keeps
-            // the engine's answer of none.
+            // The engine reads no parent from the report, but this client
+            // placed the order and was told. Prefer what it recorded; an order
+            // it did not place keeps the engine's answer of none.
             let parent_id = self.core.tracked_parent_id(update.order_id)
                 .unwrap_or(update.parent_id);
             wrapper.order_status(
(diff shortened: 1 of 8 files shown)


userFRM:
Updated in `115d60b`. The parent preference now covers all three callback paths, not one.

**A fill emits its own `order_status` from a different branch** and never reached the branch that was changed; the margin preview emits a third and hard-coded zero. So a bracket child still reported no parent on the fill callback — the one a caller is most likely to act on. All three now prefer the recorded parent, and reverting any one of them fails its own test by name. Neither of the two missed paths had a test, which is what let the gap through.

**One claim corrected.** The message said nothing on the report can carry a parent, and that tag 6107 is a client id rather than an order id. The encoder derives 6107 directly from the caller's `parent_id`, so the second half contradicts this library's own protocol model, and the first was asserted rather than shown. The narrower statement is what holds: the engine reads no parent from the report, so for an order this client placed its own record is the only source. What the gateway does or does not echo is not established here.

The test pinning 6107 as unread stays, and now says what it is — a guard on a different invariant, which passes with or without this change — rather than being presented as evidence for it.

Two smaller items: the status-loop test asserted a parent only for updates it happened to receive, so a case producing none passed by never entering the loop; it now requires exactly one update per case first. And `tracked_parent_id` was `pub` on a publicly exported module, widening the API for an internal helper — now `pub(crate)`, with `is_order_tracked`'s doc comment restored to it.

810 pass plus the two `config::expiry_tests` timezone failures. `--features python` clean.

userFRM:
Updated in `b4ba91e`.

**The Python paths now have coverage.** Reverting either Python lookup — the fill or the margin preview — left every Rust test passing, so "on both clients" rested on nothing. There are three Python tests now, using the existing `_test_*` harness. The extension was built and the tests run rather than left unexecuted: all three pass, and reverting the two Python lookups fails the first two by name while the third survives, since it is the positive control. `_test_track_order` gained an optional parent id, which is what makes a tracked child expressible from Python at all.

**On `tracked_parent_id` visibility:** narrowing it to `pub(crate)` is not a source break. The method does not exist on `main` — this PR introduces it — so it has never been part of a released surface. Narrowing it inside the same change that adds it keeps an internal helper from widening a publicly exported module's API.

810 pass plus the two `config::expiry_tests` timezone failures, and the three Python tests. All offline targets clean.

userFRM:

PR #345, opened 2026-07-29
ccp: require the side on a recovery record instead of defaulting it (ibx#321)

### LINKED ISSUE #321: ccp: the recovery insert defaults a missing Side to Sell, so later fills for that order move the position the wrong way

userFRM:
## Summary

The cross-session recovery insert derived the order's side with a catch-all:

```rust
let side = match parsed.get(&54).map(|s| s.as_str()) {
    Some("1") => Side::Buy,
    Some("5") => Side::ShortSell,
    _ => Side::Sell,
};
```

Anything unrecognised, or absent, was recorded as a sell.

Closes #321.

## Why the guess does not stay put

It is not confined to the recovered record. Once the order is in `context`, **every later fill for it books through the tracked path and takes its side from the stored order** — so a recovered buy recorded as a sell moves the position *down* by the filled quantity instead of up.

The error is twice the fill size, and nothing afterwards distinguishes a guessed side from one the report actually carried.

Compared with the alternative: a fill that cannot be booked leaves the engine short of the truth, and the position feed still disagrees visibly. A wrong sign puts a confident wrong number in its place. That is the worse of the two.

## Change

A record without a usable side is left untracked — the state the engine already handles — and says so in the log. A stated side is recovered as before, including the `2` sell case the catch-all happened to cover.

This is the same reasoning as the untracked-fill path in #322, which refuses a report without tag 54 rather than defaulting it.

## Tests

Absent, unrecognised and out-of-range sides are each refused; all three stated sides are recovered with the right one. Restoring the default fails them.

`cargo test --lib`: 804 pass. The two `config::expiry_tests` failures are pre-existing and unrelated (timezone data — fixed separately in #336).

## Test plan

### DIFF
--- a/src/engine/hot_loop/ccp.rs
+++ b/src/engine/hot_loop/ccp.rs
@@ -683,10 +683,30 @@ impl CcpState {
             && parsed.get(&39).map(|s| s.as_str()) == Some("0");
         if is_new_ack && context.order(clord_id).is_none() {
             let con_id: i64 = parsed.get(&6008).and_then(|s| s.parse().ok()).unwrap_or(0);
+            // The side has to be stated. A guess does not stay in the recovered
+            // record: every later fill for the order books through the tracked
+            // path and takes its side from here, so a recovered buy recorded as
+            // a sell moves the position down by the filled quantity instead of
+            // up — wrong by twice the fill, and indistinguishable afterwards
+            // from a side the report actually carried.
             let side = match parsed.get(&54).map(|s| s.as_str()) {
-                Some("1") => Side::Buy,
-                Some("5") => Side::ShortSell,
-                _ => Side::Sell,
+                Some("1") => Some(Side::Buy),
+                Some("2") => Some(Side::Sell),
+                Some("5") => Some(Side::ShortSell),
+                other => {
+                    // The sentinel that terminates a recovery burst, and the
+                    // mass-status echo, both parse to id 0 and carry no side.
+                    // Warning about those once per connect would cry wolf on
+                    // the one signal that matters when a real record is
+                    // refused.
+                    if clord_id != 0 {
+                        log::warn!(
+                            "Recovery record for order {} has Side={:?}; not tracking it",
+                            clord_id, other,
+                        );
+                    }
+                    None
+                }
             };
             let qty: u32 = parsed.get(&38).and_then(|s| s.parse::<f64>().ok()).map(|q| q as u32).unwrap_or(0);
             let limit_price_i64: i64 = parsed.get(&44)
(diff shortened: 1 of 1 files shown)


userFRM:
Gated the new warning on a real order id.

The sentinel that terminates a recovery burst, and the mass-status echo, both parse to id `0` and carry no side — so the diagnostic fired once per connect and once per `req_open_orders` for frames that were never orders. That is cry-wolf on the one signal that matters when a genuine record is refused.

Worth recording from the review, since it settles the question the change turns on: the engine's own encoder emits exactly `1`/`2`/`5`, the existing enrichment decode understands exactly those three, and `Side` has exactly three variants — so the only value with any claim to being "legitimately recoverable" is `6` (short-sale-exempt), which IB has deprecated and which the type cannot represent anyway. The warning makes it visible if it ever appears.

The untracked afterflow was traced end to end and there is no path where untracked is worse than guessed: the recovery ack still records the last ClOrdID, so a single cancel works (the cancel message carries no side at all); status acks no-op; the order still appears in the open-order cache with an empty action rather than a fabricated `SELL`; and fills for it are dropped while the position feed stays authoritative and visibly disagrees. That is the trade the commit message states, now confirmed rather than asserted.

The review also checked whether the *other* guessed fields on that insert deserve the same treatment — quantity, order type, TIF, limit and stop. They do not: the quantity is already gated, and none of the rest enters position arithmetic, so their failure mode is a visible gateway reject on a modify echo rather than silent corruption. Leaving them alone is deliberate.

userFRM:

Issue #223, opened 2026-07-10
robustness: CLOSED trading days not rendered; trigger-method -1 wraps to 255; out-of-range enum codes pass through


deepentropy:
### Current behavior
A few small validation/robustness gaps in schedule and enum handling:

1. **CLOSED trading day not represented.** `parse_schedule_response` (`src/control/contracts.rs:561-591`) only appends sessions where the open/close flags are set. A closed day produces no entry, so a consumer cannot distinguish "market closed" from "data missing". The gateway renders a closed day as a zero-length `<date>:CLOSED` interval.
2. **Trigger-method Unknown wraps.** `trigger_method as u8` (`src/api/types.rs:445`) turns the gateway's `-1` (Unknown) into `255`, which would be emitted as `6115=255`. Out-of-range codes (`5`, `6`, `9+`) are forwarded verbatim rather than coerced to Unknown as the gateway does.
3. **Unknown secType falls back to Stock.** `SecurityType::to_fix` returns `CS` for the `Other` variant (`src/control/contracts.rs:68`), so an unrecognized secType is sent as stock-on-wire, which can misroute. The gateway maps unknown -> NONE with an error.

### Expected behavior
Represent closed days explicitly, coerce out-of-range trigger codes to Unknown (and represent Unknown without wrapping), and avoid silently sending an unknown secType as stock.

### Impact
Individually minor, but each can produce a subtly wrong schedule, order attribute, or routing.

### Suggested fix
Emit a CLOSED marker for zero-length days, validate/normalize the trigger-method code, and reject or clearly flag unknown sec types instead of defaulting to `CS`.

deepentropy:
Fixed in e157513, all three items:

1. Closed trading days are represented: a session with neither hours flag set is kept as a zero-length session in both hour lists and renders as <date>:CLOSED (official-API convention), so "market closed" and "data missing" are distinguishable.

2. Trigger-method codes are validated: -1 no longer wraps to 255 via the raw cast, and out-of-range codes (5, 6, 9+) coerce to the default instead of going to the wire verbatim. Valid codes 0-4, 7, 8 pass through.

3. An unrecognized security type is no longer encoded as a stock on the wire — it now goes out empty, which draws a visible gateway error instead of silently searching/routing as an equity. (The user-visible reporting side of this was already fixed in #230: unclassifiable instruments report an empty sec_type.)

PR #346, opened 2026-07-29
orders: send the security type in the spelling the gateway uses (ibx#328)

### LINKED ISSUE #328: orders: the submit path identifies the contract differently from every other path — no ContractID, and STK where the gateway says CS
## Problem

Two things about how a `35=D` names its contract, both visible in one capture pair on paper:

**The submit carries no ContractID.** Every `Submit*` variant identifies the instrument by Symbol, SecurityType, ExDestination and Currency only. The replace path does send tag 6008, and `context.market.con_id(instrument)` is available at encode time on both — it is simply not emitted on submit.

```
35=D  ...|55=SPY|54=1|38=1|40=1|59=0|167=STK|100=BEST|6210=BEST|15=USD|204=0
35=G  ...|44=0|6122=c|38=1|54=2|40=3|55=SPY|167=STK|6035=SPY|59=0|6008=756733|6088=Socket
```

The asymmetry is the tell: one of the two is wrong about what the gateway needs to identify a contract.

**The submit says `167=STK`, and the gateway answers `167=CS`.** Every execution report for those orders comes back with `CS`:

```
35=8  ...|150=2|39=2|167=CS|55=SPY|38=1|32=1|31=737.53|...|6008=756733
```

`CS` is also what the rest of the client uses. The market-data subscribe maps `STK` to `CS` before sending, and `SecurityType::to_fix()` returns `CS`. The order path is the one place that puts the API-facing spelling on the wire.

## Impact

Both are latent today rather than active, and for the same reason: `validate_order_contract` restricts orders to `STK`. With one security type and a symbol resolved on `BEST`, the gateway resolves it correctly and tolerates the spelling.

Both stop being latent the moment options or futures encoding lands. A symbol plus a security type is the ambiguous way to name a contract — it is why the conId exists, why every secdef request ibx sends carries it, and why every inbound execution report is keyed on it. An order that resolves to a different contract than the caller passed to `place_order` is unbounded cost, and the only way to notice is to compare the `6008` the gateway echoes against the `con_id` you supplied.

The security type matters at the same moment: a wrong or unrecognised value on an options or futures order is exactly the silent misrouting that #223 already fixed for the request path, where an unknown type is sent empty rather than defaulted to a stock.

## Solution

Send the conId on submit, as the replace already does, and map the security type through the same conversion the rest of the client uses rather than putting the API spelling on the wire.

Doing it while orders are still STK-only is cheap and keeps the two paths from disagreeing later, when a wrong contract is no longer improbable.

userFRM:
## Summary

Order messages put the API-facing security type on tag 167. The gateway answers every execution report with the wire spelling:

```
35=D  ...|55=SPY|54=1|38=1|40=1|167=STK|100=BEST|15=USD      ← sent
35=8  ...|150=2|39=2|167=CS|55=SPY|38=1|32=1|31=737.53       ← answered
```

The rest of the client already converts before sending — the market-data subscribe maps it, and `SecurityType::to_fix` exists for exactly this. The order path was the one place putting the caller's word on the wire.

Closes #328.

## Change

`order_routing` returns the value every caller writes to tag 167 and nothing else — confirmed across all thirty-eight call sites — so it converts there rather than at each one.

It goes through the existing security-type mapping, which means an **unrecognised type is now sent empty** rather than passed through as itself. That is what the mapping already documents: an instrument the engine could not classify must not masquerade as something the gateway accepts, because that misrouting is silent (#223).

## Why now, if it is latent

Order submission is restricted to stocks and the gateway resolves `STK` for those, so nothing is broken today. It stops being latent the moment another security type can be submitted, and the fix is one line at one place while that is still true.

Verified against a live session — a limit order carrying `167=CS` is accepted exactly as before:

```
35=D  ...|55=SPY|54=1|38=1|40=2|44=100|59=0|167=CS|100=BEST|6210=BEST|15=USD
[status] oid=9951 PreSubmitted
```

## Not addressed here

The other half of #328: a submit carries no ContractID while a replace does, so the two disagree about how a contract is named. Adding it touches every submit arm and deserves its own change — I have left the issue open for that reason rather than closing it on a half fix.

`cargo test --lib`: 802 pass. The two `config::expiry_tests` failures are pre-existing and unrelated (timezone data — fixed separately in #336).

## Test plan

- [x] Mutation check: reverting the wire-spelling conversion fails `engine::market_state::tests::order_routing_rules` by name.

### DIFF
--- a/src/engine/market_state.rs
+++ b/src/engine/market_state.rs
@@ -200,20 +200,29 @@ impl MarketState {
     }
 
     /// Routing tags for an outbound order on this instrument (ibx#217):
-    /// (security type, destination). Rules: the security type is the
-    /// instrument's real one (default STK); IBKRATS resolves to IDEALPRO
-    /// for CASH and BEST otherwise; CASH without an explicit venue routes
-    /// to IDEALPRO; any other explicit non-SMART exchange is respected;
-    /// everything else routes BEST — the wire form of default routing.
-    /// The reference encoder structurally cannot emit "SMART": it
-    /// canonicalizes to it internally and translates to "BEST" at the
-    /// encode boundary (ib-agent#165), and sending "SMART" was observed to
-    /// produce NO ack at all for pre-market opening-auction orders while
-    /// the gateway answers "BEST" in ~130ms (ib-agent#164).
+    /// (security type, destination).
+    ///
+    /// The security type is the instrument's real one, in its wire spelling —
+    /// which is what every caller puts on tag 167. The API-facing name and the
+    /// wire name differ for stocks, where the gateway answers `CS` on every
+    /// execution report, so an instrument with no recorded type defaults to
+    /// `CS` rather than to the API's `STK` (ibx#328).
+    ///
+    /// Destination rules: IBKRATS resolves to IDEALPRO for CASH and BEST
+    /// otherwise; CASH without an explicit venue routes to IDEALPRO; any other
+    /// explicit non-SMART exchange is respected; everything else routes BEST —
+    /// the wire form of default routing. The reference encoder structurally
+    /// cannot emit "SMART": it canonicalizes to it internally and translates to
+    /// "BEST" at the encode boundary (ib-agent#165), and sending "SMART" was
+    /// observed to produce NO ack at all for pre-market opening-auction orders
+    /// while the gateway answers "BEST" in ~130ms (ib-agent#164).
     pub fn order_routing(&self, id: InstrumentId) -> (String, String) {
         let sec_type = self.sec_types[id as usize]
             .clone()
             .unwrap_or_else(|| "STK".to_string());
+        let sec_type = crate::control::contracts::SecurityType::from_fix(&sec_type)
+            .to_fix()
+            .to_string();
         let exchange = self.exchanges[id as usize].as_deref().unwrap_or("");
         let is_cash = sec_type == "CASH";
         let destination = match exchange {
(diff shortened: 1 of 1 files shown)


userFRM:
Updated in `61ae1bd`.

**Doc comment corrected.** The block above `order_routing` still said the security type defaults to `STK`, directly above a function this change makes return `CS`. The commit had added a second doc block beneath the first rather than editing it, so the function carried two descriptions with the older contradicting the code. Merged into one.

**One item left as is.** The changed test exercises `order_routing` directly rather than a serialized `35=D`, so a future encoder that stopped writing the result to tag 167 would not be caught. True, but that guards against a caller dropping the helper rather than against this change, and the conversion itself is covered: reverting it fails `order_routing_rules` with `("STK", "BEST") != ("CS", "BEST")`. A full order-serialization test is scope this branch does not need.

Verification beyond the diff:

- All 39 `Submit*` variants take the result unmodified for tag 167 — 36 inline, three through the shared encoder — and the same helper feeds tracked replacements, so 38 call sites move together.
- Round-trip per type: missing, `STK` and `CS` all land on `CS`; `OPT`, `FUT`, `CASH`, `IND`, `BOND` and `WAR` keep their spelling; unknown types stay empty as the mapping already documented.
- The destination rules are unaffected, which was worth confirming rather than assuming: `is_cash` compares against the *converted* value, so it would have broken silently had `CASH` differed between spellings. It does not, and `CASH` still selects IDEALPRO, non-cash defaults still route BEST, directed exchanges still pass through.

802 pass plus the two `config::expiry_tests` timezone failures.

userFRM:

PR #355, opened 2026-07-30
auth: validate the DH server hello instead of trusting and panicking (ibx#276)

### LINKED ISSUE #276: auth: the DH server public value is used without range validation, and a malformed server hello panics the connect path
> **Correction to the threat model below.** The claim that every socket runs inside TLS holds for the auth connection only. `connect_farm` opens a plain `TcpStream` and runs the key exchange and the signed frame stream over it — its own comment calls it "raw TCP" — so the trading and historical-data farm channels have no TLS beneath them. On those channels this is not a defence-in-depth layer sitting behind a working one; it is the layer.

## Threat model first

As with #275: every socket runs inside TLS with certificate validation enabled at all four construction sites, so this is not remotely exploitable against a correct TLS path. Filing publicly for the same reason — no `SECURITY.md`, private vulnerability reporting disabled, and the exposure is a missing check rather than a live exploit.

The panics in the second half need no adversary at all: a short or malformed server hello reaches them on the ordinary connect path.

## Problem

`process_server_hello` (`src/auth/dh.rs:95-103`) takes the server's public value and uses it without any validation:

```rust
pub fn process_server_hello(&mut self, fields: &[&str]) {
    let server_random = B64.decode(fields[0]).unwrap();
    let server_pub_bytes = B64.decode(fields[1]).unwrap();
    let server_pub = BigUint::from_bytes_be(&server_pub_bytes);
    let n = dh_n();
    let shared = server_pub.modpow(&self.private_key, &n);
```

There is no check that `server_pub` lies in [2, N-2], and no rejection of 0, 1 or N-1. A peer sending 0 or 1 pins the pre-master secret to a known constant, and the whole 104-byte key block derived from it (`dh.rs:123-129`) — both AES keys, both IVs, both HMAC keys — follows from that constant. Nothing downstream would notice: the one consumer that does enforce its MAC (`SecureChannel::decrypt`, `dh.rs:171-174`) would be verifying against the key the peer chose.

**The same function panics on a malformed hello.** `fields[0]`, `fields[1]` and the two `.unwrap()`s abort on a short or non-base64 payload, and the three call sites pass `&parts[2..]` (`gateway.rs:482`, `:662`, `:987`), which itself panics when the message splits into fewer than three parts. These run during connect, so a malformed hello is an abort rather than a connection error the caller can handle.

## Solution

Return `Result` instead of panicking: reject `fields.len() < 2`, reject undecodable base64, and range-check `server_pub` against [2, N-2] before the `modpow`. The call sites already have error paths beside them (`gateway.rs:978` formats a DH error), so there is somewhere for the failure to go.

## Notes

Read from v0.7.1 (9367845). The missing range check and the four panic sites are direct reads; I have not fed a degenerate hello to a live gateway.

Related: #275 covers the inbound HMAC never being enforced. The two compose — that issue notes SRP also never verifies the server's M2 proof, so with this one unfixed there is no layer beneath TLS that authenticates the peer.

userFRM:
## Summary

- `process_server_hello` (`src/auth/dh.rs`) indexed `fields[0]` and `fields[1]` directly and decoded both with `unwrap`, and its three call sites passed `&parts[2..]`, which itself panics when the message splits into fewer than three parts. All of it runs during connect, so a short or non-base64 hello aborted the process instead of failing the connection.
- It returns `io::Result` now. The call sites already had `return Err(io::Error::new(...))` immediately above them, so there was somewhere for the failure to go.
- The server's public value is range-checked against `[2, N-2]` before the modpow. 0 and 1 pin the pre-master secret to a known constant, and the whole 104-byte key block derived from it follows — both AES keys, both IVs, both HMAC keys.

## Why

The panics need no adversary; an ordinary malformed frame reaches them on the normal connect path.

The range check is defence in depth: every socket runs inside TLS with certificate validation enabled at all four construction sites, so this is not remotely exploitable against a correct TLS path. It matters because nothing downstream would notice a chosen key — the one consumer that does enforce its MAC would be verifying against the key the peer supplied. That is the reason for the layer to hold on its own rather than to lean on the one above it.

Closes #276.

## What is NOT included

The issue notes that `do_srp` never verifies the server's M2 proof. That needs an M2 function `srp.rs` does not have, and belongs in its own change. Related: #275 covers the inbound frame MAC never being enforced.

## Correction to the security rationale

The issue this closes, and this PR's first description, said every socket runs inside TLS with certificate validation, making the range check defence in depth. **That is wrong.** It holds for the auth connection only. `connect_farm` opens a plain `TcpStream` and runs the key exchange over it — the code's own comment calls it "raw TCP" — so the trading and historical-data farms have no TLS beneath them.

On those channels the DH exchange is the only thing between the client and an active intermediary. The range check removes the degenerate case, not the exposure: an in-range key from an intermediary is still accepted. I have corrected #276 and #275 as well, because both carried the same false premise and it understated them.

## Known, and stated rather than hidden

- **This is a source-breaking signature change.** `process_server_hello` goes from `()` to `io::Result<()>` on an exported type. Every in-repo caller is updated, but a downstream caller using it as a tail expression would not compile. Representing "this hello is unusable" needs a fallible return, so the break is not avoidable — only relocatable to a parallel method, which I would rather not add without your call.
- **The call sites' error propagation is not covered by test.** Reverting all three to their old ignoring form leaves both new tests passing, because they call the helper directly; the sites live inside `connect` and need a socket.
- One diagnosis in the first message was wrong: reaching the `533` branch already proves at least two parts, so `&parts[2..]` was valid and the panic came from `fields[0]` alone. The `.get(2..)` is harmless but was not load-bearing.

## Test plan

- [x] `cargo check --offline` clean on every offline target: `--lib`, `--features python`, `--bins`, `--examples`, `--test control_plane`, `--test scenarios`, `--test hot_loop_lifecycle`.
- [x] Mutation check: dropping the range check fails `a_degenerate_public_value_is_refused`; restoring the direct field indexing fails `a_malformed_hello_is_an_error_not_a_panic`. Both by name.
- [x] The range test covers 0, 1, N-1 and N, and carries a positive control — `N-2` is still accepted, so the refusals are not passing for want of a working path.

### DIFF
--- a/src/auth/dh.rs
+++ b/src/auth/dh.rs
@@ -92,12 +92,33 @@ impl SecureChannel {
     ///
     /// `fields` are the semicolon-split parts after version and msg_type:
     /// `[server_random_b64, server_pub_b64, ...]`
-    pub fn process_server_hello(&mut self, fields: &[&str]) {
-        let server_random = B64.decode(fields[0]).unwrap();
-        let server_pub_bytes = B64.decode(fields[1]).unwrap();
+    pub fn process_server_hello(&mut self, fields: &[&str]) -> std::io::Result<()> {
+        let invalid = |what: &str| {
+            std::io::Error::new(std::io::ErrorKind::InvalidData, format!("DH server hello: {what}"))
+        };
+        // Every field below was indexed directly and decoded with `unwrap`, so a
+        // short or non-base64 hello aborted the process on the ordinary connect
+        // path rather than failing the connection (ibx#276).
+        let [server_random_b64, server_pub_b64, ..] = fields else {
+            return Err(invalid(&format!("expected at least 2 fields, got {}", fields.len())));
+        };
+        let server_random = B64.decode(server_random_b64)
+            .map_err(|e| invalid(&format!("undecodable server random: {e}")))?;
+        let server_pub_bytes = B64.decode(server_pub_b64)
+            .map_err(|e| invalid(&format!("undecodable server public value: {e}")))?;
         let server_pub = BigUint::from_bytes_be(&server_pub_bytes);
 
         let n = dh_n();
+
+        // The public value has to lie in [2, N-2]. 0 and 1 pin the pre-master
+        // secret to a known constant, and so therefore the whole key block
+        // derived from it — both AES keys, both IVs, both HMAC keys. Nothing
+        // downstream would notice: the channel's own MAC would be verifying
+        // against the key the peer chose.
+        let two = BigUint::from(2u32);
+        if server_pub < two || server_pub > &n - &two {
+            return Err(invalid("public value outside [2, N-2]"));
+        }
 
         // Pre-master secret = server_pub ^ client_private mod N
         let shared = server_pub.modpow(&self.private_key, &n);
@@ -127,6 +148,7 @@ impl SecureChannel {
         self.write_mac_key = Some(key_block[64..84].to_vec());
         self.read_mac_key = Some(key_block[84..104].to_vec());
         self.key_block = Some(key_block);
+        Ok(())
     }
 
     /// Encrypt plaintext using Encrypt-then-MAC.
@@ -401,8 +423,8 @@ mod tests {
 
         // Each channel processes the other's hello as if it were a server response
         // process_server_hello expects [server_random_b64, server_pub_b64]
-        channel_a.process_server_hello(&[b_random, b_pub]);
-        channel_b.process_server_hello(&[a_random, a_pub]);
+        channel_a.process_server_hello(&[b_random, b_pub]).unwrap();
+        channel_b.process_server_hello(&[a_random, a_pub]).unwrap();
 
         // Both should now have key_blocks of 104 bytes
         let kb_a = channel_a.key_block().expect("channel_a should have key_block");
@@ -432,8 +454,8 @@ mod tests {
         let parts_b: Vec<&str> = payload_b.trim_end_matches(';').split(';').collect();
 
         // Each processes the other's hello
-        channel_a.process_server_hello(&[parts_b[4], parts_b[5]]);
-        channel_b.process_server_hello(&[parts_a[4], parts_a[5]]);
+        channel_a.process_server_hello(&[parts_b[4], parts_b[5]]).unwrap();
+        channel_b.process_server_hello(&[parts_a[4], parts_a[5]]).unwrap();
 
         // Both have valid key blocks
         assert_eq!(channel_a.key_block().unwrap().len(), 104);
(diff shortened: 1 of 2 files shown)


userFRM:
Reviewed. The functional change holds — 0, 1, N-1 and N rejected, both endpoints accepted, malformed hellos returning errors, all three callers propagating — but two things in how I described it were wrong. Corrected in `580630b`.

**The security rationale was false, and it understated the issue.** I wrote that every socket runs inside TLS with certificate validation, so this was defence in depth. That holds for the auth connection only. `connect_farm` opens a plain `TcpStream` and runs the key exchange over it — the code's own comment calls it "raw TCP". So on the trading and historical-data farms there is no TLS beneath this at all, and the DH exchange is the only thing between the client and an active intermediary. The range check removes the degenerate case, not the exposure. I have corrected #275 and #276 too, since both carried the same premise.

**One diagnosis was wrong in the other direction.** I said `&parts[2..]` panics when the message splits into fewer than three parts. Reaching the `533` branch already proves at least two, and slicing `[2..]` of a two-element slice is valid. The panic came from `fields[0]` alone. The `.get(2..)` is harmless but was not doing the work I credited it with.

**Two things I am flagging rather than fixing, because they are yours to decide:**

- This is a source-breaking signature change on an exported type. Every in-repo caller is updated, but a downstream caller using `process_server_hello` as a tail expression would not compile. Representing "this hello is unusable" needs a fallible return, so the break is not avoidable, only relocatable into a parallel method — which I would rather not add unasked.
- The call sites' propagation is not covered: reverting all three to their old ignoring form leaves both new tests passing, because they exercise the helper directly and the sites live inside `connect`, which needs a socket.

804 pass plus the two `config::expiry_tests` timezone failures. All offline targets clean.

userFRM:
Second review agrees the code is merge-ready as it stands: neither guard can refuse a legitimate hello, both are mutation-killed by name, every error path fires before any state mutation so there is no half-keyed channel, and the DH round-trip tests still prove shared-secret agreement through the real generator path.

It confirmed the correction already pushed — the TLS claim was false — and sharpened two details worth having right:

- There are **two** TLS construction sites, not four. The four I counted were `accept_invalid_certs: false` settings, which is a different thing.
- The `&parts[2..]` slice panics below **two** parts, not three, and that state cannot reach it: all three sites gate on message type 533 first, and that type parses out of `parts[1]`. So the reachable abort was `fields[0]` and the two `unwrap`s alone. The conclusion was right and the mechanism attribution was not; `.get(2..)` is belt and braces rather than the fix.

Both stated gaps stand: the signature change is source-breaking on an exported type, and the call sites' propagation is not covered because they live inside `connect`.

804 pass plus the two `config::expiry_tests` timezone failures. All offline targets clean.

userFRM:

PR #356, opened 2026-07-30
ccp: report the time-in-force that was submitted, not a plausible default (ibx#307)

### LINKED ISSUE #307: ccp: an unrecognised or absent time-in-force is reported as DAY, and the tracked value is never consulted
## Problem

The inbound order-status decode maps tag 59 with a catch-all that produces a valid value:

https://github.com/deepentropy/ibx/blob/main/src/engine/hot_loop/ccp.rs#L946

```rust
let tif_str = match tif_tag {
    "0" => "DAY", "1" => "GTC", "3" => "IOC", "4" => "FOK",
    "2" => "OPG", "6" => "GTD", "8" => "AUC", _ => "DAY",
};
```

An unrecognised code — or an absent tag 59, which leaves `tif_tag` empty and takes the same arm — is reported as `DAY`.

There is a fallback that would have covered this, but it cannot run:

https://github.com/deepentropy/ibx/blob/main/src/engine/hot_loop/ccp.rs#L1040

```rust
tif: if tif_str.is_empty() { fb_tif.to_string() } else { tif_str.to_string() },
```

`fb_tif` is `decode_tif(ctx_order.tif)` — the time-in-force from the locally tracked order, i.e. what the caller actually submitted. Because every arm of the match above yields a non-empty string, `tif_str.is_empty()` is never true, so that branch is unreachable and the tracked value is never consulted.

Note the sibling line does work: `order_type_str` can be empty, because its match ends `_ => ord_type_tag`, so `fb_ord_type` is reachable. Only the TIF fallback is dead.

## Impact

An order submitted with a time-in-force this mapping does not know reads back as `DAY` from `req_open_orders` and from order-status callbacks. `DAY` is a perfectly ordinary value, so a caller reconciling its own orders against the gateway sees a plausible answer that disagrees with what it sent, and nothing reports a mismatch.

The consequence is worst where the difference matters most: an order the caller believes is resting until cancelled, reported as expiring at the close.

## Fix

Let the match fall through to empty for an unknown code, so the existing fallback fires and the tracked value is used — `_ => ""`, matching the shape `order_type_str` already uses. That also makes an absent tag 59 resolve to what the caller submitted rather than to a guess.

If a genuinely unknown code should be surfaced rather than silently replaced, returning the raw tag would at least be visibly wrong instead of plausibly wrong.

## Separately, on the same path

`SubmitBracket` hardcodes `59="1"` (GTC) on both children at `order_builder.rs:670` and `:698` while the parent gets `59="0"` (DAY) at `:645`, and the children carry no attribute tags at all. A parent that expires at the close would leave its take-profit and stop-loss working into the next session against a position that no longer exists.

I have not filed that as its own issue because `SubmitBracket` appears unreachable from the public API — `place_order` builds brackets through `parent_id`/`oca_group` on the `SubmitEx` path, which does carry attributes. Worth confirming before anyone wires it up.

userFRM:
## Problem

The wire match on tag 59 ended in `_ => "DAY"`, so every path produced a non-empty string and the fallback that knows what this client actually submitted could never run.

An unrecognised code and an absent tag alike were reported as `DAY`. That is an ordinary value, so a caller reconciling its own orders got a plausible answer that disagreed with what it sent, and nothing said so.

## What this changes

**An absent tag leaves the string empty**, which is what makes the fallback to the tracked order reachable.

**A stated but unmapped code is reported as stated.** The gateway is authoritative when it says anything, and a code this does not name is still better seen than replaced by an unrelated local value. That mirrors the order-type sibling directly above it.

**A recovery record with no tag 59 is recorded as unstated, not guessed.** The order was not placed by this session, so there is nothing to recover it from. Either guess becomes a real instruction on the next replace — tag 59 on a `35=G` sets what the gateway is holding — and an invented `DAY` would expire an order that is resting until cancelled. A replace of an order with no stated time-in-force omits tag 59 and leaves the resting value in force.

`decode_tif` already maps anything it does not recognise to the empty string, so an unstated time-in-force reports as unstated rather than as an ordinary value.

## Tests

- `an_unknown_time_in_force_falls_back_to_the_one_that_was_submitted` — absence uses the submitted value; a stated code is reported as stated.
- `a_recovered_order_without_a_time_in_force_states_none` — the recovered order reports unstated, and a replace of it carries no tag 59.

Both fail by name against a compiling reversion of the production change they cover.

Closes #307.

## Test plan

- [x] Mutation: restoring the recovery insert's `b'0'` fallback fails `a_recovered_order_without_a_time_in_force_states_none` by name.
- [x] Mutation: removing the tag-59 removal on replace fails the same test — the mutant frame carries `59=` with the raw sentinel, which is worse than the value it replaced.
- [x] `an_unknown_time_in_force_falls_back_to_the_one_that_was_submitted` covers both directions: absence uses the submitted value, a stated code is reported as stated.

### DIFF
--- a/src/engine/hot_loop/ccp.rs
+++ b/src/engine/hot_loop/ccp.rs
@@ -698,7 +698,14 @@ impl CcpState {
                 .map(|p| (p * PRICE_SCALE as f64) as i64)
                 .unwrap_or(0);
             let ord_type_byte: u8 = parsed.get(&40).and_then(|s| s.bytes().next()).unwrap_or(b'2');
-            let tif_byte: u8 = parsed.get(&59).and_then(|s| s.bytes().next()).unwrap_or(b'1');
+            // A recovery record with no tag 59 states no time-in-force, and this
+            // order was not placed by this session, so there is nothing to
+            // recover it from. Recorded as unstated rather than guessed: either
+            // guess is restated as a real instruction on the next replace, and
+            // an invented DAY would expire a resting GTC order at the close.
+            let tif_byte: u8 = parsed.get(&59)
+                .and_then(|s| s.bytes().next())
+                .unwrap_or(crate::types::TIF_UNSTATED);
             if con_id != 0 && qty > 0 {
                 let instrument = context.register_instrument(con_id);
                 if let Some(sym) = parsed.get(&55) {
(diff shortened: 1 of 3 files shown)


userFRM:

Issue #219, opened 2026-07-10
connection: CCP liveness declared dead ~21s with 1s grace; no post-connect warm-up window


deepentropy:
### Current behavior
IBX relies solely on FIX-session heartbeats (`35=0`/`35=1`). It sends a test request when a heartbeat is `> CCP_HEARTBEAT + GRACE = 11 s` overdue and declares the CCP dead when a pending test is `> 10 s` old (~21 s total): `src/engine/hot_loop/mod.rs:24-28`, `:576`, `:578`. `HEARTBEAT_GRACE_SECS` is `1`. There is no post-connect warm-up window. Two unused duplicate heartbeat constants also exist in `src/config.rs:22-23`.

### Expected behavior
The gateway's transport-level liveness sends a heartbeat when idle >= 10 s, a test request when nothing is received for > 15 s, and only resets the connection when nothing is received for > 35 s, with a 60 s post-connect warm-up before any of this is enforced.

### Impact
IBX can declare the session dead at ~21 s versus the gateway's ~35 s, and with only a 1 s grace there is no warm-up window, so early-connection jitter can trigger a false disconnect during a period the server treats as grace.

### Suggested fix
Align the liveness thresholds (10/15/35 s) and add a 60 s post-connect warm-up before enforcing them. Remove or reconcile the unused duplicate constants in `config.rs`.

deepentropy:
Fixed in 42251f9. Liveness now matches the thresholds you documented: a test request after 15s of silence, dead only after 35s of silence (was ~21s, racing the server's own reset and losing to stalls it tolerates), and a 60s post-connect warm-up during which liveness is not enforced — heartbeats still flow, so a warm connection behaves identically. Applies to all three connections; send cadences are unchanged (10s/30s). The duplicate heartbeat-interval constants are gone: the hot loop now aliases the single definitions in config.

Validated live: the forced-disconnect farm reconnect test passes with the new thresholds active, and 100 hot-loop iterations right after connect (inside the warm-up) produce no false disconnect.

PR #357, opened 2026-07-30
gateway: reconnect to the trading farm the auth server named (ibx#295)

### LINKED ISSUE #295: gateway: the trading-farm reconnect announces the literal "usfarm", so a regional account reconnects under the wrong farm name
## Problem

The initial connect parses the trading farm out of the auth-server route and connects with it:

https://github.com/deepentropy/ibx/blob/main/src/gateway.rs#L1613

```rust
let (trading_host, trading_farm) = parse_farm_route(&trading_route)
```

The reconnect path does not. It passes a literal:

https://github.com/deepentropy/ibx/blob/main/src/engine/hot_loop/mod.rs#L882

```rust
let result = connect_farm(
    &auth.host, "usfarm",
    &auth.username, &auth.password, auth.paper,
```

`ReconnectAuth` carries the historical-data route across the reconnect, but has no field for the trading farm:

https://github.com/deepentropy/ibx/blob/main/src/gateway.rs#L400

```rust
/// Historical-data farm routing parsed from the auth-server response.
/// Used by HMDS reconnect (ibx#187) — empty when no HMDS route was parsed.
pub hmds_host: String,
pub hmds_farm: String,
```

So the HMDS side of a reconnect uses the account's real farm and the trading side uses a US name regardless of region. The farm name goes on the wire as part of tag 96 (`S<user>/<slot>/<farm>`), so a regional account announces a farm it is not on.

## Impact

The initial connection is fine, which is what makes this awkward to spot: a regional account trades normally until the first farm drop, and the reconnect then logs on under a name that does not match its route.

I have not observed what the server does with a mismatched farm name — I do not have a regional account to test against, so I am not claiming a specific failure mode. What is certain is that the reconnect announces something different from what the initial connect announced, for any account whose trading farm is not `usfarm`.

Same theme as #253, which was the channel role being derived from the farm name; this is the farm name itself not surviving a reconnect.

## Fix

Carry the trading farm on `ReconnectAuth` the way `hmds_farm` already is, and pass it at `mod.rs:882`. The value is in hand at `gateway.rs:1613` where the HMDS equivalent is already captured.

Found while auditing #255; unrelated to that change, which only touches the channel role.

### LINKED ISSUE #378: reconnect: the Rust client never supplies the credentials the scheduler requires, so no transport ever reconnects
## Problem

The Rust `EClient` never populates the credentials the reconnect scheduler requires, so no transport ever reconnects.

`Gateway::into_hot_loop_with_farms` builds a `ReconnectAuth` with `host`, `username` and `password` left empty — the comment on the struct says `// Filled by caller`. The Python path fills them (`src/python/compat/client/mod.rs:154`):

```rust
hot_loop.update_reconnect_auth(connect_host, connect_username, connect_password, connect_paper);
```

`EClient::connect_inner` (`src/api/client/mod.rs:197`) does not. It has all four values — they are on the `EClientConfig` it just used to authenticate — and simply never passes them on.

Every reconnect scheduler then refuses to run. `maybe_spawn_farm_reconnect` gates on a non-empty `auth.host`:

```rust
let auth = match self.reconnect_auth.as_ref() {
    Some(a) if !a.host.is_empty() => a,
    _ => return,
};
```

CCP and HMDS gate the same way.

## Impact

A Rust caller that loses the market-data farm, the CCP link or the historical connection stays disconnected for the life of the process. No attempt is scheduled, no backoff runs, and nothing is reported — the client reports connected while quotes stop and orders cannot be sent.

Python is unaffected, so this is a divergence between the two bindings on the recovery path rather than a shared gap. Everything the reconnect machinery does — the jittered backoff of #218, the warm-up suppression of #219 — is dead code for Rust callers today.

## Solution

Call `update_reconnect_auth` from `connect_inner` with the same four values Python passes.

## Notes

Read from v0.7.1 (9367845). Verifiable offline: `grep -rn update_reconnect_auth src/` returns the definition and exactly one caller, in the Python compat layer.

userFRM:
## Problem

The auth server names the trading farm to use, and the reconnect path re-derived one instead of keeping it — so a session routed to a regional farm reconnected somewhere else.

## What this changes

`ReconnectAuth` carries the trading host and farm as the auth response gave them, and the reconnect selects that route rather than re-deriving one.

## The routing was unreachable

Nothing selected it. The reconnect credentials were left blank at construction for the caller to fill, and only the Python binding filled them:

```rust
host: String::new(), // Filled by caller (Python EClient or Rust API)
```

`EClient::connect_inner` never did. Every scheduler gates on a non-empty `auth.host`, so **no Rust caller reconnected on any transport** — farm, CCP or HMDS. The jittered backoff of #218 and the warm-up suppression of #219 were dead code for Rust callers.

The credentials are now a constructor parameter rather than a field set afterwards. `into_hot_loop_with_farms` takes a `CallerAuth`, both bindings pass one, and the setter that could be forgotten is gone — a binding that omits them does not compile. That is what closes the class rather than the instance; the compiler found two further construction sites in the bench harnesses while this was applied.

## Breaking change to the low-level API

`Gateway::into_hot_loop` and `into_hot_loop_with_farms` take a `CallerAuth`. `HotLoop::update_reconnect_auth` is removed — pass the values at construction instead. `Client`/`EClient` and the Python surface are unchanged.

Closes #295. Closes #378.

## Test plan

- [x] The credentials being a constructor parameter is enforced by the compiler, not by a test: it found two further construction sites in the bench harnesses while this was applied.
- [x] `tests/farm_reconnect_poc` updated to the new signature and checks clean.

### DIFF
--- a/bench/bench_harness.rs
+++ b/bench/bench_harness.rs
@@ -95,7 +95,13 @@ impl BenchSession {
         let shared = Arc::new(SharedState::new());
         let (event_tx, event_rx) = bounded::<Event>(65536);
         let (mut hot_loop, control_tx) =
-            gw.into_hot_loop(shared.clone(), Some(event_tx), farm_conn, ccp_conn, hmds_conn, None);
+            gw.into_hot_loop(shared.clone(), Some(event_tx), farm_conn, ccp_conn, hmds_conn, None,
+                ibx::gateway::CallerAuth {
+                    host: gw_config.host.clone(),
+                    username: gw_config.username.clone(),
+                    password: gw_config.password.clone(),
+                    paper: gw_config.paper,
+                });
 
         let join = std::thread::spawn(move || {
             hot_loop.run();
--- a/examples/ex151_srp_fallback.rs
+++ b/examples/ex151_srp_fallback.rs
@@ -53,6 +53,8 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
         encoded: gw.encoded.clone(),
         hmds_host: gw.hmds_host.clone(),
         hmds_farm: gw.hmds_farm.clone(),
+        trading_host: String::new(),
+        trading_farm: String::new(),
     };
 
     drop(ccp);
--- a/src/api/client/mod.rs
+++ b/src/api/client/mod.rs
@@ -196,6 +196,12 @@ impl EClient {
 
         let (hot_loop, control_tx) = gw.into_hot_loop_with_farms(
             shared.clone(), event_tx, farm_conn, ccp_conn, hmds_conn, config.core_id,
+            crate::gateway::CallerAuth {
+                host: config.host.clone(),
+                username: config.username.clone(),
+                password: zeroize::Zeroizing::new(config.password.clone()),
+                paper: config.paper,
+            },
         );
 
         let handle = thread::Builder::new()
--- a/src/bin/benchmark.rs
+++ b/src/bin/benchmark.rs
@@ -185,6 +185,12 @@ fn main() {
     let (event_tx, event_rx) = bounded::<Event>(65536);
     let (mut hot_loop, control_tx) = gw.into_hot_loop(
         shared, Some(event_tx), farm_conn, ccp_conn, hmds_conn, None,
+        ibx::gateway::CallerAuth {
+            host: config.host.clone(),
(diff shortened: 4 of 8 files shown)


userFRM:

PR #358, opened 2026-07-30
tick decode: step over a tick too wide to read, rather than dropping the message (ibx#302)

### LINKED ISSUE #302: tick decode: an extended tick with a width above 8 silently discards the rest of the 35=P message
## Problem

In the `35=P` bit-stream decoder, a 5-bit tick type of 31 selects an extended entry. The width is then read as a full byte:

https://github.com/deepentropy/ibx/blob/main/src/protocol/tick_decoder.rs#L243

```rust
byte_width = match reader.read_unsigned(8) {
    Some(v) => v,
    None => return,
};
```

giving a value of 0..255, which becomes a bit count of up to 2040:

https://github.com/deepentropy/ibx/blob/main/src/protocol/tick_decoder.rs#L252

```rust
let total_value_bits = (8 * byte_width) as usize;
if reader.remaining() < total_value_bits {
    return;
}
```

Both that guard and `read_unsigned` itself — which returns `None` for any `n > 64`, so any width above 8 — end the whole decode with a bare `return`.

## Impact

A `35=P` message carries entries for many instruments in one bit-stream. Returning at the first entry the decoder cannot represent discards **every remaining entry in that message**, for every instrument after it, with no error and no log line. The caller sees a shorter batch and cannot tell it was truncated.

That makes it the same failure shape as #281: quotes simply stop updating for whatever came later in the batch, while the connection stays healthy. Unlike #281 it would be intermittent, tracking whatever instrument happens to sit later in the stream.

## Worth checking

An eight-bit width is a wide field for a value the same decoder otherwise caps at eight. It is worth confirming from a capture whether those eight bits are one field at all — if the high nibble carries something else (a decimal exponent would fit the shape, since a self-describing per-tick scale would explain fractional quantities), then every extended tick with a non-zero high nibble is currently being read as an enormous width and truncating the message. When the high nibble is zero the two readings agree exactly, which would explain why this has not surfaced.

The observation that settles it: find a record whose 5-bit type field reads 31, then decode the following byte both ways — as one 8-bit width, and as two nibbles. Only one of them leaves the remaining entries landing exactly on a record boundary with the has-more bit clear.

## Fix

Independent of that question, the decoder should not treat an entry it cannot represent as end-of-message. Skipping the entry — or at minimum logging before returning — keeps one unrepresentable tick from silently deleting the rest of the batch.

userFRM:
## Summary

- An extended `35=P` entry states its width in a full byte, so it can name a value wider than this decoder returns. The decoder abandoned the whole message at that point, so every tick after it went with it — including the other server tags in the same message.
- The field is now stepped over and decoding continues. The entry itself is still dropped rather than guessed at, and a genuinely truncated message still ends where it did.

## Why

A quote update sitting behind one of these simply never arrived, and nothing recorded that it had. One entry the decoder cannot represent should cost that entry, not the rest of the message.

Closes #302.

## Notes

Stepping over needs a `skip` on the bit reader. Its width comes off the wire, so it is bounded by what is actually left rather than by an addition that wraps in release and rewinds the reader instead of stopping it. It is private — only this decoder steps.

The test payload builder gained the ability to emit a field wider than a `u64`, which is what building one of these requires.

## Test plan

- [x] `cargo test --offline --lib` — 804 passed. The 2 failures are `config::expiry_tests::{named_zone_converts_with_dst, instant_round_trips_to_wire}`, which fail on the base commit too (fixed separately in #336).
- [x] `tests/ib_paper_compat` unchanged from base — identical sorted diagnostic sets, not merely equal counts.
- [x] Mutation check: restoring the bare `return` fails `a_tick_too_wide_to_read_does_not_discard_the_rest_of_the_message`; making `skip` add unchecked fails `skipping_past_the_end_refuses_rather_than_wrapping`. Both by name.
- [x] The test asserts the ticks after the wide entry **and** the whole server tag after it, so it distinguishes "kept going" from "stopped one entry later".

### DIFF
--- a/src/protocol/tick_decoder.rs
+++ b/src/protocol/tick_decoder.rs
@@ -30,6 +30,20 @@ impl<'a> BitReader<'a> {
         self.total_bits.saturating_sub(self.bit_pos)
     }
 
+    /// Advance past `n` bits without decoding them. Used to step over a field
+    /// this decoder cannot represent, so the rest of the message still decodes.
+    ///
+    /// Compares against `remaining`, which saturates, so a width read off the
+    /// wire cannot overflow the position.
+    #[inline]
+    fn skip(&mut self, n: usize) -> bool {
+        if n > self.remaining() {
+            return false;
+        }
+        self.bit_pos += n;
+        true
+    }
+
     /// Read n bits as unsigned integer (MSB first).
     /// Uses word-aligned reads for performance (1-3 ops instead of n iterations).
     #[inline]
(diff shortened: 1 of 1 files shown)


userFRM:

PR #359, opened 2026-07-30
orders: assert that outside-RTH is absent when it was not asked for (ibx#352)

### LINKED ISSUE #352: tests: the outside-RTH assertions are one-sided, so a path that always sets the flag passes
## Problem

Making every submit encoder emit `6433=1` unconditionally fails no test.

The assertions check that the flag is **present when the caller asked for it**. None checks that it is **absent when they did not**. A mutation that always sets it therefore satisfies every one of them.

## Impact

This is the shape of #247: the replace carried a hard-coded `6433=1`, so any order placed regular-hours-only became eligible outside them on its first modify. That survived because the tests could not express the direction it broke in. The submit paths still have the same blind spot, so the same defect can be reintroduced there without any test noticing.

An order silently widened to the extended session is filled at prices the caller never intended to trade at, and nothing in the callbacks distinguishes it from an order they asked for.

## Solution

Assert both polarities wherever the flag is emitted: present with the captured position when set, and absent when not. An absence assertion needs a positive control beside it or it passes just as well against an encoder that emits nothing at all.

Worth applying the same check to the other conditional attribute tags, which are written the same way.

userFRM:
## Summary

- Every submit encoder guards tag 6433, and nothing asserted the guard. The tests checked that the flag is present when the caller set it, never that it is absent when they did not — so making all four paths emit it unconditionally failed no test at all.
- One test, both polarities, across all four paths that emit the tag: the three GTC variants and the shared extended encoder every attributed order routes through.

## Why

This is the shape #247 took on the replace path, where a hard-coded `6433=1` opted every modified order into the extended session. It survived because the coverage could not express the direction it broke in — and the same blind spot is still on the submit side.

An order silently widened to outside regular hours fills at prices the caller never meant to trade at, and nothing in the callbacks distinguishes it from one they asked for.

The test reads the bytes off a socket rather than inspecting the request enum, because the enum is not where the tag is decided.

**No production change**: the guards were already correct. This is the assertion that keeps them so.

Closes #352.

## Test plan

- [x] `cargo test --offline --lib` — 803 passed. The 2 failures are `config::expiry_tests::{named_zone_converts_with_dst, instant_round_trips_to_wire}`, which fail on the base commit too (fixed separately in #336).
- [x] Mutation check: making all four paths emit `6433` unconditionally fails `every_submit_path_emits_outside_rth_only_when_it_was_asked_for` by name — and so does mutating **only** the shared extended encoder, so the test is not passing on the strength of one path.

### DIFF
--- a/src/engine/hot_loop/order_builder.rs
+++ b/src/engine/hot_loop/order_builder.rs
@@ -2057,3 +2057,68 @@ mod tests {
         assert!(shared.orders.drain_order_updates().is_empty());
     }
 }
+
+#[cfg(test)]
+mod outside_rth_polarity_tests {
+    use super::*;
+    use crate::protocol::connection::Connection;
+    use std::io::Read;
+
+    /// Drive the queued orders and read what actually reaches the socket.
+    fn drain(context: &mut Context) -> String {
+        let listener = std::net::TcpListener::bind("127.0.0.1:0").unwrap();
+        let addr = listener.local_addr().unwrap();
+        let client = std::net::TcpStream::connect(addr).unwrap();
+        let (mut peer, _) = listener.accept().unwrap();
+        peer.set_read_timeout(Some(std::time::Duration::from_secs(5))).unwrap();
+        let mut conn = Some(Connection::new_raw(client).unwrap());
+
+        let shared = std::sync::Arc::new(SharedState::new());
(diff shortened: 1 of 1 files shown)


userFRM:

PR #360, opened 2026-07-30
bridge: a completed order does not return to the open book (ibx#262)

### LINKED ISSUE #262: orders: a replayed frame resurrects a filled order into the open book, in both the engine and the API cache
## Problem

Nothing remembers that an order reached a terminal state, so a later frame can put it back into the open book. Two surfaces, one root cause.

**The engine book.** A fill removes the order (`src/engine/hot_loop/ccp.rs:1121`). A subsequent `150=0/39=0` frame carrying 6008 and 38 finds `context.order(id)` empty and takes the recovery insert at `:684`, re-adding it as `Submitted` with `filled: 0`. The resurrected entry also blocks `try_reclaim_instrument` (`mod.rs:257`), so the instrument slot leaks with it.

**The API cache.** The enrich block at `:900` runs on the current frame's status and does not consult the #212 guard's `status_changed` verdict at `:847`, so it writes `RichOrderInfo { status: "Submitted" }` over the existing `"Filled"` entry. `push_order_info` (`src/bridge.rs:331-333`) is a bare `insert` with no ordering rule:

```rust
pub fn push_order_info(&self, order_id: u64, info: RichOrderInfo) {
    self.order_cache.lock().unwrap().insert(order_id, info);
}
```

`drain_open_orders` (`bridge.rs:290`) then filters on that string and reports a filled order as working.

The trigger is ordinary: the reconnect open-order burst racing a fill, or any frame the gateway replays.

## Impact

A filled order reappears as live. A strategy re-manages a position it already has, or cancels an order that no longer exists, and `req_open_orders` corroborates the wrong picture. Because #212 already guards the *status* transition, the surviving path is the cache write, which looks correct in isolation.

## Solution

Put the ordering rule where both surfaces already route: `push_order_info` refuses to overwrite a terminal cached status with an open one. That is one guard covering the API cache regardless of which caller writes it.

Gate the recovery insert on the same knowledge, so a completed order is not re-inserted into the engine book by a replayed frame — keeping a terminal marker for recently-completed ids is enough, and #212 established that the engine already needs to reason about status ordering rather than last-write-wins.

## Notes

Read from v0.7.1 (9367845). Both writes are unconditional in the source; the race that produces the stale frame is inferred from the reconnect replay behaviour rather than captured.

userFRM:
## Problem

Nothing remembered that an order had reached a terminal state, so a replayed frame put it back into the open book.

`push_order_info` was a bare insert with no ordering rule, and `drain_open_orders` filters on the string it writes. The reconnect open-order burst racing a fill — or any message the gateway resends — made `req_open_orders` report a completed order as live. A strategy then re-manages a position it already has, or cancels an order that no longer exists, with the open-order snapshot corroborating the wrong picture.

## Why the cached status cannot carry it

Two ways a status-only guard is bypassed:

- **Completing an order evicts its cache row.** The replayed frame finds nothing to refuse and inserts itself. That is the ordinary path, not an edge case.
- **A cached string is overwritten by the next terminal report.** A report landing between the completion and the replay erases the evidence, and the replay then passes.

## What this changes

A bounded set of completed order ids carries the knowledge instead, recorded where the completed-order record is already written and surviving the eviction. It is capped and oldest-first, because a session completes orders indefinitely; the width only has to outlive a reconnect's replay of recent activity, which is what resurrects an order.

The rule covers every status the engine treats as ending an order's life — `Filled`, `Cancelled` and `Rejected` are the three it acts on by removing the order from its book. `Inactive` is not among them, because it returns to working when the condition holding the order clears, and neither is `Uncertain`, which states the opposite of a conclusion.

A **trade cancel (`150=H`) or trade correction (`150=G`)** restates an execution the gateway has already reported and can legitimately return a filled order to a working quantity. That is the gateway's own statement, not a replay of an older one, so those two execution types write through a separate entry point that supersedes the completion rather than being refused by it.

## Tests

- `a_completed_order_is_not_returned_to_the_open_book` — all three terminal statuses, all four open ones.
- `a_completion_outlives_the_cache_row_it_evicts` — the ordinary completion path.
- `an_intervening_report_does_not_erase_the_completion` — `Filled → Cancelled → Submitted`.
- `a_trade_correction_can_reopen_a_completed_order` — and the order stops being remembered afterwards.
- `a_fill_still_writes_over_a_working_status` — the ordinary direction, so the guards are not passing against a cache that refuses everything.
- `the_completed_memory_is_bounded` — capped, oldest evicted first.

Closes #262.

## Test plan

- [x] Mutation: removing the completed-id memory fails `a_completion_outlives_the_cache_row_it_evicts` by name.
- [x] Mutation: narrowing the terminal set to `Filled` fails `a_completed_order_is_not_returned_to_the_open_book`.
- [x] Mutation: leaving the id remembered after a correction fails `a_trade_correction_can_reopen_a_completed_order`.
- [x] Mutation: removing the cap fails `the_completed_memory_is_bounded`.
- [x] Positive control: `a_fill_still_writes_over_a_working_status`, so the guards are not passing against a cache that refuses everything.

### DIFF
--- a/src/bridge.rs
+++ b/src/bridge.rs
@@ -11,6 +11,7 @@
 //! - External callers read snapshots and poll events without blocking the hot loop.
 
 use std::sync::atomic::{AtomicBool, AtomicU64, Ordering};
+use std::time::{Duration, Instant};
 use std::sync::{Condvar, Mutex};
 use std::cell::UnsafeCell;
 
@@ -249,7 +250,31 @@ pub struct OrderState {
     completed_orders: Mutex<Vec<CompletedOrder>>,
     /// Enriched order info from CCP exec reports (order_id -> RichOrderInfo).
     order_cache: Mutex<HashMap<u64, RichOrderInfo>>,
+    /// Orders that reached a terminal state, and when. The cache row is evicted
+    /// when an order completes, so the cached status alone cannot say an order
+    /// is done — a replayed frame would find nothing to refuse and insert it as
+    /// open.
+    completed: Mutex<HashMap<u64, Instant>>,
 }
+
+/// How long a completion is remembered.
+///
+/// Held by age rather than by count. What this has to outlive is the window in
+/// which a stale frame for the order can still arrive — a reconnect replays
+/// recent activity within seconds — and that window is a duration, not a number
+/// of orders. Counting instead meant a busy session's unrelated completions
+/// pushed a still-relevant one out, and the replay it was there to refuse got
+/// back in.
+const COMPLETED_RETENTION: Duration = Duration::from_secs(300);
+
+/// Hard cap on how many completions are remembered at once, regardless of
+/// age. Expired entries are pruned first; a session that completes orders
+/// faster than they expire would otherwise leave every young entry in
+/// place, so once pruning alone cannot bring the map back under this bound,
+/// the oldest survivors are evicted until it does. Generous enough that
+/// reaching it at all means completions are arriving far faster than any
+/// legitimate replay could still be racing the ones being dropped.
+const COMPLETED_MAX: usize = 65_536;
 
 impl OrderState {
     fn new() -> Self {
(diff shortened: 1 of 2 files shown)


userFRM:

PR #361, opened 2026-07-30
ccp: read the resend markers and book replays on cumulative quantity (ibx#320, ibx#260)

### LINKED ISSUE #260: ccp: duplicate-fill dedup is skipped when ExecID is absent, so a replayed fill double-counts the position
## Problem

The duplicate-fill guard from #198 is skipped entirely when the frame carries no ExecID (`src/engine/hot_loop/ccp.rs:851`):

```rust
    return;
}
```

The `!exec_id.is_empty() &&` short-circuits, so a frame with tag 17 absent or empty never enters the dedup window at all. A replayed fill is then applied a second time: the position is updated twice and a second `Fill` event is emitted.

The absent-tag case is not exotic — it is the shape a replay or a partially-populated frame takes, and the dedup guard exists precisely because the gateway does resend fills.

Related, same block: on a genuine duplicate the handler returns at `:853`, above the terminal cleanup at `:1107`. A duplicate frame carrying `39=2` therefore leaves the entry in `context.open_orders` and never pushes the completed order — the dedup skips the fill and the bookkeeping with it.

## Impact

Position accounting silently doubles for the affected fill, and a strategy reading `Event::Fill` acts twice on one execution. Everything downstream — risk checks, P&L, the position feed — inherits the wrong number, and nothing surfaces an error.

## Solution

Do not skip the window when the ID is missing: synthesise a key from the fields that identify an execution — `(clord_id, tag 60, tag 32, tag 31)` — and record that instead, so a frame without tag 17 is deduped on its content rather than trusted.

Separately, move the terminal cleanup above the duplicate return, or run it before returning, so a duplicate carrying a terminal status still completes the order.

## Notes

Read from v0.7.1 (9367845). The short-circuit is plain in the condition; the replay frequency of ExecID-less frames is inferred from the adversary model rather than observed on a live session.

### LINKED ISSUE #320: ccp: on a fresh process the recovery insert tracks a replayed order, so its replayed execution books as a new fill

### LINKED ISSUE #344: ccp: the ExecID window is 1024 entries, so a replay batch deeper than that double-counts its own head
## Problem

Fill deduplication keys on a window of the most recent ExecIDs:

```rust
const EXEC_ID_WINDOW: usize = 1024;
```

Entries are evicted oldest-first. If a duplicate arrives more than 1024 distinct executions after its first copy, the window no longer holds it and the fill is booked a second time — the position double-counts.

Verified with a probe driving two copies of one execution separated by more than 1024 distinct ones: booked twice, position moved twice. The same probe behaves identically against the code before #331, so this is the bound itself rather than anything that change introduced.

## When it is reachable

The window is per engine and deliberately survives a reconnect, so within a session it normally covers the replay burst. It stops covering it when the replay is deeper than the window.

Two things narrow it further:

- a fresh process replays prior executions carrying `97=Y`, and the untracked-fill path refuses those — but only for orders that stay untracked
- for an order that **is** tracked, the window is the only guard

So the exposure is a same-session replay deeper than 1024 between the two copies of one execution. #317 shows the gateway currently delivers each replayed message twice; if that doubling is batch-shaped — the whole batch, then the whole batch again — rather than adjacent, any replay batch over 1024 double-books its head.

## Solution

Raise the window above the gateway's replay depth. Each entry is on the order of fifty bytes, so 8k or 16k entries is a few hundred kilobytes and removes the cliff for any plausible batch.

Keying deduplication on something bounded by the account rather than by a fixed count would be better still, but the constant is what makes the failure reachable today.

userFRM:
## Problem

At session start the gateway replays recent activity. A New/New report inserts the order through the cross-session recovery path, then its executions arrive behind it carrying their original ExecIDs and a resend marker. By then the order is tracked, so each execution books through the normal path — a fill event, a position move — for something that happened before the process started. The ExecID window cannot stop it: a fresh process has never seen those IDs. Neither `97` (PossResend) nor `43` (PossDupFlag) was read anywhere in the handler.

## What this changes

**The recovery insert seeds `filled` from the record's own CumQty** instead of assuming zero, so a recovered order arrives with the quantity it has actually filled rather than looking untouched.

**A marked report books on what it adds to that baseline, not on the increment it reports.** Its CumQty is the order's cumulative filled quantity as of that execution, so what is news to this session is CumQty minus what the order already holds — nothing when it restates. Reading the cumulative figure is what makes replay safe in general:

- the same report twice adds nothing the second time;
- copies arriving out of order settle on the true total instead of losing whichever lands behind a later one;
- a session long enough to have rolled those executions out of the ExecID window is not overcounted.

Live reports are unchanged: an increment is new quantity by definition, and the window is what stops a repeat.

**A marked report is not vetoed outright.** A CCP reconnect keeps this state and the gateway replays recent executions on the new session, so a fill that executed during the outage arrives marked with an ID this session has never seen and is the first news of it. It carries cumulative quantity the order lacks, so it books.

**An execution with no ExecID is deduplicated on its content.** It skipped the window entirely, so a replayed copy booked twice — the shape a replay takes, and precisely when the window matters. The fallback key includes CumQty, the one field that separates two slices of the same size at the same price inside one timestamp tick, including across a replacement that raised the total and left the same quantity behind twice. The key is spent only where the fill can actually book, so an execution arriving ahead of the recovery record for its order is not refused when that record finally lands.

**`update_order_filled` saturates** rather than wrapping, so a quantity large enough to overflow the counter cannot take an order's filled quantity round to nothing.

## Tests

- `a_resent_execution_does_not_book_a_fill` — both markers, with a live positive control.
- `a_fresh_process_does_not_book_the_history_it_is_replayed` — end to end: recovery record then its replayed execution.
- `a_resent_execution_carrying_new_quantity_is_still_booked` — the fill that ran during an outage.
- `a_replay_of_booked_history_adds_nothing_to_the_order` — unordered, ExecID-less replay of already-booked executions, and a genuinely new one in the same batch.
- `a_marked_execution_delivered_twice_books_once`.
- `a_raised_total_does_not_collapse_two_slices_into_one`.
- `a_key_is_not_spent_before_the_order_exists`.
- `two_same_priced_slices_in_one_tick_are_not_one_execution`, `an_execution_without_an_exec_id_is_still_deduplicated`.
- `update_order_filled_saturates`.

No new wire traffic: this reads tags the gateway already sends.

**The dedup window stops being the guard** for a replayed execution, which is what made its depth a cliff (#344): a batch deeper than the window no longer held its own head, and for an order this session tracks the window was the only thing between the second copy and a second booking.

Closes #320. Closes #260. Closes #344.

## Test plan

- [x] Mutation: reverting the resend path to the live one fails `a_resent_execution_does_not_book_a_fill`, `a_fresh_process_does_not_book_the_history_it_is_replayed`, `a_marked_execution_delivered_twice_books_once` and `a_replay_of_booked_history_adds_nothing_to_the_order` by name.
- [x] Mutation: keying the fallback on `LeavesQty` instead of `CumQty` fails `a_raised_total_does_not_collapse_two_slices_into_one`.
- [x] Mutation: spending the key before the order lookup fails `a_key_is_not_spent_before_the_order_exists`.
- [x] Mutation: seeding the recovery insert with zero fails `a_fresh_process_does_not_book_the_history_it_is_replayed`.
- [x] Mutation: not recording a marked ExecID fails `a_marked_execution_is_remembered_for_its_unmarked_twin`.
- [x] `a_replay_deeper_than_the_exec_id_window_does_not_double_count` covers #344 — the window is no longer the guard for a replayed execution.

### DIFF
--- a/src/engine/context.rs
+++ b/src/engine/context.rs
@@ -958,7 +958,7 @@ impl Context {
 
     pub fn update_order_filled(&mut self, order_id: OrderId, last_shares: u32) {
         if let Some(order) = self.open_orders.get_mut(&order_id) {
-            order.filled += last_shares;
+            order.filled = order.filled.saturating_add(last_shares);
         }
     }
 
@@ -1531,6 +1531,21 @@ mod tests {
         assert_eq!(ctx.order(1).unwrap().filled, 30);
         ctx.update_order_filled(1, 50);
         assert_eq!(ctx.order(1).unwrap().filled, 80);
+    }
+
+    /// A gateway figure large enough to overflow the counter must not wrap the
+    /// order's filled quantity round to nothing.
+    #[test]
+    fn update_order_filled_saturates() {
+        let mut ctx = Context::new();
+        ctx.insert_order(Order {
+            order_id: 1, instrument: 0, side: Side::Buy,
+            price: PRICE_SCALE, qty: u32::MAX, filled: u32::MAX - 1,
+            status: OrderStatus::PartiallyFilled,
+            ord_type: b'2', tif: b'0', stop_price: 0,
+        });
+        ctx.update_order_filled(1, 10);
+        assert_eq!(ctx.order(1).unwrap().filled, u32::MAX);
     }
 
     #[test]
--- a/src/engine/hot_loop/ccp.rs
+++ b/src/engine/hot_loop/ccp.rs
@@ -710,7 +710,14 @@ impl CcpState {
                     side,
                     price: limit_price_i64,
                     qty,
-                    filled: 0,
+                    // Seeded from the recovery push rather than assumed zero.
+                    // Without it a fresh process believes nothing has filled,
+                    // and the replayed executions behind this record all look
+                    // like new quantity (ibx#320).
+                    filled: parsed.get(&14)
+                        .and_then(|s| s.parse::<f64>().ok())
+                        .map(|c| c as u32)
+                        .unwrap_or(0),
                     status: crate::types::OrderStatus::Submitted,
                     ord_type: ord_type_byte,
                     tif: tif_byte,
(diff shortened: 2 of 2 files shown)


userFRM:

PR #363, opened 2026-07-30
protocol: bound the socket writes and stop sending on a broken frame (ibx#254)

### LINKED ISSUE #254: engine: no write timeouts anywhere, so a stalled peer blocks the single hot-loop thread and wedges liveness, shutdown and reconnect
## Problem

No socket in the engine has a write timeout — `set_write_timeout` is never called anywhere in `src/`. Reads are bounded (`set_read_timeout` on the auth and farm-logon paths), writes are not.

The hot loop is a single thread (`poll_once`, `src/engine/hot_loop/mod.rs:186`) driving all three transports. Every outbound message goes out through a blocking `write_all` on that thread: order submits, cancels, subscriptions, and the heartbeats themselves.

If a peer stops draining — the receive window closes, the socket buffer fills, the far end is wedged rather than closed — `write_all` blocks with no bound. The single thread then cannot poll any of the three sockets, cannot evaluate the liveness deadlines, cannot process `Shutdown`, and cannot poll reconnect results.

The failure mode is self-defeating: the heartbeat exists to detect a dead peer, and it is the write that blocks on it. A half-dead connection is exactly the case the liveness machinery was built for, and it is the case where the detector cannot run.

## Impact

The engine wedges with no timeout, no event, and no disconnect — a caller sees a live client that has stopped doing anything, including reconnecting. Orders queued behind that write never go out and never error.

## Solution

Bound the writes. A timed-out `write_all` needs care rather than a retry: outbound frames are HMAC-chained, so a partial write leaves the transport's signature chain desynchronised. The safe treatment is to mark that transport unrecoverable on any incomplete or timed-out write and go straight to disconnect and reconnect, rather than attempting to resume mid-frame.

## Notes

Read from v0.7.1 (9367845): `grep -rn set_write_timeout src/` returns nothing, and the three connections are driven from one thread. Not reproduced against a live gateway — provoking it needs a peer that stalls without closing.

userFRM:
## Problem

No socket in the engine had a write timeout — `set_write_timeout` was called nowhere in `src/`. Reads were bounded; writes were not.

The hot loop is one thread driving all three transports, and every outbound message goes out through a blocking write on it: order submits, cancels, subscriptions, and the heartbeats themselves. A peer that stops draining without closing blocks that thread, so liveness cannot be evaluated, shutdown cannot be serviced, and a reconnect cannot be polled. The failure is self-defeating: the heartbeat exists to detect a dead peer, and it is the write that blocks on it.

## What this changes

**One construction path.** Both constructors reach the socket through `from_stream`, so a transport cannot be given a different set of timeouts by being built down a different path, and there is a single place the bound can be dropped from.

**A broken frame ends the transport; a stalled one does not.** Outbound frames are HMAC-chained, so a timed-out write needs more care than a retry. `write_frame` tracks how much of the frame reached the socket:

- **part of it went out** — the peer holds a prefix it cannot verify and every later frame would be signed from state it does not share. That connection is finished, and every subsequent send is refused rather than putting more on the wire.
- **none of it went out** — nothing reached the wire, the chain is intact, and the frame can go again. A peer that is merely slow does not cost the transport.
- **interrupted by a signal** — retried, counted as neither.

The bound is on a single write syscall rather than a whole frame, so a peer draining a trickle keeps resetting it. The case it exists for is the peer making no progress at all, which trips the first timeout; a peer that stalls indefinitely reaches the liveness deadlines, which is where it belongs.

**The keep-up-to-date historical request no longer advances its signing IV for a frame that never went out.** That IV is derived from the frame, so committing it before the send desynchronises the chain and the next authentic frame no longer verifies. It is now held across the send and committed only when the frame is away, along with the send timestamp the liveness clock reads.

## What this does not do

It does not treat an abandoned write as a disconnect. That belongs with the reconnect paths, and three of them do not currently support it — each reachable today through the read-side liveness path, none introduced here:

- HMDS sets its disconnected flag without releasing the connection, so `maybe_spawn_hmds_reconnect` returns early forever (#367);
- the farm's `handle_disconnect` clears `instrument_md_reqs`, the list its own `reconnect` resubscribes from (#368);
- a request whose write fails is still recorded as pending, with no response and no rejection (#369).

Until those are fixed, recovery stays on the existing route — the read deadlines — and the write bound is what stops the thread being held meanwhile.

## Tests

- `a_write_timeout_is_configured_on_both_constructors` — the bound is set, the read cadence is unchanged, and a fresh connection writes.
- `the_write_bound_stays_within_the_liveness_cadence` — the bound cannot outlast the liveness probe interval.
- `only_a_frame_that_partly_went_out_finishes_the_transport` — stall with nothing sent is retryable; the same error mid-frame is not; a hard error never is.
- `a_failed_write_finishes_the_transport` — `send_raw`, `send_fix` and `send_fixcomp` are all refused afterwards, and no sequence number is consumed.

Closes #254.

## Test plan

- [x] Mutation: removing `set_write_timeout` fails `a_write_timeout_is_configured_on_both_constructors` by name.
- [x] Mutation: raising the bound to 50s fails `the_write_bound_stays_within_the_liveness_cadence`.
- [x] Mutation: treating every error as recoverable fails `only_a_frame_that_partly_went_out_finishes_the_transport`.
- [x] Mutation: letting `send_fixcomp` bypass the latch, and letting the latch stop refusing, each fail `a_failed_write_finishes_the_transport`.
- [x] Both constructors reach the socket through one function, so the bound cannot be present on one transport and missing on another.

### DIFF
--- a/src/engine/hot_loop/hmds.rs
+++ b/src/engine/hot_loop/hmds.rs
@@ -775,13 +775,16 @@ impl HmdsState {
                 (6118, &xml),
             ], 0);
             let compressed = fixcomp::fixcomp_build(&raw);
-            let to_send = if !sign_key.is_empty() {
-                let mut iv_guard = sign_iv.lock().unwrap();
+            // Held across the send: the next IV is derived from this frame, so
+            // committing it for a frame the transport refuses or fails to put
+            // on the wire desynchronises the chain the peer verifies against,
+            // and the next authentic frame no longer matches (ibx#254).
+            let mut iv_guard = sign_iv.lock().unwrap();
+            let (to_send, next_iv) = if !sign_key.is_empty() {
                 let (signed, new_iv) = fix::fix_sign(&compressed, sign_key, &iv_guard);
-                *iv_guard = new_iv;
-                signed
+                (signed, Some(new_iv))
             } else {
-                compressed
+                (compressed, None)
             };
             // Debug: dump first 80 bytes hex for wire comparison
             let hex: String = to_send.iter().take(80).map(|b| format!("{:02x}", b)).collect();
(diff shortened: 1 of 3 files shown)


userFRM:

PR #366, opened 2026-07-30
protocol: compare the NS prefix in place rather than through an uppercased copy (ibx#365)

### LINKED ISSUE #365: protocol: an NS payload whose prefix changes length when uppercased panics the hot loop
## Problem

`ns_parse` (`src/protocol/ns.rs:57-62`) tests the prefix on an uppercased copy and then slices the original:

```rust
let text = if text.to_uppercase().starts_with("MISC") {
    &text[4..]
} else {
    text
};
```

Uppercasing is not length-preserving. `U+0131` (dotless i) and `U+017F` (long s) each occupy two bytes and uppercase to a single ASCII byte — `I` and `S`. So a payload beginning `mıſc` uppercases to `MISC`, passes the check, and occupies six bytes rather than four. Byte 4 then sits inside a character, and slicing a `&str` there is a panic.

Reproduced directly:

```
panicked at src/protocol/ns.rs:58:14: byte index 4 is not a char boundary;
it is inside 'ſ' (bytes 3..5) of `mıſc;1;2`
```

## Impact

The path is `ns_recv` → `parse_test_request_timestamp` → `ns_parse`: an inbound heartbeat probe, on the hot loop. So this takes the engine down rather than dropping a message.

Unlike #258 this is not a lossy-decode artefact — `ns_parse` takes `from_utf8().ok()?`, so the bytes are genuinely valid UTF-8 and no upstream ASCII guard would catch it. The defect is entirely in comparing a transformed copy and indexing the original.

## Solution

Compare the prefix in place rather than through a transformation that can change its length: take the first four bytes if they exist, compare them ASCII-case-insensitively, and take the remainder if it exists. Every step then asks for a range instead of assuming one.

## Notes

Read from v0.7.1 (9367845). Same class as #258, different mechanism and different file.

userFRM:
## Problem

`ns_parse` uppercased the payload to test for a `MISC` prefix, then sliced the original at byte 4.

Uppercasing is not length-preserving: `U+0131` and `U+017F` each occupy two bytes and uppercase to one ASCII byte. So a payload whose first four characters uppercase to `MISC` while occupying more than four bytes passed the check with byte 4 inside a character, and slicing a `&str` there panics.

## Scope

The framed receive path does not reach it. `is_ns_text` admits only an ASCII digit or a literal `MISC` before dispatching here, so this hardens the parser for callers that hold it directly rather than closing a live crash.

## What this changes

The prefix is compared in place against an ASCII `MISC`, and the remainder is taken with a checked slice.

The behaviour change is confined to the panic. The old code also recognised non-ASCII spellings, but it removed four *bytes* and each of those characters occupies two — so such a payload either put byte 4 inside a character, or shifted the cut so the remainder began mid-prefix and failed to parse. None of them ever parsed.

Also drops an allocation per inbound NS message on that path.

## Tests

- `a_prefix_that_changes_length_when_uppercased_does_not_panic` — the hostile spellings plus the short and empty shapes. Fails by name against the previous parser.
- `a_genuine_misc_prefix_is_still_stripped` — the positive control, in any ASCII casing.
- `the_prefix_is_an_ascii_misc_and_nothing_else` and `stripping_leaves_the_defined_shapes_alone` — pin the boundary the prefix rule draws. Both agree with the previous behaviour, which is the point; the guard for the change is the panic test.

Closes #365.

## Test plan

- [x] Mutation: restoring the uppercased-copy parser fails `a_prefix_that_changes_length_when_uppercased_does_not_panic` by name.
- [x] Positive control: `a_genuine_misc_prefix_is_still_stripped`, in any ASCII casing.
- [x] `the_prefix_is_an_ascii_misc_and_nothing_else` and `stripping_leaves_the_defined_shapes_alone` pin the boundary the prefix rule draws; both agree with the previous behaviour, which is the point.

### DIFF
--- a/src/protocol/ns.rs
+++ b/src/protocol/ns.rs
@@ -53,12 +53,22 @@ pub fn ns_build(version: u32, msg_type: u32, fields: &[&str], prefix: &str) -> V
 /// Parse NS payload into (version, msg_type, remaining_fields).
 pub fn ns_parse(payload: &[u8]) -> Option<(u32, u32, Vec<String>)> {
     let text = std::str::from_utf8(payload).ok()?;
-    // Strip MISC prefix if present
-    let text = if text.to_uppercase().starts_with("MISC") {
-        &text[4..]
-    } else {
-        text
-    };
+    // Strip MISC prefix if present.
+    //
+    // Compared in place rather than through `to_uppercase`, because
+    // uppercasing is not length-preserving: U+0131 and U+017F both uppercase
+    // to a single ASCII byte, so a payload beginning "m\u{131}\u{17F}c" passed the
+    // check while byte 4 sat inside a character — and slicing a `&str` there
+    // is a panic (same class as ibx#258).
+    //
+    // The framed receive path does not reach it: `is_ns_text` admits only an
+    // ASCII digit or a literal "MISC" before dispatching here. This is the
+    // parser's own guard, for callers that hold it directly.
+    let text = text
+        .get(..4)
+        .filter(|p| p.eq_ignore_ascii_case("MISC"))
+        .and_then(|_| text.get(4..))
+        .unwrap_or(text);
     let parts: Vec<&str> = text.split(';').collect();
     if parts.len() < 2 {
         return None;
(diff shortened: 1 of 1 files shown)


userFRM:

PR #370, opened 2026-07-30
gateway: scan the inflated init content without putting it in front of the engine (ibx#317)

### LINKED ISSUE #317: ccp: the session-start execution replay is delivered twice and only fills are deduplicated, so every other callback fires twice

userFRM:
## Problem

The logon ACK arrives DEFLATE-compressed inside `8=FIXCOMP` envelopes, and the routing tags the connect path needs are in the plaintext. `Gateway::connect` decompressed every segment and appended the result to `init_data` so the tag scan below could see it.

`init_data` is also the buffer handed to the engine — and the engine is given the burst still compressed, so it decompresses the same segments itself.

The hot loop therefore read each message in the init burst twice from a single delivery: once out of the segment it inflated, once from the plaintext copy appended behind it. Every account record, resting order and bulletin in the burst was handled twice. The duplicate was locally manufactured, not something the gateway sent.

## What this changes

The inflation is taken into a copy. `init_scan_buffer` takes the burst by reference and returns the buffer to scan, so the buffer the engine is handed cannot be the one that grows — reintroducing this takes changing the signature, not adding a line.

Nothing else changes: the tag scan sees exactly what it saw, and the engine is handed exactly what arrived.

This is deliberately not a deduplication layer. Deduplicating in the hot loop would suppress messages that are legitimately byte-identical — a `35=G` binary bar carries no sequence number or sending time, so an unchanged bar resent deliberately is indistinguishable from a repeat — and it would leave the duplicate being generated.

## Tests

`the_inflated_init_content_is_scanned_but_not_handed_to_the_engine` — the scan buffer gains exactly one inflated copy of a segment's content while the buffer the engine receives is unchanged. It fails by name when the inflation is dropped, and when the scan buffer stops carrying what arrived.

Closes #317.

## Test plan

- [x] Mutation: dropping the inflation fails `the_inflated_init_content_is_scanned_but_not_handed_to_the_engine` by name.
- [x] Mutation: having the scan buffer stop carrying what arrived fails the same test.
- [x] The `&[u8]` signature makes the original defect a compile error to reintroduce, not a review item: taking the burst by reference means the engine's copy cannot be the one that grows.

### DIFF
--- a/src/gateway.rs
+++ b/src/gateway.rs
@@ -914,6 +914,48 @@ pub struct GatewayConfig {
     pub code_provider: Option<session::CodeProvider>,
 }
 
+/// The buffer the init burst is scanned in.
+///
+/// The auth-server's logon ACK arrives DEFLATE-compressed inside one or more
+/// `8=FIXCOMP` envelopes (per ib-agent#129). The compressed body is ~30 kB on
+/// the wire but expands to ~48 kB of plaintext carrying the routing tags
+/// 6145/6171/8008, which the tag scan needs to see.
+///
+/// The inflated plaintext belongs to that scan and nowhere else. The engine is
+/// handed the burst exactly as it arrived and decompresses the same segments
+/// itself, so appending the plaintext to the engine's copy delivered every
+/// message in the burst twice from a single delivery — once from the segment,
+/// once from the appended copy (ibx#317). Takes the burst by reference so the
+/// engine's copy cannot be the one that grows.
+fn init_scan_buffer(init_data: &[u8]) -> Vec<u8> {
+    let mut scan = init_data.to_vec();
+    let mut cursor = 0usize;
+    while cursor + 12 < init_data.len() {
+        if init_data[cursor..].starts_with(b"8=FIXCOMP\x01") {
+            if let Some(total_len) = fixcomp::fixcomp_length(&init_data[cursor..]) {
+                let segment = &init_data[cursor..cursor + total_len.min(init_data.len() - cursor)];
+                let inflated = fixcomp::fixcomp_decompress(segment).unwrap_or_else(|e| {
+                    log::warn!("Init FIXCOMP segment at offset {}: dropping malformed frame: {}", cursor, e);
+                    Vec::new()
+                });
+                let inflated_bytes: usize = inflated.iter().map(|m| m.len() + 1).sum();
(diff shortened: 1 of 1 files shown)


userFRM:

Issue #372, opened 2026-07-30
orders: a modify cannot move a stop — the trigger is never read and the replace restates the old one


userFRM:
## Problem

`OrderRequest::Modify` carries a single price, and the API fills it from `lmt_price`. A stop order's trigger lives in `aux_price`, which is never read, so a modify of a stop sends the limit price it does not have and leaves the trigger unchanged.

`src/api/client/orders.rs:33`:

```rust
let price = (order.lmt_price * PRICE_SCALE_F) as i64;
let qty = order.total_quantity as u32;
ControlCommand::Order(OrderRequest::Modify { new_order_id: oid, order_id: oid, price, qty })
```

`lmt_price` is `0.0` on a plain `STP`, so the replace goes out with `44=0`. The encoder then restates the trigger from the tracked record (`order_builder.rs`, tag 99 from `o.stop_price`), which is the *old* value — the one the caller was trying to change.

The Python path builds the same request the same way.

## Impact

Moving a stop does nothing to the stop, and sends a zero limit price alongside it. The order remains on the book at its original trigger while the caller believes it has moved. On a protective stop that is the difference between the intended exit and none.

`STP`, `MIT` and `STP PRT` are all reachable this way — each is defined by `aux_price` and each is accepted for modification.

`STP LMT` is affected in the other direction: the limit moves, the trigger does not.

## Solution

Give `Modify` the trigger alongside the price, filled from `aux_price`, and have the replace encoder emit tag 99 from the request rather than from the tracked record. The tracked record is the right fallback when the caller passes no trigger, not the authority when it passes one.

## Notes

Read from v0.7.1 (9367845). Verifiable offline: `Modify` has no trigger field in `src/types.rs`, `aux_price` appears nowhere in the modify branch of `src/api/client/orders.rs`, and the replace reads tag 99 from `orig.stop_price`.

userFRM:
Closing this to clear the tracker. #409 carries the fork's main if the detail here is still useful.

deepentropy:
Same defect as #324, fixed on main in 8e2eb48: a stop modify moves the trigger, and a stop-limit modify moves both the limit and the trigger.

Issue #371, opened 2026-07-30
orders: every modify sends 6433=1, so a regular-hours order silently becomes eligible outside RTH


userFRM:
## Problem

The replace message sends `6433=1` unconditionally, so every modify enables extended-hours execution regardless of what the order was submitted with.

`src/engine/hot_loop/order_builder.rs:1494`:

```rust
(6433, "1"),         // OutsideRTH (preserve from original)
```

The comment states the intent — preserve the original — but the value is a literal. The submit path is conditional and only emits the tag when the caller asked for it (`order_builder.rs:1791`, `if attrs.outside_rth`), so a regular-hours order carries no 6433 on the way in and `6433=1` on every subsequent modify.

Nothing in the tracked `Order` record carries the flag either, so the encoder has nothing to preserve it from: `types::Order` holds side, price, quantity, order type, time-in-force and stop price.

## Impact

A caller that submits a regular-hours-only order and then moves its price has silently opted into extended-hours execution. The order can fill outside the session on the second message when it could not on the first, and nothing reports the change — `req_open_orders` and the order status both keep saying what the caller originally set.

This is the ordinary modify path, not an edge case. Any price or quantity change reaches it.

## Solution

Carry `outside_rth` on the tracked order so the replace can restate what was submitted, and emit tag 6433 conditionally the way the submit path does. Until the flag is tracked, the encoder has no correct value to send — omitting the tag entirely is closer to "unchanged" than asserting `1`, but which of those the gateway applies on a `35=G` is worth establishing from a capture before choosing.

## Notes

Read from v0.7.1 (9367845). Verifiable offline: `grep -n '6433' src/engine/hot_loop/order_builder.rs` shows the literal in the replace block and the conditional in the submit block, and `types::Order` has no field for it.

userFRM:
Closing this to clear the tracker. #409 carries the fork's main if the detail here is still useful.

deepentropy:
Same defect as #247, fixed on main in 8e2eb48: outside-RTH is now sent on a replace only when the order has it.

PR #373, opened 2026-07-30
orders: refuse a modify that would strip the order's attributes (ibx#248)

### LINKED ISSUE #248: orders: a modify is rebuilt from the tracked record, so the replace drops OCA (583/6209), the parent link (6107) and every other extended attribute
## Summary

A modify is rebuilt from the tracked `types::Order` record, which carries only `{side, price, qty, ord_type, tif, stop_price}`. Every extended attribute the order was placed with is therefore absent from the `35=G` replace — including the parent link and the OCA group that make a bracket a bracket.

The caller supplies all of it on the modify call, and `client_core` has the placed order cached. Neither is consulted.

## What is dropped

The Modify arm's field list (`src/engine/hot_loop/order_builder.rs:1486-1514`, v0.7.1 / 9367845) is 35=G, SendingTime, 11, 41, 44, 1, 6122, 6433, 38, 54, 40, 55, 167, 6035, 59, 6008, 6088, 6211, 6238, plus 99 when the tracked record has a stop price. Nothing else.

Against what placement emits (`send_order_ex:1794-1860`, and the bracket paths at `:677-679` / `:705-707`), a replace loses:

- **583 / 6209** — OCA group and type
- **6107** — parent order id
- **432 / 126** — GTD expiry, in either form
- **111** display size, **6135** hidden, **6115** trigger method, **9813** discretionary amount, **6102** sweep-to-fill, **18=G** all-or-none, **168** good-after-time, **6136+** conditions

The deliberate-omission comment at `:1485` lists only the identity tags (6121, 6119, 231, 100, 15, 204). The dropped attributes are not in that list — the message was transcribed from a wire capture of a plain order that happened to carry none of them (`ad1bfe0`, "matches ib-agent#101 wire capture"), so their absence looks incidental rather than decided.

## What I cannot determine

Whether the gateway treats an omitted 583/6107 on a `35=G` as *unchanged* or as *cleared*. Nothing in the repo settles it, and I have not run the paper test that would.

If omitted means unchanged, this is latent. If omitted means cleared, then modifying a bracket child detaches it: the stop stops being OCA-linked to the take-profit, and a fill on one leg no longer cancels the other, which leaves a naked resting order against a position that is already closed. The check is cheap — modify one leg of a paper bracket, fill the other, see whether the sibling cancels.

Worth noting the surrounding design points at "resubmission": `place_order` with a tracked id *is* the modify entry point (`src/api/client/orders.rs:32-41`), mirroring TWS, where a modify restates the whole order. Under that reading omitted means defaulted.

## Nothing compensates

- `client_core` keeps the full `ApiOrder` including `oca_group`, `parent_id` and `outside_rth` (`track_order`, `src/client_core.rs:768-773`), but the modify path forwards only price and qty.
- Session-recovery hydration (`ccp.rs:677-723`) parses 6008/54/38/44/99/40/59/55 — it cannot restore these either.
- There is no open-order resend on reconnect, so nothing re-asserts them later.
- No test would catch a regression: the only `35=G` wire test is a plain SPY limit (`tests/ib_paper_compat/orders.rs:194-236`), and `protocol_vectors.rs` has no replace vector.

## Suggested shape

Per-tag patching would leave the rest broken, so the fix belongs at the root: carry `tif` + `OrderAttrs` on `ControlCommand::Modify` from the `Order` the caller passed, and re-emit them in the replace the same way placement does. The data is already in hand at the call site.

## Related: 6107 assumes the parent is at version 0

`send_order_ex:1821-1825` emits the parent link as `format!("{}.0", attrs.parent_id)` while the child's own ClOrdID is version-aware (`:1640`) and a parent replace bumps the parent to `{id}.{ver}` (`:1457-1469`). A child attached after its parent has been modified therefore points at a ClOrdID that no longer exists.

This is the same class as #179: the server's actual ClOrdID string is kept in `last_clord` (`ccp.rs:740`) and the cancel path consults it (`:1399-1403`) — the parent-link path does not. A parent recovered from a prior session with a legacy un-versioned id makes `.0` wrong with zero modifies involved.

## Environment

v0.7.1 (9367845), read from the encoder path.

userFRM:
## Problem

The replace is rebuilt from the tracked record, which holds side, price, quantity, order type, time-in-force and trigger — and nothing else. Every other attribute rides a tag the replace does not carry, so a modify states the order without it.

The bracket links are the costly pair. A replace that omits the parent link or the OCA group leaves a child resting alone: a fill on one leg no longer cancels the other, and the position is left with a naked order against it.

## What this changes

The modify is refused for an order defined by any attribute the replace cannot restate:

- OCA group, parent link
- good-till expiry, good-after time
- iceberg display size, minimum quantity
- discretionary amount, sweep-to-fill
- non-default trigger method

**Refused, not carried.** Whether the gateway reads an omitted `583` or `6107` as unchanged or as cleared is not established, and the difference between latent and detached is the whole risk.

**Both sides are checked.** The resting order is the one the replace has to restate; the incoming order is what the caller is asking it to become. A modify that *adds* a bracket link or an OCA group states an order that has neither, so examining only the record on the book let the attribute through on the very message that was supposed to carry it. The check and its wording now live in one place, so the two bindings cannot diverge on either.

**`MTL`, `BOX TOP` and `MKT PRT` stay restatable.** The replace renders the same byte they were submitted under, which is the whole test for membership.

## Not reached here

Two defects on the ordinary modify path, neither a matter of dropped attributes:

- the replace asserts `6433=1` unconditionally, so any modify makes a regular-hours order eligible outside them (#371);
- it carries no trigger, so a modify of a stop sends a zero limit price and restates the old trigger (#372).

## Tests

- `an_order_defined_by_more_than_its_type_is_not_modified` — every protected attribute, each submitting first so the refusal proves something.
- `an_attribute_added_by_the_modify_is_refused_too` — adding one on the modify, not just holding one.
- `every_restatable_type_still_modifies` — including the three that were modifiable before this gate.
- `a_limit_if_touched_is_not_modified` — asserts tracking rather than skipping when untracked.

Closes #248.

## Test plan

- [x] Mutation: dropping the incoming-order check fails `an_attribute_added_by_the_modify_is_refused_too` by name.
- [x] Mutation: removing the `min_qty` guard fails `an_order_defined_by_more_than_its_type_is_not_modified`.
- [x] Mutation: refusing `MTL`/`BOX TOP`/`MKT PRT` again fails `every_restatable_type_still_modifies` — the regression guard for the three that were modifiable before this gate.
- [x] `a_limit_if_touched_is_not_modified` asserts the order is tracked first, so the refusal proves something rather than passing on an untracked id.

### DIFF
--- a/src/api/client/orders.rs
+++ b/src/api/client/orders.rs
@@ -31,6 +31,17 @@ impl EClient {
 
         // If orderId is already tracked, this is a modification — emit Modify instead of Submit.
         let cmd = if self.core.is_order_tracked(oid) {
+            // A replace states the order type, the limit price and the trigger.
+            // An order defined by anything else cannot survive one, so refuse
+            // rather than send a message that destroys it.
+            if let Some(tracked) = self.core.tracked_order(oid) {
+                if let Some(why) = ClientCore::replace_cannot_restate(&tracked) {
+                    return Err(format!(
+                        "{why} cannot be modified: the replace does not carry the fields that \
+                         define it, and sending one would cancel the order"
+                    ));
+                }
+            }
             let price = (order.lmt_price * PRICE_SCALE_F) as i64;
             let qty = order.total_quantity as u32;
             ControlCommand::Order(OrderRequest::Modify {
--- a/src/api/client/orders.rs
+++ b/src/api/client/orders.rs
@@ -34,13 +34,8 @@ impl EClient {
             // A replace states the order type, the limit price and the trigger.
             // An order defined by anything else cannot survive one, so refuse
             // rather than send a message that destroys it.
-            if let Some(tracked) = self.core.tracked_order(oid) {
-                if let Some(why) = ClientCore::replace_cannot_restate(&tracked) {
-                    return Err(format!(
-                        "{why} cannot be modified: the replace does not carry the fields that \
-                         define it, and sending one would cancel the order"
-                    ));
-                }
+            if let Some(refusal) = self.core.modify_refusal(oid, order) {
+                return Err(refusal);
             }
             let price = (order.lmt_price * PRICE_SCALE_F) as i64;
             let qty = order.total_quantity as u32;
--- a/src/api/client/tests.rs
+++ b/src/api/client/tests.rs
@@ -76,6 +76,19 @@ fn an_order_defined_by_more_than_its_type_is_not_modified() {
         ("conditional", |o| o.conditions.push(
             crate::types::OrderCondition::Time { time: "20260311-09:30:00".into(), is_more: true },
         )),
+        // Every attribute below rides a tag the replace does not carry, so a
+        // modify would state the order without it (ibx#248). The bracket links
+        // are the costly pair: a child sent without its parent or OCA group
+        // rests alone, and a fill on the sibling no longer cancels it.
+        ("bracket child", |o| o.parent_id = 4242),
+        ("OCA member", |o| o.oca_group = "bracket_1".into()),
+        ("good-till expiry", |o| o.good_till_date = "20260311 16:00:00".into()),
+        ("good-after time", |o| o.good_after_time = "20260311 09:30:00".into()),
+        ("iceberg", |o| o.display_size = 100),
+        ("minimum quantity", |o| o.min_qty = 50),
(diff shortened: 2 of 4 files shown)


userFRM:

PR #374, opened 2026-07-30
orders: route adaptive, algo and what-if through the shared encoder (ibx#318)

### LINKED ISSUE #318: orders: adaptive, algo, what-if and adjustable-stop orders silently drop outsideRth, parentId and the OCA group

userFRM:
> Stacked on #246 — review the tip commit only.

## Problem

`build_order_request` computes the extended-attribute block partway down, and three order types returned before reaching it. Their request variants carried no attribute block at all, so `Order::attrs()` was never consulted on those paths and their encoders emitted none of it.

Setting `outsideRth` on an adaptive, algo or what-if order was accepted by the API and silently ignored — the order went out regular-hours-only with nothing reporting the difference. The same early return bypassed the parent link and the OCA group, so an adaptive or algo order used as a bracket child was submitted unlinked and outside its group, which is what the attribute block exists to prevent. The time-in-force was hard-coded to `DAY` on all three.

## What this changes

Patching the missing tags into each dedicated encoder would leave the rest of the block — hidden, display size, minimum quantity, good-after, GTD expiry, discretionary amount, sweep-to-fill, all-or-none, trigger method, cash quantity, conditions — still dropped on those paths, and would put the emission in four places.

These route instead, the way the adjustable stop does in #246: `Adaptive`, `Algo` and `WhatIf` are `OrderKind` variants carried by `SubmitEx`, and the three standalone encoder arms are gone.

Their own tags keep their values and their position — after tag 204 and the attribute block, where the encoders this replaces put them: `18=e` and the adaptive priority parameter, the algo strategy with 849 and its parameter pairs, and the what-if flag. A what-if is still tracked under its marker so the response is recognised as a preview.

## Breaking change to the low-level Rust API

Pre-1.0, and the same shape #246 already applied to the adjustable stop:

- `OrderRequest::SubmitAdaptive`, `SubmitAlgo` and `SubmitWhatIf` are removed — use `SubmitEx { kind: OrderKind::Adaptive | Algo | WhatIf, tif, attrs }`.
- `Context::submit_adaptive`, `submit_algo` and `submit_what_if` take a time-in-force and an `OrderAttrs`, as `submit_adjustable_stop` does.
- `OrderKind` is no longer `Copy`, because the algo parameters it now carries own their strings.

The `Client`/`EClient` and Python surfaces are unchanged.

## Tests

- `the_algo_order_types_carry_the_attributes_the_caller_set` — at the API boundary, where the drop happened: all three types, outside-RTH, parent link, OCA group and a non-DAY tif.
- `adaptive_wire_carries_the_attributes_and_keeps_its_algo_tags`
- `algo_wire_carries_the_attributes_and_keeps_its_algo_tags`
- `what_if_wire_carries_the_attributes_and_keeps_its_preview_flag`

The three wire tests assert on the bytes, because the enum-level tests passed throughout the period the children shipped naked. Each fails by name against a compiling reversion of the production change it covers.

Closes #318.

## Test plan

- [x] Mutation: dropping the caller's attributes and tif on each of the three routing arms fails `the_algo_order_types_carry_the_attributes_the_caller_set` by name.
- [x] Mutation: dropping the what-if flag fails `what_if_wire_carries_the_attributes_and_keeps_its_preview_flag`.
- [x] Mutation: dropping `18=e` fails `adaptive_wire_carries_the_attributes_and_keeps_its_algo_tags`.
- [x] The three wire tests assert on the bytes, because the enum-level tests passed throughout the period the children shipped naked.
- [x] Wire output verified byte-identical to the standalone encoders — including field order, body length and checksum — for adaptive, what-if and all six algo strategies with default attributes.


userFRM:

Issue #224, opened 2026-07-15
orders: OrderAttrs and tif dropped on all types except LMT and TRAIL-pct — bracket children go live unlinked


deepentropy:
## Summary

`OrderAttrs` (`parent_id`, `oca_group`, `outside_rth`, …) and a non-DAY `tif` are carried by only **two** order types. Every other type maps to a plain `OrderRequest` variant that has no `attrs` and no `tif` field, so those values are accepted at the API boundary and silently dropped before the wire.

The user-visible result: **bracket children are transmitted as unlinked DAY orders**. A child stop-loss (`STP`, `tif="GTC"`, `parent_id=<entry>`) does not wait for its parent to fill — it goes live immediately as a standalone DAY stop. This is a live-money bug reachable from the documented `place_order` API with no error, no warning, and no log line.

## Affected types

`build_order_request` (src/client_core.rs:1298-1392) gates on `has_extended_attrs() || tif != "DAY"` in exactly two arms:

- `"LMT"` → `SubmitLimitEx` (:1302) — OK
- `"TRAIL"` with `trailing_percent > 0` → `SubmitTrailingStopPctEx` (:1324) — OK

Everything else builds a plain variant and drops attrs + tif:

| type | site | dropped |
|---|---|---|
| `STP` | :1312 | `parent_id`, `oca_group`, `tif`, `outside_rth` |
| `STP LMT` | :1316 | same |
| `TRAIL` by amount | :1333 | same |
| `MIT` | :1356 | same |
| `LIT` | :1360 | same |
| `MIDPRICE` / `MIDPX` | :1383 | same |
| `MKT` | :1299 | same |
| `MOC`, `LOC`, `TRAIL LIMIT`, `REL`, `PEG *`, `MTL`, `MKT PRT`, `STP PRT`, `SNAP *`, `BOX TOP` | :1338-1390 | same |

Two consequences beyond brackets:

- **`outside_rth` is lost on `MKT` and `MIDPRICE`.** A close-position order placed pre/post-market is accepted and then does not execute.
- **`oca_group` is lost on every type except `LMT`/`TRAIL`-pct**, so a take-profit and a stop-loss do not cancel each other on fill.

## Repro

```python
from ibx import EClient, EWrapper, Contract, Order

# ... connect, paper ...
parent_id = client.next_order_id()
child_id  = client.next_order_id()

parent = Order()
parent.action = "BUY"; parent.order_type = "STP"; parent.aux_price = 250.0
[...]

PR #376, opened 2026-07-30
orders: encode every order type through one path (ibx#375)

### LINKED ISSUE #375: orders: six order types send an optional tag in a different wire position depending on which encoder builds them
## Problem

Six order types put an optional tag in a different position depending on which encoder builds the message. The per-type encoders append it after tag 204; the shared extended-attribute encoder (`send_order_ex`) emits it in among the order-type tags, before 59.

Same tags, same values, different order:

```
MIDPX   own: … 40=MIDPX|59=0|167=STK|100=ISLAND|6210=ISLAND|15=USD|204=0|44=100
      shared: … 40=MIDPX|44=100|59=0|167=STK|100=ISLAND|6210=ISLAND|15=USD|204=0

TRAIL   own: … 40=P|99=0.5|211=0.5|18=a|59=0|…|204=0|6117=90
      shared: … 40=P|99=0.5|211=0.5|18=a|6117=90|59=0|…|204=0
```

Affected, and the tag that moves:

| type | tag |
|---|---|
| `MIDPX` | 44, the price cap |
| `PEG MKT` | 211, the offset |
| `PEG MID` | 211 and the two mid-offset tags 8403/8404 |
| `TRAIL`, `TRAIL LIMIT`, `TRAIL PCT` | 6117, the initial stop trigger |

Which path an order takes depends only on whether the caller set an extended attribute or a non-DAY time-in-force, so the same order type goes out with two different field orders depending on something unrelated to its type.

The per-type encoders are the ones carrying capture citations for these tags — `ib-agent#173` for 6117, and the position after 204 is what the other per-type arms use. So the shared encoder is the one that differs from what was observed.

## Impact

Any order of those six types carrying an extended attribute — a bracket link, an OCA group, outside-RTH, a GTD expiry, or simply a non-DAY tif — is sent with a field order the capture does not show. The other fifteen order types are byte-identical between the two paths, so this is not a general "field order does not matter" situation; it is six arms that were written differently from their siblings.

Whether the gateway is order-sensitive for these particular tags is not established. That is the reason to make them agree rather than to assume it does not matter.

## Solution

Emit the optional tags after 204 in `send_order_ex`, where the per-type encoders put them, so both paths produce the same message. Cover it with a test that encodes each type both ways and compares the frames, so the two cannot separate again.

## Notes

Read from v0.7.1 (9367845). Reproducible offline by encoding each type through both paths and comparing the frames with the sequence number and timestamps removed.

userFRM:
> Stacked on #374 — review the tip commit only.

## Problem

`build_order_request` chose per order type between a request that carries the extended attributes and a plain one that does not, and the engine had a separate encoder for each. Twenty-one order types were encoded twice, in two places, from two field lists.

That duplication is where the attribute bugs come from. An order type whose own encoder drifts from the shared one ships without something the caller set: bracket children submitted unlinked and DAY (#224), then the adjustable stop the same way (#240), then adaptive, algo and what-if (#318). Each was the same defect found again in a different arm.

## What this changes

Every type routes through the shared encoder. The plain-versus-extended choice is gone, so there is no arm for a type to drift into.

A test encodes each type both ways and compares the frames, which is what makes that safe to do and what stops the two paths separating again while both exist.

**It found six that did not agree** — all the same shape: an optional tag the per-type encoder appends after 204, emitted in among the order-type tags by the shared one.

```
MIDPX   own: … 40=MIDPX|59=0|167=STK|100=ISLAND|6210=ISLAND|15=USD|204=0|44=100
      shared: … 40=MIDPX|44=100|59=0|…|204=0
```

| type | tag |
|---|---|
| `MIDPX` | 44, the price cap |
| `PEG MKT` | 211, the offset |
| `PEG MID` | 211 and the mid-offset pair 8403/8404 |
| `TRAIL`, `TRAIL LIMIT`, `TRAIL PCT` | 6117, the initial stop trigger |

Those now sit where the per-type encoders put them, which is where the captures show them. The values and the conditions are unchanged; only the position moves, and only on the shared path. The other fifteen types were already byte-identical.

The remaining per-type request variants are unreachable from the API surface after this and are removed separately.

## Wire impact

None for an order with no extended attributes: the two encodings are byte-identical, tag for tag and value for value, for all twenty-one types.

For an order of those six types that does carry an extended attribute or a non-DAY time-in-force, the affected tag moves to the position the per-type encoder used.

## Tests

`the_shared_encoder_restates_every_type_exactly_as_its_own_variant_does` encodes all twenty-one types through both paths and compares the frames with the sequence number, timestamps, body length and checksum removed. It fails by name when a moved tag is put back, and when a tag is added to only one of the two encoders.

Closes #375.

## Test plan

- [x] `the_shared_encoder_restates_every_type_exactly_as_its_own_variant_does` encodes all twenty-one types through both paths and compares the frames with sequence number, timestamps, body length and checksum removed.
- [x] Mutation: putting the MIDPX price cap back inside the order-type block fails that test by name.
- [x] Mutation: dropping tag 6117 entirely fails it for all three trailing types.
- [x] Mutation: adding a tag to only the shared Market encoder also fails it — so the test catches unilateral drift, which is the reason it exists.


userFRM:

PR #377, opened 2026-07-30
orders: remove the per-type order request variants

### LINKED ISSUE #375: orders: six order types send an optional tag in a different wire position depending on which encoder builds them

userFRM:
> Stacked on #376 — review the tip commit only. Pure removal; no behaviour change.

## What this removes

With every order type encoding through the shared path (#376), the per-type request variants and their encoders are unreachable from the API surface and carry no behaviour of their own.

```
OrderRequest variants         42 → 15
conn.send_fix sites in one file  42 → 17
order_builder.rs            2059 → 1443 lines
```

Twenty-three variants and their encoder arms are gone.

## What does not change

The `Context` helpers that constructed them build the equivalent `SubmitEx` instead, with the same defaults they already implied. `submit_market`, `submit_limit`, `submit_stop` and the rest **keep their signatures**, so a caller of the low-level engine API sees no change at all.

`OrderRequest::order_id` already answers what the integration harness was answering with a forty-line match over every submit variant; it calls the accessor now.

## The equivalence test goes too

Its subject was the second encoding of each type, and it existed to establish that the two agreed before one of them was removed. What it guarded against — one encoder gaining a tag the other does not — is no longer expressible with a single encoder.

## `ib_paper_compat`

Three of the diagnostics that target already produced were call sites constructing a trailing-stop variant without its optional trigger. Migrating them supplied it, so the target now reports **nine rather than twelve**. The nine are the same `filters` omissions it reported before, unchanged.

## Verification

The removed encoders were byte-identical to the shared one for every type they covered, established by the equivalence test in #376 across all twenty-one types.

`--lib`, `--lib --features python`, `--bins`, `--examples` and all eight integration targets check clean.

Follows #376, which closes #375.

## Test plan

- [x] Behaviour preservation rests on the equivalence test in #376, which established the removed encoders were byte-identical to the shared one for every type they covered.
- [x] `Context` helpers keep their signatures, so a caller of the low-level engine API sees no change.
- [x] `--bins` included in the check sweep: two bench targets still constructed a removed variant and were migrated.
- [x] `ib_paper_compat` reports nine diagnostics rather than twelve — the three that disappeared were trailing-stop initializers missing their optional trigger, supplied during the migration; the remaining nine are unchanged.


userFRM:

PR #379, opened 2026-07-30
ccp: a cancel rejected as unknown retires the order (ibx#252)

### LINKED ISSUE #252: cancel-reject: an UnknownOrder rejection force-restores the order to Submitted in the engine while removing it from the cache

userFRM:
> Stacked on #322 — review the tip commit only.

## Problem

FIX `CxlRejReason 1` is UnknownOrder: the gateway is stating that the order does not exist on its side.

`handle_cancel_reject` restored it to working anyway. The restore ran first and unconditionally, and only afterwards did anything branch on the reason — then it removed the shared-cache row. Engine and cache were left asserting opposite things.

The engine's own view governs subsequent cancels, modifies and reconnect bookkeeping, so a phantom order persisted there with no cache row to make it visible — the inverse of the usual stale-cache problem and harder to notice. It compounds after a reconnect, where a cancel aimed at an order that no longer exists is the expected shape.

## What this changes

**The reason is read before the status is chosen**, and read as a positive statement rather than an absence. A missing or unparseable tag 102 is synthesized as `-1` and says nothing, so it takes the same path as the reasons that do mean the order is still working and the cancel arrived at the wrong moment.

**An unknown order is retired** — terminal status, removed from the book, cache row dropped.

Holding the record in a non-working status instead is not an option: those are excluded from the open-order count that guards instrument reclamation, so the slot could be handed to another contract while a retained order still pointed at it, and a late fill would move the wrong position.

**A fill that races the rejection is recoverable**, on the terms the untracked-fill path from #322 sets — hence the stacking. The execution has to carry its contract id, because nothing else says which instrument moved, and it must not be resend-marked, because a replayed execution for an order this session does not track is history rather than news. One carrying neither is dropped, as it already was for any order removed from the book ahead of a late fill.

**No synthetic status update is queued.** The cancel-reject is the report, and both dispatchers drain fills ahead of order updates; an update queued here would reach a caller *after* the fill that raced it, stating the order was gone when they had just been told it filled.

**Both dispatchers retire the client's own record** when they deliver that rejection. `collect_open_orders` unions the client's records with the cached view, so retiring only the engine's left the order being reported by the very surface this issue is about.

## Tests

- `an_unknown_order_rejection_retires_the_order` — order and cache row gone, cancel-reject delivered, no synthetic update.
- `an_execution_racing_an_unknown_order_rejection_still_books` — the fill books and the position moves.
- `any_other_rejection_leaves_the_order_in_place` — stated reasons that mean it is working, and the absent/unparseable case.

Closes #252.

## Test plan

- [x] Mutation: treating UnknownOrder like every other reason fails `an_unknown_order_rejection_retires_the_order` by name.
- [x] Mutation: leaving the order in the book fails the same test.
- [x] Mutation: retiring on any non-zero reason fails `any_other_rejection_leaves_the_order_in_place`.
- [x] `an_execution_racing_an_unknown_order_rejection_still_books` covers the case retiring the record could have cost.

### DIFF
--- a/src/api/client/dispatch.rs
+++ b/src/api/client/dispatch.rs
@@ -119,6 +119,13 @@ impl EClient {
         for reject in self.shared.orders.drain_cancel_rejects() {
             let code = if reject.reject_type == 1 { 202 } else { 10147 };
             let msg = format!("Order {} cancel/modify rejected (reason: {})", reject.order_id, reject.reason_code);
+            // Reason 1 is UnknownOrder: the gateway has said the order does not
+            // exist, and the engine has already retired its record. The client's
+            // own record has to go with it, or the open-order snapshot keeps
+            // reporting the order the rejection was about (ibx#252).
+            if reject.reason_code == 1 {
+                self.core.untrack_order(reject.order_id);
+            }
             wrapper.error(reject.order_id as i64, code, &msg, "");
         }
 
--- a/src/client_core.rs
+++ b/src/client_core.rs
@@ -784,6 +784,16 @@ impl ClientCore {
         }
     }
 
+    /// Stop tracking an order the gateway has said does not exist.
+    ///
+    /// A cancel rejected as UnknownOrder retires the engine's record, and the
+    /// client's own record has to go with it — the open-order snapshot unions
+    /// the two, so leaving this one behind kept reporting the order the
+    /// rejection was about (ibx#252).
+    pub fn untrack_order(&self, order_id: u64) {
+        self.open_orders.lock().unwrap().remove(&order_id);
+    }
+
     /// Update a tracked order status from an order update event.
     pub fn update_order_status(&self, order_id: u64, status: &str, filled: f64, remaining: f64) {
         let mut orders = self.open_orders.lock().unwrap();
--- a/src/engine/context.rs
+++ b/src/engine/context.rs
@@ -894,6 +894,13 @@ impl Context {
         self.market.register(con_id)
     }
 
+    /// Register without panicking when the instrument table is full. Inbound
+    /// message handling must use this — a full table is a condition to report,
+    /// not one to abort the engine on.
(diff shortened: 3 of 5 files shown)


userFRM:

PR #380, opened 2026-07-30
orders: a modify states the type, time-in-force and trigger it carries (ibx#349, ibx#372)

### LINKED ISSUE #349: orders: a modify silently ignores a changed time-in-force or order type, and the two tests for that behaviour assert nothing about it
## Problem

`OrderRequest::Modify` carries a price and a quantity. It carries neither the time-in-force nor the order type, and the encoder re-asserts whatever the tracked original held.

So re-placing a tracked order id with a different TIF, or a different order type, is accepted and acknowledged — and neither change reaches the gateway.

## Impact

A caller converting a DAY order to GTC through the documented modify path gets an order that still expires at the close. A caller changing a limit to a stop gets a limit. In both cases the modify is acked normally and nothing reports that the change was dropped.

This is the same shape as #247, which covered the outside-RTH flag, and as #334 for the fields that define an order type — the remaining fields on the same message.

## The existing tests do not cover it

`modify_tif_day_to_gtc_via_resubmit` and `modify_order_type_lmt_to_stp` both assert only that a `Modify` request was emitted. Neither asserts the resulting TIF or order type, so both pass while the behaviour they are named for does not happen. They should assert the emitted wire values, and will fail until the request carries the fields.

## Solution

Carry the time-in-force and the order type on the modify request, and emit them, rather than restating the tracked original.

### LINKED ISSUE #372: orders: a modify cannot move a stop — the trigger is never read and the replace restates the old one

userFRM:
> Stacked on #373 — review the tip commit only.

## Problem

`OrderRequest::Modify` carried a price and a quantity. It carried neither the order type, the time-in-force nor the trigger, and the encoder re-asserted whatever the tracked original held.

So re-placing a tracked order id with a different time-in-force, a different order type, or a moved stop was accepted and acknowledged, and none of it reached the gateway:

- a caller converting a DAY order to GTC got an order that still expired at the close;
- a caller moving a protective stop got the original trigger, and a zero limit price alongside it — the price the request does carry is read from `lmt_price`, which a stop does not have.

## What this changes

The request carries the three fields and the encoder emits them. **A zero states nothing** and leaves the resting order's value in force, which is every caller that only moves a price or a quantity — `Context::modify` keeps its signature and passes zeroes; `modify_ex` states the rest.

`Order::ord_type_byte` maps the order-type string to the byte the order tracks under, and only the types a replace can express are mapped. Everything else states nothing, which is safe because the gate in #373 refuses those modifies before they reach the encoder — otherwise a type the replace cannot express would be restated as the resting one while the client cached the new one.

**Both price-like fields are snapped to the instrument's grid.** `Modify` carries no instrument so `snap_prices` cannot reach it, and only the limit was being snapped here; a moved trigger off the grid is rejected by the gateway the same way a limit is.

## Tests

The two tests named for this behaviour asserted only that a `Modify` was emitted, so both passed for as long as the change was dropped. They now assert the value that was being dropped.

- `modify_tif_day_to_gtc_via_resubmit`, `modify_order_type_lmt_to_stp` — the request carries the change.
- `a_modify_states_the_type_tif_and_trigger_it_carries` — tags 40, 59 and 99 on the wire.
- `a_modify_that_states_nothing_keeps_the_resting_values` — the other direction.
- `a_moved_trigger_is_snapped_to_the_tick_grid` — `149.03` on a five-cent grid goes out as `149.05`.
- `a_modify_to_an_unrepresentable_type_is_refused` — `REL`, `TRAIL`, `LIT`, `MIDPX`, `SNAP MKT`.

Closes #349. Closes #372.

## Test plan

- [x] Mutation: restating the tracked type, tif or trigger instead of the request's fails `a_modify_states_the_type_tif_and_trigger_it_carries` by name — one mutation per field.
- [x] Mutation: dropping the type at the API boundary fails `modify_order_type_lmt_to_stp`.
- [x] Mutation: leaving the trigger unsnapped fails `a_moved_trigger_is_snapped_to_the_tick_grid`.
- [x] `a_modify_that_states_nothing_keeps_the_resting_values` covers the other direction, which is every caller that only moves a price or a quantity.
- [x] `a_modify_to_an_unrepresentable_type_is_refused` covers `REL`, `TRAIL`, `LIT`, `MIDPX` and `SNAP MKT`, so the zero-means-unstated convention cannot silently restate a type.

### DIFF
--- a/src/api/client/orders.rs
+++ b/src/api/client/orders.rs
@@ -44,6 +44,9 @@ impl EClient {
                 order_id: oid,
                 price,
                 qty,
+                ord_type: order.ord_type_byte(),
+                tif: order.tif_byte(),
+                stop_price: (order.aux_price * PRICE_SCALE_F) as i64,
             })
         } else {
             ClientCore::build_order_request(order, oid, instrument)?
--- a/src/api/client/orders.rs
+++ b/src/api/client/orders.rs
@@ -31,6 +31,17 @@ impl EClient {
 
         // If orderId is already tracked, this is a modification — emit Modify instead of Submit.
         let cmd = if self.core.is_order_tracked(oid) {
+            // A replace states the order type, the limit price and the trigger.
+            // An order defined by anything else cannot survive one, so refuse
+            // rather than send a message that destroys it.
+            if let Some(tracked) = self.core.tracked_order(oid) {
+                if let Some(why) = ClientCore::replace_cannot_restate(&tracked) {
+                    return Err(format!(
+                        "{why} cannot be modified: the replace does not carry the fields that \
+                         define it, and sending one would cancel the order"
+                    ));
+                }
+            }
             let price = (order.lmt_price * PRICE_SCALE_F) as i64;
             let qty = order.total_quantity as u32;
             ControlCommand::Order(OrderRequest::Modify {
--- a/src/api/client/orders.rs
+++ b/src/api/client/orders.rs
@@ -34,13 +34,8 @@ impl EClient {
             // A replace states the order type, the limit price and the trigger.
             // An order defined by anything else cannot survive one, so refuse
             // rather than send a message that destroys it.
-            if let Some(tracked) = self.core.tracked_order(oid) {
-                if let Some(why) = ClientCore::replace_cannot_restate(&tracked) {
-                    return Err(format!(
-                        "{why} cannot be modified: the replace does not carry the fields that \
-                         define it, and sending one would cancel the order"
-                    ));
-                }
+            if let Some(refusal) = self.core.modify_refusal(oid, order) {
+                return Err(refusal);
             }
             let price = (order.lmt_price * PRICE_SCALE_F) as i64;
             let qty = order.total_quantity as u32;
--- a/src/api/client/tests.rs
+++ b/src/api/client/tests.rs
@@ -76,6 +76,19 @@ fn an_order_defined_by_more_than_its_type_is_not_modified() {
         ("conditional", |o| o.conditions.push(
             crate::types::OrderCondition::Time { time: "20260311-09:30:00".into(), is_more: true },
         )),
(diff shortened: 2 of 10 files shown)


userFRM:

PR #382, opened 2026-07-31
hmds: release the connection when the transport is given up on (ibx#367)

### LINKED ISSUE #367: hmds: a liveness timeout leaves the connection in place, so the HMDS reconnect never runs
## Problem

An HMDS liveness timeout marks the transport dead but leaves the connection in place, and the reconnect path refuses to run while a connection is present.

`poll_once` sets the flag on timeout (`src/engine/hot_loop/mod.rs:743`):

```rust
log::error!("HMDS liveness timeout ({}s silent) — connection lost", since_recv);
self.hmds.disconnected = true;
```

`self.hmds_conn` still holds the dead socket. `maybe_spawn_hmds_reconnect` opens with `if self.hmds_conn.is_some() { return; }` (`mod.rs:1006`), so it returns immediately and never schedules an attempt.

The other two transports do not have this shape: CCP gates its reconnect on `!self.ccp.disconnected` (`mod.rs:858`) and the farm on `!self.farm.disconnected` (`mod.rs:832`), so setting the flag is sufficient for both. Only HMDS gates on the connection instead of the flag.

## Impact

The first HMDS liveness timeout is permanent for the life of the process. The heartbeat block is also skipped from then on — it is guarded by `!self.hmds.disconnected && self.hmds_conn.is_some()` — so nothing further is sent on that socket and nothing re-establishes it. Historical requests issued afterwards target the abandoned connection and never complete. Recovery requires restarting the engine.

## Solution

Release the connection where the flag is set, so the existing reconnect gate sees what it is written to look for:

```rust
self.hmds.disconnected = true;
self.hmds_conn = None;
```

Alternatively, gate `maybe_spawn_hmds_reconnect` on `self.hmds.disconnected` the way the other two transports are gated. The first is the smaller change and matches how `hmds_conn` is treated on the reconnect success path (`mod.rs:1056`).

## Notes

Read from v0.7.1 (9367845). Verifiable offline: the flag is set at `mod.rs:743`, the connection is untouched, and `mod.rs:1006` returns on `is_some()`.

userFRM:
## Problem

Giving up on the HMDS transport set the disconnected flag and left the dead socket in place.

`maybe_spawn_hmds_reconnect` returns early while a connection is present, so it never scheduled an attempt — the transport was stuck for the life of the process. The heartbeat block is guarded on the same pair, so nothing further was sent on it either, and historical requests issued afterwards targeted a connection nothing was going to replace. Recovery needed restarting the engine.

Both routes there did it: the liveness timeout, and the ordinary receive error — a closed peer, or any other failure out of `try_recv`, which is how the transport usually dies.

## What this changes

The flag and the connection move together in one method that both callers use, so setting one without the other is not something a caller can express.

**A reconnect no longer inherits the dead session's outstanding probe.** The liveness check sends a TestRequest only when none is pending, so a stale one suppressed every probe on the new connection and it went from silent to declared dead without being asked anything. The warm-up window starts with the new session as well, rather than being counted from the old one and already expired.

CCP and the farm gate their reconnects on the flag alone and were unaffected.

## Tests

- `an_hmds_disconnect_lets_its_reconnect_run` — driven through the receive path with the peer gone: the transport is marked dead, the socket is released, and the scheduler arms an attempt instead of skipping forever.
- `an_hmds_reconnect_does_not_inherit_the_dead_session_probe` — drives `poll_hmds_reconnect` with a successful result and asserts the probe is cleared and the warm-up restarted.

Each fails by name against a compiling reversion of the production line it covers.

Closes #367.

## Test plan

- [x] Mutation: not releasing the connection fails `an_hmds_disconnect_lets_its_reconnect_run` by name — the test seeds a live socket first, or releasing it would be unobservable.
- [x] Mutation: keeping the dead session's pending probe fails `an_hmds_reconnect_does_not_inherit_the_dead_session_probe`.
- [x] The disconnect test drives the ordinary receive path with the peer gone, rather than calling the helper — that is how the transport usually dies.

### DIFF
--- a/src/engine/hot_loop/hmds.rs
+++ b/src/engine/hot_loop/hmds.rs
@@ -75,6 +75,18 @@ impl HmdsState {
         }
     }
 
+    /// Give up on the transport.
+    ///
+    /// The flag and the connection move together. Setting the flag alone left
+    /// the dead socket in place, and the reconnect scheduler returns early
+    /// while a connection is present — so the transport was stuck for the life
+    /// of the process, on both the liveness timeout and the ordinary
+    /// receive-error path (ibx#367).
+    pub(crate) fn disconnect(&mut self, hmds_conn: &mut Option<Connection>) {
+        self.disconnected = true;
+        *hmds_conn = None;
+    }
+
     pub(crate) fn poll(
         &mut self,
         hmds_conn: &mut Option<Connection>,
@@ -91,7 +103,7 @@ impl HmdsState {
                     Ok(0) => {}
                     Err(e) => {
                         log::error!("HMDS connection lost: {}", e);
-                        self.disconnected = true;
+                        self.disconnect(hmds_conn);
                         return;
                     }
                     Ok(n) => {
--- a/src/engine/hot_loop/mod.rs
+++ b/src/engine/hot_loop/mod.rs
@@ -740,7 +740,7 @@ impl HotLoop {
             if warmed_up && since_recv > LIVENESS_TEST_SECS {
                 if since_recv > LIVENESS_DEAD_SECS {
                     log::error!("HMDS liveness timeout ({}s silent) — connection lost", since_recv);
-                    self.hmds.disconnected = true;
+                    self.hmds.disconnect(&mut self.hmds_conn);
                 } else if self.hb.pending_hmds_test.is_none() {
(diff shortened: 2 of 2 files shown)


userFRM:

Issue #228, opened 2026-07-15
req_matching_symbols: FIFO matching cross-attributes responses, and an empty result permanently poisons the queue


deepentropy:
## Summary

`req_matching_symbols` matches responses to requests **FIFO** rather than by the tag-320 SecurityReqID the server echoes. Two concurrent calls get cross-attributed. And an empty result set is dropped without popping the queue, which **permanently** misattributes every subsequent response for the life of the process.

This is the same bug that was already found and fixed for `pending_secdef`. The fix comment is still there at `ccp.rs:461-466`; it was never applied here.

## The code

`src/engine/hot_loop/ccp.rs:317-326`:

```rust
"186" => {
    if let Some(matches) = crate::control::contracts::parse_matching_symbols_response(msg) {
        if !matches.is_empty() {
            if let Some(req_id) = self.pending_matching_symbols.first().copied() {
                self.pending_matching_symbols.remove(0);
                shared.reference.push_matching_symbols(req_id, matches);
            }
        }
    }
}
```

Two defects:

**1. `.first()` instead of tag 320.** The response carries the req_id the server echoed; it is ignored. Two in-flight `req_matching_symbols` calls that complete out of order deliver request #2's results under request #1's req_id.

**2. `if !matches.is_empty()` never pops.** An unknown pattern returns zero matches. Then:
- `symbol_samples` never fires — no callback, no error, no end. The caller waits forever.
- `pending_matching_symbols.remove(0)` never runs, so the stale entry stays at the head **permanently**.

After (2), the queue is poisoned: every later response pops the *stale* head, so every result from then on is delivered under the wrong req_id. One typo'd ticker breaks symbol search for the rest of the session, silently and unrecoverably.

## This was already fixed once, next door

`ccp.rs:461-466` — the secdef path explains exactly why it matches on tag 320:

> ```rust
> // Match the reply to a pending request by the req_id the server echoed
> // in tag 320, NOT by queue order: replies can arrive out of order and
> // FIFO matching cross-attributes them. (ibx#...)
> ```

`pending_secdef` got the treatment. `pending_matching_symbols`, ~150 lines away in the same file and the same `match`, did not.

## Repro

```python
# Poison, permanently:
c.req_matching_symbols(1, "ZZZZQQ")     # no matches -> symbol_samples never fires,
                                        #               queue head never pops
c.req_matching_symbols(2, "AAPL")       # fires symbol_samples(1, [AAPL...])  <- wrong req_id
[...]

PR #383, opened 2026-07-31
ccp: a matching-symbols request is recorded only when it was sent (ibx#369)

### LINKED ISSUE #369: ccp: a matching-symbols request is recorded as pending even when it was never sent, and nothing times it out
## Problem

`send_matching_symbols_request` records the request as pending whether or not it was sent, and nothing ever times it out.

`src/engine/hot_loop/ccp.rs:1586`:

```rust
if let Some(conn) = ccp_conn.as_mut() {
    ...
    let _ = conn.send_fix(&[ ... ]);
    hb.last_ccp_sent = Instant::now();
    log::info!("Sent matching symbols request: req_id={} pattern='{}'", req_id, pattern);
}
self.pending_matching_symbols.push(req_id);
```

Two ways the entry outlives the request. The send error is discarded, so a failed write still records a pending id. And the push is outside the `if let`, so a request issued while the CCP transport is down is recorded without any send being attempted at all — the log line that would say otherwise is inside the block.

`pending_matching_symbols` is only ever drained by a matching response (`ccp.rs:362-379`). There is no deadline on it, unlike `pending_historical`, which carries `Instant::now() + HISTORICAL_IDLE_TIMEOUT`. Reconnect neither resends nor rejects the entry.

## Impact

The caller waits forever: no response, no timeout, no error callback. The stale id also stays in the list, and the response matcher falls back to position 0 when a reply carries no echoed request id (`ccp.rs:365`), so a later request's response can be delivered against the abandoned entry.

## Solution

Record the request only where it was actually sent, and give it a deadline the way historical requests have one, so a request that goes out and is never answered terminates rather than hanging. A request issued with no connection should be rejected to the caller at the point of issue.

## Notes

Read from v0.7.1 (9367845). Verifiable offline: the send result is discarded at `ccp.rs:1590`, the push is unconditional at `ccp.rs:1600`, and `grep -rn pending_matching_symbols src/` shows no timeout sweep.

userFRM:
## Problem

The request was pushed onto the pending queue whether or not it went out. The send error was discarded, and the push sat outside the block that needs a connection at all:

```rust
if let Some(conn) = ccp_conn.as_mut() {
    let _ = conn.send_fix(&[ ... ]);
    ...
}
self.pending_matching_symbols.push(req_id);
```

So a request issued while the CCP transport was down was queued as pending with nothing on the wire to answer it.

Nothing expired the queue either. An unanswered request stayed in it for the life of the process — and the reply matcher falls back to the head of that queue when a reply carries no echoed request id, so a stale entry could absorb a later request's answer. That is the failure #228 fixed for a different cause.

## What this changes

A request is recorded only where it was sent, and carries a deadline after which it is given up on — matching what historical requests already do, since both are one round trip to the gateway.

## What this does not do

A request that could not be sent is logged and not recorded. Surfacing that to the caller would need an error channel this request type does not have: the reply arrives through the symbol-samples callback and there is nothing beside it to carry a failure.

## Tests

- `a_matching_symbols_request_that_was_not_sent_is_not_recorded` — with no transport, nothing is awaited; with one, it is.
- `an_unanswered_matching_symbols_request_is_given_up_on` — the expired entry is dropped and the live one kept.

Each fails by name against a compiling reversion of the production line it covers.

Closes #369.

## Test plan

- [x] Mutation: recording the request with no transport fails `a_matching_symbols_request_that_was_not_sent_is_not_recorded` by name.
- [x] Mutation: never expiring an entry fails `an_unanswered_matching_symbols_request_is_given_up_on` — anchored on the intended sweep, since the file contains a second deadline check.
- [x] The existing reply-attribution tests were updated to the deadline-carrying queue and still pass.

### DIFF
--- a/src/engine/hot_loop/ccp.rs
+++ b/src/engine/hot_loop/ccp.rs
@@ -1,6 +1,10 @@
 use std::collections::{HashSet, VecDeque};
 use std::time::{Duration, Instant};
 
+/// How long a matching-symbols request waits for its reply. Matches the
+/// historical-request idle timeout: both are one round trip to the gateway.
+const MATCHING_SYMBOLS_TIMEOUT: Duration = Duration::from_secs(60);
+
 use crate::bridge::{Event, RichOrderInfo, SharedState};
 use crate::api::types as api;
 use crate::engine::context::Context;
@@ -77,7 +81,10 @@ pub(crate) struct CcpState {
     /// surfaces error 200 + contract_details_end instead of hanging
     /// forever (ibx#227).
     pub(crate) pending_secdef: Vec<(u32, bool, Instant)>,
-    pub(crate) pending_matching_symbols: Vec<u32>,
+    /// Requests awaiting a matching-symbols reply, with the deadline after
+    /// which one is given up on. Recorded only for a request that actually went
+    /// out, and expired so a stale head cannot absorb a later reply (ibx#369).
+    pub(crate) pending_matching_symbols: Vec<(u32, Instant)>,
     /// keepUpToDate historical queries routed through CCP: (query_id, req_id)
     pub(crate) pending_kut_historical: Vec<(String, u32)>,
     /// tickerId → req_id mapping for keepUpToDate 35=G bar updates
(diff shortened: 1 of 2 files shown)


userFRM:

Issue #388, opened 2026-07-31
bridge: SeqQuote's payload is accessed non-atomically, so a torn read is UB before the version check can reject it


userFRM:
## Problem

`SeqQuote`'s payload is read and written through plain, non-atomic access while a concurrent reader may be looking at it. That is a data race by the Rust abstract machine regardless of how the version counter is ordered, and Miri reports it as UB.

The seqlock's version counter tells a reader whether the snapshot it took is trustworthy, and the retry loop discards a torn one. What it cannot do is make the torn read itself defined: the reader has already performed a non-atomic read of memory a writer is concurrently writing, and that is UB before the version check ever runs.

Reproduced with Miri across ten interleaving seeds. It fires at the same rate whether the odd-version transition is annotated `Release` or `AcqRel`, which is what distinguishes it from the ordering defect in #267 — that one is about where the payload write may be reordered to, this one is about the access itself.

## Impact

No misbehaviour has been observed on x86, where a core does not reorder its own accesses relative to each other and the compiler has not chosen to exploit the UB. That is a statement about one target and one optimiser, not a guarantee. A compiler is entitled to assume the race cannot happen and optimise accordingly, and aarch64 has a weaker memory model.

It also means the type cannot be validated under Miri, so a genuine future defect in it would be lost among the reported races.

## Solution

Give the payload per-field atomic or volatile access so the read and the write are both defined, and keep the version counter as the freshness test it already is. `Quote` is a small struct of scalars, so per-field `AtomicI64`/`AtomicU32` with relaxed ordering is the usual shape — the seqlock's fences already supply the ordering, and relaxed atomics only need to make the access itself defined.

This is materially larger than the ordering fix in #267 and is deliberately separated from it.

## Notes

Read from v0.7.1 (9367845). Confirmed with Miri on the nightly toolchain; a plain 30-million-write stress across twelve threads in both debug and release showed no observable divergence on x86, which is why this needs the tool rather than a test.

userFRM:
Closing this to clear the tracker. #409 carries the fork's main if the detail here is still useful.

PR #389, opened 2026-07-31
bridge: fence SeqQuote's payload write inside the odd-version window (ibx#267)

### LINKED ISSUE #267: bridge: SeqQuote races non-atomically and its odd-version marker does not fence the payload write
## Problem

`SeqQuote` (`src/bridge.rs:85-119`) is a seqlock whose ordering does not establish what its `unsafe impl Sync` claims. Two separate issues in the same twelve lines.

**The data access is a data race by the model.** The writer does `*self.data.get() = *quote` and the reader does `*self.data.get()` — both plain non-atomic accesses to the same location from different threads. Miri reports it directly:

```
Undefined Behavior: Data race detected between (1) non-atomic read ...
and (2) non-atomic write ... at src/bridge.rs:115:30
```

The retry loop catches interleavings that actually happen on hardware, which is why nothing has been observed in the field, but the compiler is entitled to assume the race cannot occur.

**The odd marker does not fence the data write.**

```rust
self.version.store(v + 1, Ordering::Release); // odd = writing
unsafe { *self.data.get() = *quote; }
```

A `Release` store orders operations *before* it. Nothing here forbids the data write being reordered *above* the odd marker, and if it is, a reader can observe `v1 == v2` even while the payload is mid-update — the version check passes over a torn quote. The ordering needed is the opposite direction from the one used.

The `SAFETY` comment at `:87-88` — "The version counter ensures readers see consistent data" — states the guarantee that these two issues undermine.

## Impact

Correctness by construction rather than an observed failure: on x86 with current LLVM no tear has been seen, and I am not reporting one. It matters because this is the hottest read path in the engine, the payload is quote data every strategy reads, and the failure mode if it ever does reorder is a silently torn quote rather than a crash.

## Solution

Make the odd transition an RMW — `version.fetch_add(1, Ordering::AcqRel)` — so later accesses cannot move above it, and give the payload atomic or volatile per-field accesses so the reader and writer are no longer racing non-atomically. Then the `SAFETY` comment can state something the code actually provides.

## Notes

Read from v0.7.1 (9367845); the UB half is a miri verdict on this type, not a static reading.

userFRM:
## Problem

`SeqQuote` is a seqlock: the writer makes the version odd, writes the payload, then makes it even, and a reader retries while the version is odd or changed under it. The odd-version transition used a `Release` store.

`Release` orders what comes *before* the store. It places no constraint on what comes after, so nothing stopped the payload write being reordered above the store that was supposed to announce it — a reader could observe an even version and a payload already half-updated, with the retry loop none the wiser.

## What this changes

The odd transition is a `fetch_add` with `AcqRel`. The Acquire half is what pins the following payload write inside the odd window. The closing transition keeps `Release`, which was already correct.

## What the test can and cannot show

`seqquote_no_torn_reads` runs four readers against one writer over twenty thousand quotes, comparing *every* field against a sentinel rather than the two the existing test checked. It is a genuine regression guard on the retry contract and strictly stronger than what was there.

It does **not** fail when the ordering annotation alone is reverted, and no test on this target would. x86 does not reorder a core's own accesses relative to each other, so the defect is real by the Rust abstract machine — where the compiler is free to hoist the payload write — but not something an x86 execution can be expected to expose. Thirty million writes across twelve threads, in debug and release, showed no divergence between the two orderings.

This is stated rather than papered over: the guard here is the reasoning about the memory model, and the test protects the surrounding contract.

## Separately

Miri reports a data race in this type regardless of how the version is ordered: the payload itself is read and written non-atomically, so a torn read is undefined before the version check can reject it. That needs per-field atomic or volatile access and is materially larger than this — filed as #388.

Closes #267.

### DIFF
--- a/src/bridge.rs
+++ b/src/bridge.rs
@@ -100,10 +100,12 @@ impl SeqQuote {
     /// Write a quote (hot loop side). Never blocks.
     #[inline]
     pub fn write(&self, quote: &Quote) {
-        let v = self.version.load(Ordering::Relaxed);
-        self.version.store(v + 1, Ordering::Release); // odd = writing
+        // AcqRel, not Release: the payload write below must not be reordered
+        // above this store. Release alone only fences what precedes it; the
+        // Acquire half is what pins *following* accesses inside the odd window.
+        self.version.fetch_add(1, Ordering::AcqRel); // odd = writing
         unsafe { *self.data.get() = *quote; }
-        self.version.store(v + 2, Ordering::Release); // even = stable
+        self.version.fetch_add(1, Ordering::Release); // even = stable
     }
 
     /// Read a consistent quote snapshot (reader side). Spins on conflict.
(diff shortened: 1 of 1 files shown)


userFRM:

Issue #381, opened 2026-07-31
python: no Rust-side unit test under the python feature has ever run — the harness does not compile, then does not link


userFRM:
## Problem

No Rust-side unit test under the `python` feature has ever run. `cargo test --lib --features python` fails in two stages:

1. **Compile.** `src/python/compat/wrapper.rs:301` calls `EWrapper::new()` with no arguments. The PyO3 `#[new]` constructor takes `(&Bound<PyTuple>, Option<&Bound<PyDict>>)`, so the test has not compiled since that signature was introduced.

2. **Link.** With that test removed, the build reaches the linker and fails with `undefined symbol: PyObject_GetIter`. The crate builds the extension as a `cdylib` against the Python ABI, so a test binary has no libpython to resolve against.

`cargo check --lib --features python` passes, which is why this is invisible: the check gate compiles the library but never builds the test harness.

## Impact

Everything in `src/python/` is unverified by Rust tests — the compat `Contract`/`Order` conversions, the dispatch loop, the callback reconstruction. `tests/python/*.py` exercises the built extension end to end, but nothing covers the Rust side in isolation, and a unit test added there today would silently never run.

The gap has already cost something concrete: the open-order callback rebuilt `Order` from a hand-picked subset of fields for as long as it existed, and no Rust test could have caught it.

## Solution

Two parts, in order:

- Fix or remove the stale `ewrapper_can_be_constructed` test so the harness compiles. The assertion it makes — that the type is constructible — is what the compiler already checks.
- Make the test harness linkable. The usual approach is to keep `extension-module` off for test builds and let the harness link libpython, gating it behind a feature so the shipped wheel is unaffected.

Then add `cargo test --lib --features python` to whatever runs the check gate, so the harness cannot silently stop building again.

## Notes

Read from v0.7.1 (9367845). Reproducible offline with `cargo test --offline --lib --features python --no-run`.

userFRM:
Closing this to clear the tracker. #409 carries the fork's main if the detail here is still useful.

PR #390, opened 2026-07-31
python: let KeyboardInterrupt and SystemExit out of a callback (ibx#270)

### LINKED ISSUE #270: python: callback exceptions are swallowed wholesale, so Ctrl-C during a callback is lost and SystemExit cannot stop the loop
## Problem

`call_wrapper!` (`src/python/compat/client/dispatch.rs:22-31`) swallows every exception a callback raises, including the two that are not supposed to be caught:

```rust
if let Err(e) = $wrapper.call_method($py, $method, $args, None) {
    log::error!("Python callback {}() raised: {}", $method, e);
    e.restore($py);
    unsafe { pyo3::ffi::PyErr_Clear(); }
}
```

The `restore` followed immediately by `PyErr_Clear` nets out to dropping the exception; the pair does nothing the drop alone would not.

Catching ordinary exceptions so one bad callback cannot kill the dispatch loop is a reasonable decision. `KeyboardInterrupt` and `SystemExit` are a different matter — they derive from `BaseException` rather than `Exception` precisely so that blanket handlers do not eat them.

**Ctrl-C is lost while a callback is running.** CPython delivers the signal by raising `KeyboardInterrupt` in the executing frame. If that frame is user callback code invoked from here, the exception surfaces as the `call_method` error and is cleared. `run`'s `py.check_signals` (`mod.rs:217`) never sees it, because the signal has already been converted into the exception that was just discarded. Only a Ctrl-C landing in the wait window works — which on a busy feed is the smaller share of the time.

**A callback cannot stop the loop.** `raise SystemExit` inside `error()` or `order_status()` is logged and ignored, where ibapi propagates it and `run()` terminates. That is a divergence on the operator-control path.

## Solution

In the `Err` arm, re-raise when the exception is a `BaseException` that is not an `Exception` — `KeyboardInterrupt`, `SystemExit` — and keep logging and swallowing everything else. The `restore`/`PyErr_Clear` pair should go regardless: dropping the `PyErr` is sufficient and says what it means.

## Notes

Read from v0.7.1 (9367845), including the CPython signal semantics that make the Ctrl-C case follow. Static trace: pyo3 is built with `extension-module` here, so there is no embedded interpreter for a Rust test to raise inside.

userFRM:
## Problem

Every Python callback was invoked through a macro that caught whatever it raised, logged it, and carried on. That is right for an ordinary exception — one bad callback should not kill the dispatch loop — but it applied to `KeyboardInterrupt` and `SystemExit` too.

So Ctrl-C pressed while a callback was running was logged and discarded, and a callback raising `SystemExit` could not stop `run()`.

## What this changes

An exception that derives from `BaseException` but not `Exception` is re-raised instead of logged. That is exactly the set Python reserves for control flow — `KeyboardInterrupt`, `SystemExit`, `GeneratorExit` — and it is the same line ibapi draws. Ordinary exceptions are still caught and logged, unchanged.

Every call site sits directly in `dispatch_once`'s body, so the re-raise unwinds through its `PyResult<()>`, through `run()`'s `?`, and back to the Python caller as the original exception. Verified by walking the brace structure of the function rather than by grep: the function contains ten closure bodies, and no call site is inside one — a `return` in a closure would have exited the closure and swallowed the interrupt anyway.

The `restore`-then-`PyErr_Clear` pair on the swallow path is removed. It nets out to dropping the error, which is what letting it fall out of scope does.

## Verification

`cargo check --offline --lib --features python` is clean, and the full gate passes.

Not run: `cargo test --lib --features python`, which does not link in this tree — a stale test blocks compilation and libpython is then missing (#381). The propagation path is therefore established by reading the call-site placement, the return-type plumbing, and pyo3's exception hierarchy rather than by executing it.

Closes #270.

### DIFF
--- a/src/python/compat/client/dispatch.rs
+++ b/src/python/compat/client/dispatch.rs
@@ -18,14 +18,18 @@ use super::super::contract::{Contract, ContractDescription, ContractDetails, Bar
 use super::super::tick_types::*;
 use super::super::super::types::PRICE_SCALE_F;
 
-/// Call a Python wrapper method, catching and logging any exception instead of propagating.
-/// This prevents user callback exceptions from killing the dispatch loop.
+/// Call a Python wrapper method, catching and logging an ordinary exception instead of
+/// propagating it so one bad callback cannot kill the dispatch loop. `KeyboardInterrupt`,
+/// `SystemExit`, and any other exception deriving from `BaseException` rather than
+/// `Exception` are re-raised so Ctrl-C during a callback still stops `run()` and a
+/// callback-raised `SystemExit` still terminates it, matching ibapi.
 macro_rules! call_wrapper {
     ($wrapper:expr, $py:expr, $method:expr, $args:expr) => {
         if let Err(e) = $wrapper.call_method($py, $method, $args, None) {
+            if !e.is_instance_of::<pyo3::exceptions::PyException>($py) {
+                return Err(e);
+            }
             log::error!("Python callback {}() raised: {}", $method, e);
-            e.restore($py);
-            unsafe { pyo3::ffi::PyErr_Clear(); }
         }
     };
 }


userFRM:

PR #391, opened 2026-07-31
python: release the GIL around the operations that block (ibx#271)

### LINKED ISSUE #271: python: the GIL is held across thread joins, control-channel sends and registration waits, so one stalled call freezes the whole interpreter
## Problem

The Python surface holds the GIL across operations that can block for an unbounded time, which turns a one-caller stall into a freeze of every Python thread in the process.

The repo already establishes the right pattern in two places: `connect()` detaches around the gateway login, which can run to the second-factor window (`src/python/compat/client/mod.rs:140`), and `run()` detaches around `wait_for_data` (`:229-231`). The remaining blocking sites do not.

**Thread joins.** `disconnect()` joins the engine thread with the GIL held (`mod.rs:191-192`), as does `Drop` (`mod.rs:70-71`), which runs at garbage-collection time. If the engine is wedged — #254 describes a concrete way — the join never returns and the interpreter is gone with it.

**Control-channel sends.** All 33 `control_tx.send` call sites across `orders.rs`, `market_data.rs`, `reference.rs` and `account.rs` are blocking sends on a `bounded(64)` channel, with the GIL held. A full channel blocks every Python thread rather than the caller.

**Registration waits.** `find_or_register_instrument` and `register_mkt_data` block in `recv_timeout` for up to `REGISTRATION_TIMEOUT` (5 s) per call with the GIL held, so a slow or stalled engine freezes the interpreter for that long on an ordinary subscribe.

## Impact

None of these is wrong on its own — they are the same blocking calls the Rust API makes. The GIL is what changes the severity: in Rust a stalled call blocks its caller, in Python it blocks the interpreter, so an unrelated thread doing unrelated work stops too.

## Solution

`py.detach` around the join, the send and the registration wait, following the two sites that already do it. For the control sends, `try_send` with a Python-visible error is arguably better than detaching: a full channel means the engine is not draining, and a caller is better told that than parked.

## Notes

Read from v0.7.1 (9367845). Inventory is static; the wedge that makes the joins matter is #254.

userFRM:
## Problem

The Python layer held the GIL across operations that block. A thread join on `disconnect()` and on `Drop`, the registration waits that can sit for up to five seconds, and every control-channel send: each one blocked with the interpreter locked, so one stalled call froze every other Python thread in the process.

## What this changes

The GIL is released around the wait and reacquired after — the blocking operation is what moves inside the detach, and nothing Python-backed crosses that boundary. Commands are built while attached, so only an owned `ControlCommand` and a channel reference travel with the closure.

The sends stay **blocking**. The channel is bounded, so a burst applies backpressure and then succeeds, which is the behaviour callers already have; what was wrong was holding the GIL while waiting, not the waiting itself.

`Drop` mints its token with `Python::try_attach` rather than `Python::attach`. Dealloc can run during interpreter shutdown, and the common `App(EWrapper, EClient)` shape means the cyclic collector rather than refcounting may be what calls it. Where the token cannot be had, the join still happens — the handle is never dropped unjoined.

`ClientCore` is untouched. It is shared with the plain Rust API, which has no GIL to release, so the fix stays on the Python side that wraps it.

## Verification

`cargo check --offline --lib --features python` is clean, and the full gate passes.

Not run: `cargo test --lib --features python` and the pytest suite, neither of which links in this tree (#381). The unblocking behaviour and the `try_attach` shutdown path are established by reading pyo3 0.29's `Ungil` bounds and `detach`/`try_attach` contracts against this repo's existing usage in `connect()` and `run()`, rather than by exercising an interpreter.

Closes #271.

### DIFF
--- a/src/python/compat/client/account.rs
+++ b/src/python/compat/client/account.rs
@@ -1,6 +1,5 @@
 //! Account-related methods: positions, PnL, account summary/updates.
 
-use pyo3::exceptions::PyRuntimeError;
 use pyo3::prelude::*;
 
 use crate::types::*;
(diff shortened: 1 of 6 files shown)


userFRM:

PR #392, opened 2026-07-31
orders: reject non-finite and malformed order parameters (ibx#263)

### LINKED ISSUE #263: orders: malformed algo parameters and non-finite numbers are coerced into valid-looking values instead of being rejected
## Problem

Malformed order input is coerced into a valid-looking value and sent, rather than rejected. This is the same family as #224, #225 and #226 — the field is accepted, appears honoured, and the wire carries something the caller did not ask for. Two clusters remain.

**Algo parameters silently take defaults.** `parse_algo_params` reads every key through a lookup that returns an empty string when absent, then parses with a fallback:

- a missing or malformed numeric becomes `0.0`, a malformed boolean becomes `false` (`src/api/client/orders.rs:175`)
- an unknown `riskAversion` becomes `Neutral` (`:203`)
- a malformed DarkIce `displaySize` becomes `100` (`:232`)
- an unknown Adaptive priority becomes `Normal` (`src/client_core.rs:1420`)

Those coerced values are then encoded as real algo parameters (`order_builder.rs:850`). So `riskAversion="Aggresive"` submits a Neutral arrival-price algo, and `displaySize="abc"` submits a 100-share DarkIce display. The typo is invisible: no error, and the order works.

**Non-finite and out-of-range numbers reach the casts.** There is no finiteness check on the order path — `grep -n "is_nan\|is_finite" src/client_core.rs` returns nothing. Values go through saturating float-to-integer casts, so a NaN limit price becomes `0` and an infinite one becomes `i64::MAX`. A negative quantity, a negative `display_size`/`min_qty`/`parent_id` (`src/api/types.rs:425`, `:441`) and a negative `trailing_percent` are likewise clamped or reinterpreted rather than refused — validation only distinguishes exactly zero for the trail mode (`client_core.rs:1342`, `:1527`).

`trailing_percent` also loses precision: it is truncated to basis points with `(value * 100.0) as u32` (`client_core.rs:1527`) and re-rendered with two decimals (`order_builder.rs:1707`), so `1.239%` goes out as `1.23%`.

## Impact

An order that the caller believes is parameterised one way is submitted parameterised another, with no error anywhere. A NaN price becoming `0` is the worst of them, because a zero price is a plausible-looking value rather than an obviously broken one.

## Solution

Reject rather than default, which is the line #226 already drew for `transmit`: make the algo parsers return `Err` on unknown keys, unknown enum values and malformed numerics instead of falling back, and validate finiteness, sign and representable range before any cast.

For `trailing_percent`, either reject precision the wire cannot carry or document the basis-point granularity — silently rounding a trail distance is the same class of surprise.

## Notes

Read from v0.7.1 (9367845). Every coercion above is a literal `unwrap_or`/fallback arm in the source; no live run is needed to see them.

userFRM:
## Problem

Malformed and non-finite order parameters were coerced into valid-looking values rather than refused. `"NaN"` and `"inf"` parse successfully as `f64`, and the float-to-int casts that follow are saturating: NaN becomes 0, infinity becomes `i64::MAX`, and a finite magnitude too large for the wire's fixed point saturates the same way.

So an order carrying a nonsense price was accepted, silently turned into a different order, and sent.

## What this changes

**Algo parameters** are refused when present but malformed: an unparseable or non-finite number, a boolean that is not `0`/`1`/`true`/`false`, an unrecognised `riskAversion` or `adaptivePriority`, a `displaySize` that is not a non-negative integer. A *missing* key still defaults exactly as before — only a present-but-malformed value is now an error.

**Numeric order fields** are checked at `validate_order`, which both the Rust and Python `place_order` already call before any cast. Non-finite values are refused for every price and amount field, as is a finite magnitude whose fixed-point form would overflow `i64`; negative values are refused for quantity, display size, minimum quantity, parent id and trailing percent.

The three fields that use `f64::MAX` as their "not set" sentinel are only checked when they hold something else, so an unset field is not mistaken for an invalid one.

`trailing_percent`'s basis-point truncation is left alone and documented at the cast site: that is rounding a well-formed value, not coercing a nonsense one.

## Note on `aux_price`

The pre-existing zero-check on `aux_price` cannot see NaN, because `NaN != 0.0`. The blanket finiteness check is what catches it.

## Tests

Twenty-two, covering each rejection and each preserved default. Twelve production mutations were applied one at a time, each confirmed to compile, the matching test confirmed to fail by name, and the revert diffed byte-for-byte against a saved baseline to confirm it landed on the intended line — including `require_finite_price`'s overflow clause in isolation from its finiteness clause, so the range check is shown not to be redundant.

Closes #263.

### DIFF
--- a/src/api/client/orders.rs
+++ b/src/api/client/orders.rs
@@ -173,74 +173,106 @@ impl EClient {
 }
 
 /// Parse algo strategy and TagValue params into internal AlgoParams.
+///
+/// A key the caller never set defaults the way IB's own algos do (0.0,
+/// false, or the documented default enum value). A key the caller *did*
+/// set — even to an empty string — is refused if it doesn't parse, instead
+/// of silently taking that same default: a typo like `riskAversion="Aggresive"`
+/// used to submit a Neutral algo with no error, and `maxPctVol=""` used to
+/// submit 0.0. See ibx#263.
 pub fn parse_algo_params(strategy: &str, params: &[TagValue]) -> Result<AlgoParams, String> {
-    let get = |key: &str| -> String {
-        params.iter()
-            .find(|tv| tv.tag == key)
-            .map(|tv| tv.value.clone())
-            .unwrap_or_default()
+    let get = |key: &str| -> Option<String> {
+        params.iter().find(|tv| tv.tag == key).map(|tv| tv.value.clone())
     };
-    let get_f64 = |key: &str| -> f64 { get(key).parse().unwrap_or(0.0) };
-    let get_bool = |key: &str| -> bool {
-        let v = get(key);
-        v == "1" || v.eq_ignore_ascii_case("true")
+    let get_str = |key: &str| -> String { get(key).unwrap_or_default() };
+    let get_f64 = |key: &str| -> Result<f64, String> {
+        let raw = match get(key) {
+            None => return Ok(0.0),
+            Some(raw) => raw,
+        };
+        let v: f64 = raw.parse().map_err(|_| format!("Invalid {} '{}': expected a number", key, raw))?;
+        if !v.is_finite() {
+            return Err(format!("Invalid {} '{}': must be a finite number", key, raw));
+        }
+        Ok(v)
+    };
+    let get_bool = |key: &str| -> Result<bool, String> {
+        let raw = match get(key) {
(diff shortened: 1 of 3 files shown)


userFRM:

PR #394, opened 2026-07-31
orders: distinguish Rejected from Inactive once the ibapi string collapses them (ibx#250)

### LINKED ISSUE #250: orders: Rejected and Inactive collapse to one status string, so the reason is dropped and a parked order vanishes from the open-order snapshot
## Problem

The wire distinguishes `39=8` (Rejected) from `39=I` (Inactive), and the engine maps them to distinct internal states (`src/engine/hot_loop/ccp.rs:837-838`). `order_status_str` then collapses both to the string `"Inactive"` (`src/client_core.rs:271-276`) — the comment explains why, since ibapi has no Rejected string. From that point on a caller cannot tell a dead order from a held one.

Three consequences follow from that single collapse.

**The reason is dropped.** The gateway's text (tag 58) and reject code (tag 103) are parsed and written to the log (`ccp.rs:802-812`), but reach the API only for `39=8`, through `OrderState.completed_status` (`ccp.rs:1064-1071`, `_ => String::new()`). A `39=I` transition surfaces as a bare status change: no `error()` callback, and the reason exists nowhere a caller can reach.

**The order disappears from the snapshot, and it has to.** `is_open_status` (`src/client_core.rs:248-258`) omits `"Inactive"`, applied at `collect_open_orders` (`:822`, `:842`) and `drain_open_orders` (`src/bridge.rs:293`). It cannot simply admit `"Inactive"`, because after stringification that value is also every rejected order — widening the whitelist would resurrect genuinely dead orders into the open book. So a reactivatable order is hidden by a filter that is only defending itself against the conflation above.

That contradicts the engine one layer down, which deliberately models Inactive as non-terminal and reactivatable: it sits in the working tier alongside Submitted with the comment "Inactive orders can reactivate" (`src/types.rs:85-87`), `is_terminal` excludes it (`:96-98`), and the order stays in every map.

**A replace that the gateway parks looks like success.** Nothing reconciles on Inactive — the only restore path is the `35=9` cancel-reject handler (`ccp.rs:1125-1180`). The engine record is overwritten with the caller's new values when the replace is sent (`order_builder.rs:1451-1456`), and the client cache is overwritten with status `"PendingSubmit"` at the end of `place_order` (`track_order`, `client_core.rs:768-773`). So every client-visible surface asserts the modify succeeded, while `req_open_orders` returns nothing for that order.

## Solution

Discriminate Rejected from Inactive before stringification — carry it on the update or on `TrackedOrder` — rather than widening `is_open_status`, which would readmit rejected orders.

Then route 58/103 into the update or error path for non-`39=8` transitions, and include genuinely-Inactive rows in the snapshots once they can be told apart.

## Impact

An order can be parked by the gateway with a stated reason, and a caller sees a status string shared with rejections, no reason at all, and an empty open-order book. For a resting stop that is the difference between "held, will reactivate" and "gone".

It is also an ibapi parity divergence: TWS-API consumers receive Inactive orders through `openOrder`/`orderStatus` and rely on seeing them — the Rejected-to-`"Inactive"` mapping exists precisely because those rows are user-visible there.

## Notes

Everything above is from reading v0.7.1 (9367845); no live run is cited and none is needed to confirm it. A `39=I` transition can be observed with `RUST_LOG` at info through the existing `ExecReport: 39=I ... 58=...` line, which is the code's own logging rather than an external capture.

userFRM:
## Problem

The wire distinguishes `39=8` (Rejected) from `39=I` (Inactive), and the engine models them as distinct states — but `order_status_str` collapses both to the string `"Inactive"`, correctly, because ibapi has no Rejected string. From that point a caller cannot tell a dead order from a held one, and three consequences follow.

**The reason is dropped.** Tags 58 and 103 are parsed and logged, but reach the API only for `39=8`, through `OrderState.completed_status`. A `39=I` transition surfaces as a bare status change with the reason nowhere a caller can reach.

**The order disappears from the snapshot, and it has to.** `is_open_status` omits `"Inactive"` — and cannot simply admit it, because after stringification that value is also every rejected order. Widening it would resurrect genuinely dead orders. So a reactivatable order is hidden by a filter defending itself against the conflation above, contradicting the engine one layer down, which deliberately models Inactive as non-terminal and reactivatable.

**A replace the gateway parks looks like success.** Every client-visible surface asserts the modify succeeded while `req_open_orders` returns nothing for that order.

## What this changes

`order_status_str` is untouched — both still map to `"Inactive"`, which is the ibapi contract.

`completed_status` is the discriminator. It is already non-empty only for a terminal status and empty otherwise, and a rejection falls back to the literal `"Rejected"` when tag 58 is absent, so it is never empty for one. `is_open_or_reactivatable(status, completed_status)` admits `"Inactive"` only when it is empty. **`is_open_status` itself is not widened**, which is the trap the issue names.

`TrackedOrder` carries whether it was rejected, set at its one write choke point. That function now takes the `OrderStatus` enum rather than the already-collapsed string, so the distinction is read before it is lost and the compiler enforces the call sites.

A `39=I` transition delivers its reason through the wrapper's `error()` callback — the same path a cancel reject already uses — alongside the status change.

## Note for merging

This changes `ClientCore::update_order_status`'s signature. #362 changes the same function for an unrelated reason (an order recovered from an earlier session has no local entry to update). The two are independent but will want resolving together.

## Tests

Eleven, covering the discriminator in all four quadrants, both `collect_open_orders` admission sites, the `drain_open_orders` filter, the `TrackedOrder` write, the ccp gating condition including the Rejected/Inactive conflation trap, the reason formatting, and the dispatch plumbing.

Each production line touched was independently reverted and the intended test confirmed to fail by name.

Closes #250.

### DIFF
--- a/src/api/client/dispatch.rs
+++ b/src/api/client/dispatch.rs
@@ -112,7 +112,7 @@ impl EClient {
                 update.order_id as i64, status, update.filled_qty as f64,
                 update.remaining_qty as f64, 0.0, update.perm_id, update.parent_id, 0.0, 0, "", 0.0,
             );
-            self.core.update_order_status(update.order_id, status, update.filled_qty as f64, update.remaining_qty as f64);
+            self.core.update_order_status(update.order_id, update.status, update.filled_qty as f64, update.remaining_qty as f64);
         }
 
         // Cancel rejects → error
@@ -120,6 +120,12 @@ impl EClient {
             let code = if reject.reject_type == 1 { 202 } else { 10147 };
             let msg = format!("Order {} cancel/modify rejected (reason: {})", reject.order_id, reject.reason_code);
             wrapper.error(reject.order_id as i64, code, &msg, "");
+        }
+
+        // Inactive (39=I) order reasons → error (ibx#250). order_status above
+        // already reported the "Inactive" string; this carries why.
+        for (order_id, code, msg) in self.shared.orders.drain_order_inactive() {
+            wrapper.error(order_id as i64, code as i64, &msg, "");
         }
 
         // What-if → open_order(contract, order, OrderState) + order_status (iso with ibapi)
--- a/src/api/client/tests.rs
+++ b/src/api/client/tests.rs
@@ -1699,6 +1699,51 @@ fn process_msgs_dispatches_order_updates() {
     assert!(w.events.iter().any(|e| e.starts_with("order_status:45:Inactive")));
 }
 
+/// ibx#250: a parked (39=I) order's reason reaches the caller through
+/// Wrapper::error, on top of the order_status "Inactive" callback above —
+/// ibapi has no callback dedicated to "order held with reason".
+#[test]
+fn process_msgs_dispatches_inactive_reason_as_error() {
+    let (client, _rx, shared) = test_client();
+    shared.orders.push_order_inactive(46, 399, "Order held pending margin check".into());
+    let mut w = RecordingWrapper::default();
+    client.process_msgs(&mut w);
+    assert!(w.events.iter().any(|e| e == "error:46:399:Order held pending margin check"));
+}
+
+/// ibx#250 end-to-end: a genuinely-Inactive order dispatched through the real
+/// `process_msgs` path (not a direct `ClientCore` call) stays in the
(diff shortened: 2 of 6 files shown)


userFRM:

PR #396, opened 2026-07-31
farm: resubscribe from the list the disconnect preserves (ibx#368)

### LINKED ISSUE #368: farm: reconnect resubscribes from the collection disconnect clears, so every farm reconnect comes back with no L1 data
## Problem

`FarmState::handle_disconnect` clears the collection that `FarmState::reconnect` reads to rebuild the subscriptions.

Disconnect (`src/engine/hot_loop/farm.rs:883`):

```rust
self.disconnected = true;
self.md_req_to_instrument.clear();
self.instrument_md_reqs.clear();
```

Reconnect derives the resubscription list from that same collection (`farm.rs:917`):

```rust
let active: Vec<...> = self.instrument_md_reqs.iter()
    .filter_map(|(id, _)| { ... })
```

By the time `reconnect` runs, `instrument_md_reqs` is empty, so `active` is empty and no L1 subscription is re-issued.

The depth path does not have this problem, and the comment above the clears says why: `depth_resub_info is preserved for resubscription on reconnect`. `md_resub_info` is preserved in the same way — it holds the symbol, exchange, security type and the rest — but nothing reads it on its own, because the iteration is driven by `instrument_md_reqs`.

## Impact

Every farm reconnect comes back with zero market data. A caller still holds valid `InstrumentId`s and the client reports connected, but quotes stay at the values `zero_all_quotes` left behind and never update. Depth subscriptions recover; L1 does not.

This is the ordinary path, not an edge case — it runs on any farm liveness timeout.

## Solution

Drive the resubscription from what disconnect preserves rather than from what it clears. `md_resub_info` already carries everything `active` needs except the con_id, which comes from `context.market.con_id(id)`, so iterating `md_resub_info` and looking up each con_id yields the same tuple without depending on the cleared collection.

## Notes

Read from v0.7.1 (9367845). Verifiable offline from the two functions: `instrument_md_reqs` is cleared at `farm.rs:886` and is the iteration source at `farm.rs:917`.

userFRM:
## Problem

`handle_disconnect` clears `instrument_md_reqs`, and `reconnect` built its resubscription list from that same collection. By the time a reconnect read it the list was empty, so no L1 subscription was ever re-issued.

Every farm reconnect came back with no market data. A caller still held valid instrument ids and the client reported connected, while quotes stayed at the values `zero_all_quotes` left behind and never updated again. Depth recovered; L1 did not.

## What this changes

The resubscription is driven by `md_resub_info`, which the disconnect preserves for exactly this reason — the comment above those clears says so, and the depth path beside it already works this way.

**An unsubscribe issued while the farm is down still takes effect.** The disconnect clears the live request list, so the cancel used to return early without dropping the resubscription record, and the reconnect brought back a subscription the caller had cancelled.

**A record is replaced rather than skipped** when one already exists for the slot. A slot holds one contract at a time, so keeping the earlier entry meant a slot reused during an outage had no record for its new contract, and the identity check then dropped the stale one — leaving a live subscription unrestored.

**Identity is the contract, not the slot.** Instrument ids are reused, so a slot freed while the farm was down and handed to another contract would otherwise have the old symbol replayed under it. Each record carries the con_id it was issued under. Zero is a valid con_id — a descriptive subscription is identified by its fields rather than a number — so those are told apart by symbol instead.

## Tests

- `a_farm_reconnect_resubscribes_what_was_subscribed`
- `an_unsubscribe_while_disconnected_is_not_replayed`
- `a_reused_slot_records_the_contract_that_holds_it`
- `a_reused_instrument_slot_does_not_replay_the_old_subscription`
- `a_reused_slot_without_a_con_id_is_told_apart_by_its_symbol`

Each fails by name against a compiling reversion of the production line it covers. The socket reads are bounded so a test that stops receiving fails rather than hanging, and the frame is read to its declared FIXCOMP length rather than trusting a single `read`.

Closes #368.

### DIFF
--- a/src/engine/hot_loop/farm.rs
+++ b/src/engine/hot_loop/farm.rs
@@ -25,7 +25,9 @@ pub(crate) struct FarmState {
     /// Primary depth subscription params for reconnect: (req_id, con_id, exchange, sec_type, num_rows, is_smart_depth).
     depth_resub_info: Vec<(u32, i64, String, String, i32, bool)>,
     /// Option resub info: (instrument, symbol, exchange, sec_type, last_trade_date, strike, right, multiplier, mode_9887).
-    md_resub_info: Vec<(InstrumentId, String, String, String, String, f64, String, String, i32)>,
+    /// Carries the con_id the subscription was issued under, so a reconnect
+    /// can tell a reused instrument slot from the contract it was taken for.
+    md_resub_info: Vec<(InstrumentId, i64, String, String, String, String, f64, String, String, i32)>,
     pub(crate) disconnected: bool,
     pub(crate) tick_buf: Vec<tick_decoder::RawTick>,
     pub(crate) farm_msg_buf: Vec<Vec<u8>>,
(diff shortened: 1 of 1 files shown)


userFRM:
Same defect as ibx#288: the reconnect re-subscribes from a collection the disconnect has already cleared, so a farm reconnect comes back with no L1 data. ibx#294 carries the fix, driving the resubscribe off the record that survives the disconnect.

Closing this in favour of ibx#288 and ibx#294, which reached it first.

PR #398, opened 2026-07-31
tests: make the paper-trading compatibility suite compile again (ibx#397)

### LINKED ISSUE #397: tests: the paper-trading compatibility suite does not compile, so all 154 phases are dead
## Problem

`tests/ib_paper_compat` does not compile. Thirteen call sites construct request and command variants that have since gained fields:

```
error[E0063]: missing field `filters` in initializer of `ibx::types::ControlCommand`
error[E0063]: missing field `trail_stop_price` in initializer of `ibx::types::OrderRequest`
error[E0063]: missing fields `adjustable_trailing_unit` and `adjusted_trailing_amount` ...
```

So the entire target is dead: fourteen entry points driving a hundred and fifty-four phases against a paper account, none of which can be built, let alone run.

## Why it went unnoticed

The suite only runs with live credentials, so ordinary development never builds it. `cargo check` without a target name does not reach it, and `cargo test` without credentials skips it — the failure only appears if someone asks for that target by name.

It has also become the baseline: work in this repo compares the suite's error set before and after a change to prove nothing was broken. Thirteen errors is the number being matched, which makes a broken target look like a fixed point rather than a defect.

## Impact

The only end-to-end coverage against a real gateway is unbuildable. Every wire-level claim in this repo currently rests on unit tests and reasoning, with nothing exercising a full session — which is the coverage that would catch a field emitted in the wrong position, an order the gateway rejects, or a reply the parser mishandles.

## Solution

Give each site the value it had before those fields existed: default contract-details filters, no initial trailing-stop trigger, and no adjustable trailing amount or unit. No phase behaviour changes.

Then keep it building: a `cargo check --tests` in CI reaches every target by name and would have caught each of these the day the field was added.

userFRM:
## Problem

`tests/ib_paper_compat` does not build. Thirteen call sites construct request and command variants that have since gained fields, so the whole target fails to compile and every phase in it is dead — **fourteen entry points driving a hundred and fifty-four phases** against a paper account.

Nothing reported it. The suite only runs with live credentials, so ordinary development never builds it, and `cargo check` without a target name does not reach it.

It had also quietly become the baseline: work in this repo compares the suite's error set before and after a change to show nothing broke. Thirteen errors is the number being matched, which makes a broken target look like a fixed point.

## What this changes

Each site gets the value it would have had before those fields existed — default contract-details filters, no initial trailing-stop trigger, no adjustable trailing amount or unit. No phase behaviour changes.

The target now compiles clean and enumerates its fourteen tests.

## Why it matters

This is the only end-to-end coverage against a real gateway. Every wire-level claim in this repo currently rests on unit tests and reasoning, with nothing exercising a full session — which is the coverage that catches a tag emitted in the wrong position, an order the gateway rejects, or a reply the parser mishandles.

## Run against a paper account

With it building, the suite runs. All seven phases pass:

```
cancel_by_perm_id_phase_live         PASS   (ibx#191 PR B)
cross_session_recovery_phase_live    PASS   (ibx#191 PR A)
reclaim_and_symbol_search_phase_live PASS   (ibx#233 / ibx#228)
rtt_ping_phase_live                  PASS   (ibx#158)   RTT 3.78 ms
snap_to_tick_phase_live              PASS   (ibx#216)
submit_ex_bracket_child_phase_live   PASS   (ibx#224 / ibx#215)
timeout_sweeps_phase_live            PASS   (ibx#231 / ibx#227)

test result: ok. 7 passed; 0 failed; finished in 195.65s
```

So the phases still describe the server's behaviour correctly — the suite had gone stale in its call sites, not in what it asserts.

## Keeping it built

`cargo check --tests` reaches every target by name and would have caught each of these on the day the field was added. #300 adds a CI workflow; this belongs in it.

Closes #397.

### DIFF
--- a/tests/ib_paper_compat/account.rs
+++ b/tests/ib_paper_compat/account.rs
@@ -389,6 +389,7 @@ pub(super) fn phase_enriched_order_cache(conns: Conns) -> Conns {
     control_tx.send(ControlCommand::FetchContractDetails {
         req_id: 9999, con_id: 756733, symbol: String::new(),
         sec_type: String::new(), exchange: String::new(), currency: String::new(),
+        filters: Default::default(),
     }).unwrap();
 
     let order_id = next_order_id();
@@ -548,6 +549,7 @@ pub(super) fn phase_enriched_open_orders(conns: Conns) -> Conns {
     control_tx.send(ControlCommand::FetchContractDetails {
         req_id: 9998, con_id: 756733, symbol: String::new(),
         sec_type: String::new(), exchange: String::new(), currency: String::new(),
+        filters: Default::default(),
     }).unwrap();
 
     let order_id = next_order_id();
@@ -773,6 +775,7 @@ pub(super) fn phase_enriched_exec_details(conns: Conns) -> Conns {
     control_tx.send(ControlCommand::FetchContractDetails {
         req_id: 9997, con_id: 756733, symbol: String::new(),
         sec_type: String::new(), exchange: String::new(), currency: String::new(),
+        filters: Default::default(),
     }).unwrap();
 
     let order_id = next_order_id();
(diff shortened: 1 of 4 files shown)


userFRM:

PR #401, opened 2026-08-03
engine: park the hot loop while every transport is down (ibx#399)

### LINKED ISSUE #399: Rust EClient fails to reconnect and causes 100% CPU infinite spin-loop when connection is lost
### Environment
- **Crate Version:** v0.7.1 (or latest commit)
- **Language Binding:** Rust
- **OS:** macOS / Linux (Laptop sleep/resume or network interruption)

### Problem Description
When running a Rust client, if the connection is interrupted (e.g., closing the laptop lid, putting the system to sleep, or temporary network drops), the `ibx` engine fails to reconnect and immediately traps the process in an infinite spin-loop, driving **CPU usage to 100%**.

### Root Cause & Log Analysis
Upon waking up the system or experiencing a network flash-drop, the `hot_loop` continuously spams the following warning at a zero-delay frequency:
```text
2026-07-31T09:48:08.364531Z ERROR ibx::engine::hot_loop::ccp: CCP connection lost: connection closed
2026-07-31T09:48:11.537858Z  WARN ibx::engine::hot_loop: CCP auto-reconnect skipped: no credentials
2026-07-31T09:48:13.579228Z ERROR ibx::engine::hot_loop::hmds: HMDS connection lost: Connection reset by peer
2026-07-31T09:48:18.489932Z  WARN ibx::engine::hot_loop: Farm auto-reconnect skipped: no credentials (host empty or auth missing)
2026-07-31T09:50:21.157930Z  WARN ibx::engine::hot_loop: CCP auto-reconnect skipped: no credentials
... (repeats indefinitely, hogging the entire CPU core)
```

This behavior stems directly from the known gap where `EClient::connect_inner` (`src/api/client/mod.rs`) never propagates credentials via `update_reconnect_auth` to the `hot_loop` for Rust callers, while Python handles it perfectly.

Because `auth.host` remains empty, the reconnect schedulers (CCP, HMDS, and Farms) all bail out early via safety guards like:
```rust
let auth = match self.reconnect_auth.as_ref() {
    Some(a) if !a.host.is_empty() => a,
    _ => return, 
};
```
However, since it returns instantly without executing any asynchronous network I/O or performing any explicit time-deferred backoff yields (because the network attempt was technically skipped), the event loop or retry mechanism drops into a tight **busy-wait / infinite spin**. It constantly retries, prints the warning, and loops back instantly, starving the CPU thread.

### Impact
- Any temporary connection loss or system sleep completely freezes/overloads the Rust process, making it impossible to recover without killing the entire application.
- The beautiful jittered backoff mechanics introduced in `#218` and `#219` become entirely unexecuted dead code for Rust callers under this state.

### Suggested Solution
1. **Fix the Core Gap**: Ensure `update_reconnect_auth` is properly invoked within `connect_inner` for Rust callers, passing the four authentication items just like the Python path does.
2. **Defensive Looping**: In the `hot_loop` or the scheduler, if a reconnect attempt is explicitly `skipped` due to missing credentials, ensure the task explicitly yields or sleeps (e.g., `tokio::time::sleep`) instead of instantly slamming back into the loop cycle, avoiding 100% CPU spinning.

userFRM:
## Summary

- The hot loop busy-polls the farm, auth and historical sockets and never yields. While a session is up that is the design: a poll that finds a tick returns it with no syscall in the way, and `core_id` exists so the spin can own a core.
- With every transport down there is nothing to poll. Each pass runs the same three polls, gets nothing, and starts again — measured at roughly 925,000 iterations a second, which is one core pinned at 100% for the length of the disconnect.
- Parking 1 ms in that state, and only that state, leaves the connected path byte for byte as it was. Nothing waits on the loop meanwhile: reconnects are already scheduled on the ibx#218 backoff in seconds, and shutdown is read once per pass.

## On the reported cause

The report puts the CPU on the reconnect scheduler retrying in a tight loop. The timestamps in its own log are two minutes apart:

```
09:48:11.537  CCP auto-reconnect skipped: no credentials
09:50:21.157  CCP auto-reconnect skipped: no credentials
```

That is the 60-second re-check in `maybe_spawn_ccp_reconnect` doing its job. The spin is the loop body itself, and it is there whether or not credentials are cached — a connected engine spins too, which is intended, and a disconnected one spins for nothing, which is not.

Closes #399.

## Test plan

- [x] `cargo test --lib` — 803 pass. The two `config::expiry_tests` failures are pre-existing on main and identical on a pristine checkout of 9367845.
- [x] `a_loop_with_every_transport_down_does_not_spin` drives `run()` with all three transports marked down, shuts it down after 60 ms and asserts the iteration count. Counting iterations rather than CPU time is what separates a parked loop from a spinning one on any machine.
- [x] Mutation: deleting the sleep takes that test from ~60 iterations to 55,554 in the same window, and it fails by name.
- [ ] Not covered: the laptop sleep/resume path the report describes. The guard is on the disconnected flags, so it does not depend on how the disconnect happened.

### DIFF
--- a/src/engine/hot_loop/mod.rs
+++ b/src/engine/hot_loop/mod.rs
@@ -361,6 +361,17 @@ impl HotLoop {
 
             // 6. Wake any waiting consumers (e.g. Python event loop)
             self.shared.notify();
+
+            // 7. With every transport down there is no socket to poll and no
+            //    socket to be quick for, so the spin above buys nothing and
+            //    costs a core — on a laptop that is the battery and the fans,
+            //    which is how this gets noticed (ibx#399). Parking only in that
+            //    state leaves the connected path exactly as it was; a reconnect
+            //    is scheduled on a backoff measured in seconds, so a millisecond
+            //    here delays nothing, including shutdown.
+            if self.farm.disconnected && self.ccp.disconnected && self.hmds.disconnected {
+                std::thread::sleep(std::time::Duration::from_millis(1));
+            }
         }
     }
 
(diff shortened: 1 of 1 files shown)


userFRM:

PR #409, opened 2026-08-26
Fast-forward main onto the fork

userFRM:
This is `userFRM:main`, offered as a fast-forward.

## TL;DR

| | |
|---|---|
| Commits | **1,654** — 907 fixes, 208 features, 172 test, 232 docs, 73 refactors, and one merge of your `main` |
| Files | 475 |
| Merge | Clean fast-forward. Your `main` through `e9d919f` is merged in, `main..upstream/main` is 0, nothing of yours rewritten or dropped |
| Undo | `git reset --hard e9d919f` |
| CI | Green on the head — [actions](https://github.com/userFRM/ibx/actions) |
| Tested | 154 phases against a real session, on paper **and** live. **3,660 offline tests + 197 broker-session tests** |
| Soak | 14 hours held open on one session, through the venue's maintenance window — 0 losses, 0 reconnects |
| Order fields dropped | **0** of 154 |
| Call arguments dropped | **0** of 632 — 597 read, 35 stated by name |
| API parity | **78 of 78** calls and 90 of 90 callbacks, served on both languages |
| Beyond the documented API | **77 more calls**, and **128 series** the venue states that no documented call asks for |
| Order types | 24 |
| Requests that reach the venue | 74 of 80 — 1 answered from the session, 1 refused by name, 4 withdraw something never running. **0 missing, 0 silent** |

---

## Merge safety

- `git merge-base upstream/main main` = `e9d919f` = the tip of `main` here.
- `git rev-list --count main..upstream/main` = 0 — nothing to resolve, no conflicts.
- Nothing of yours is rewritten, reordered or dropped.
- One command undoes it: `git reset --hard e9d919f`.
- Partial take: cut at `7cfa6e56` (see [Taking it](#taking-it)).

---

## Where the counts come from

The fix counts on each section below are that section's own scope, countable
from the log:

```
git log --oneline e9d919f..main -E --grep="^fix\(orders\)" | wc -l
```

They sum to 440 of the 907 fixes. The other 467 carry no scope and are not
attributed to a section rather than being assigned to one by guesswork — the
tables under each heading are what changed in that area, not an exhaustive list.

---

## How it was tested

- **Reproduced first.** Fixes were reproduced before they were made, and landed with tests that fail without them — 172 test commits.
- **Against a real session.** 154 phases end to end: order placement, replacement, cancellation, brackets and legs, executions, commission reports, account and position downloads, market data, depth, historical series, contract lookups, reconnection.
- **Live as well as paper.** Paper and live speak the same wire — identical messages, fields and sequencing, differing only in which prints arrive. Order paths have additionally been run against a live account, real fills, a real book.
- **No silent passes.** A phase cannot skip on silence — each names what refused it. A phase cannot pass on an order that never reached the wire.
- **Held open.** One session run for 14 unbroken hours across the venue's nightly maintenance window: no losses, no reconnects, quotes and orders still answering at the end.
- **Counted by script, re-derived in CI.** A documentation claim fails the build when the source stops supporting it.

```
python scripts/gate.py          # runs what CI runs; --list prints the suites first
cargo test --lib
git diff --exit-code docs/      # the generated reference matches its source

IB_USERNAME=... IB_PASSWORD=... cargo test --test ib_paper_compat -- --test-threads=1 --nocapture
```

---

## Orders — 98 fixes

| Was | Now |
|---|---|
| Algo params, conditions and unmodelled params dropped between caller and wire | Travel as stated; condition flags sit on the tags that mean them |
| Flags the caller never set were sent anyway | An algo flag, time or aversion not stated is not sent |
| Pegged orders lost offset, cap, snap-to offset, benchmark price | All carried, including an offset of zero |
| Prices truncated; sign lost on prices with no whole part | Rounded; sign kept |
| 13 order types the venue replaces were refused | Accepted, plus volatility, both relative combinations, passively relative, peg-to-best, peg-by-minute, delayed activation |
| Replace kept terms the venue had refused | The venue's word settles it; refused terms do not stand |
| A correction the venue sent twice reopened a finished order and republished its quantity as working | Read once; the second copy changes nothing |
| A second refusal for the same cancel read the echoed name as digits and retired an unrelated live order | Resolved through the cancel's own name |
| The async client's two withdrawal calls waited out a reconnect's replay on the reactor | Off it, as every other order call there |
| An order's quantity was bounded where a double stops multiplying exactly, so a cash order stated in currency units was refused | Bounded where the field itself ends, four orders of magnitude further; the whole part scales in integer arithmetic |
| A rejection answering a replace was read as how the order finished | It is not the order's state |
| A replacement could not itself be replaced | It can |
| Type change on a linked order corrupted links | Refused; links kept both ways |
| Trigger survived a replace to a type without one | Dropped |
| A working order this client did not place was placed again | Replaced |
| Bracket legs lost group and parent on replace | Kept |
| A held parent withdrawn orphaned its children | Children go with it; a parent releases them |
| A set that could not be placed whole left partial legs | Withdrawn |
| Filled orders reported as working | Not reported as working |
| Orders done but unfilled tracked forever | Stop being tracked |
| Order ids counted from a file | Counted from what the account is working, past every id the venue has named |
| A cancel the venue refused left the order reading as working | Reads correctly |
| A failed write reported as a rejection | Reported as unknown |
| A refusal with no words did not end the order | Ends it |

---

## Python surface — 69 fixes

| Was | Now |
|---|---|
| A figure the venue left out of the option model was handed to the wrapper as nothing at all, and the wrapper's own method declared it a number — so the call was refused, and the refusal left the reading loop and closed the session on every caller that had not written that method | Declared as a figure the venue may have left out. Every test here writes that method itself, which is why nothing saw it; a check now reads each call against the signature it lands on |
| Programs written against the reference client would not import | Its module layout, constants, enumerations and run-together names all answered |
| A model calculation on a described contract claimed the watch the request already held, and cancelling it withdrew that subscription | Only a watch this call opened is its own |
| A handler that closed the session from inside `connectAck` still received the accounts and the next order id | The startup stops with the session |
| Constructor spellings rejected | The reference API's own spelling accepted |
| Enum-typed fields handed back raw values | Hand back the member |
| `KeyboardInterrupt` / `SystemExit` swallowed in callbacks | Propagate |
| A raising callback lost its answer | Keeps it |
| A raising position handler ended the dispatch loop | Does not |
| A handler raising during connect failed the session | Does not |
| GIL held across blocking engine calls and blocking waits | Released |
| Two connects could both win | Cannot |
| A client holding itself was never collected | Collected |
| Request ids truncated; callbacks handed partial contracts | Whole ids, whole contract |
| Ticks named differently from the reference | Every tick the venue numbers named as the reference names it |
| 5 places Python answered differently from Rust | Answer the same |
| 4 things the model classes changed or dropped in transit | Carried |

---

## Engine — 66 fixes

| Was | Now |
|---|---|
| Slots reused while something still pointed at them | Not reused; resident while holding market data |
| Withdrawing a quote took the decoder mapping for a news subscription asked for under its own request | News is withdrawn only by its own withdrawal |
| A historical or calendar farm that came up long past its deadline was installed anyway | Each farm's clock is read before its result is taken |
| A slot that changed hands carried the wrong contract's order | An order states the contract it means |
| A slot did not return when its last order ended | Returns to the table |
| 5 answers belonged to another question | Go to the request that asked |
| 5 requests answered from something other than the venue | Answer from the venue |
| 4 answers taken after the moment they belonged to | Discarded |
| 6 answers said something had happened when it had not | Do not |
| Transports shared one recovery clock | Each has its own |
| Hot loop spun with every transport down | Parks |
| Hot loop could wait on something that never comes | Cannot |
| A quote feed the venue would not take back killed the session | Ends the feed only |
| Fills lost for orders the session does not track | Booked |
| A duplicate execution suppressed the whole report | Suppresses the fill only |
| Missing position quantity read as flat | Read as no quantity |
| Orders unaccounted for after recovery were silent | Reported |

---

## Historical data — 39 fixes

| Was | Now |
|---|---|
| 7 series the venue carries were refused, incl. `ADJUSTED_LAST` | Served |
| Midpoint, rate, volatility asked under names the venue does not answer | Asked under the venue's names |
| `TRADES` meant something other than the vendor's `TRADES` | Same meaning |
| Corporate actions discarded | Read |
| Bars dated on the wrong clock, written in the wrong shape | Dated on the venue's clock, written as the request asked |
| A bar the fold could not build was folded anyway | Refused |
| A refusal ended queries it did not name | Reaches only the query it names |
| A withdrawal took more than it withdrew | Takes what it withdrew, only |
| A sloppy answer came back as a plausible wrong number | Refused |
| The historical farm gave up after six attempts | Keeps retrying |

---

## Market data — 70 fixes

| Was | Now |
|---|---|
| A contract named by id alone was refused | Looked up at the venue |
| The slot getter answered from the cache after the engine had given the slot to another contract | Answers through the same lookup every other reader uses |
| Dropping a live-bar stream never sent its withdrawal, so the venue went on closing bars at a reader that had gone | Withdrawn under the mark its subscribe was numbered with |
| An unknown contract was assumed a US stock | The venue is asked what it is |
| A request joining a live subscription got no quote | Owed the quote as it stands, and its `tickReqParams` |
| A snapshot was handed to a second caller as a subscription | It is not; the chargeable snapshot is asked for |
| A finished snapshot withdrew what replaced it | Withdraws itself |
| A snapshot ended on a pause | Ends where the venue ends it |
| A stream was withdrawn by the wrong number | Withdrawn by the number the venue gave it |
| A feed given up on kept taking subscriptions | Takes none; refused where the caller asks |
| L1 did not survive a disconnect | Re-subscribes from the surviving record |
| A drop mid-replay lost what the replay had not reached | Kept |
| A refused book kept being asked for | Stops |
| A series framed behind a count of its length was read from the first byte, so the count's low byte joined the record's first field name — three contracts stated one field under three different names | Read from behind the count |
| A discarded payload was written to the log in full, so one discard of eight megabytes became a single line of sixteen million characters | Five hundred and twelve bytes of it, with the size that went stated beside it |
| What a contract's series stated was kept for every contract a session had ever watched | Kept for as many as this client can hold a slot for, the one heard of longest ago making way |

---

## What the session states that no documented call asks for

A quote subscription carries far more than prices. The venue publishes well over
a hundred series against a contract and the documented API forwards a handful;
the rest have no message, so a program had no way to ask. **128 are read now**,
against nine when this was opened, and they reach a caller through calls of
their own rather than under tick numbers this client made up.

| | |
|---|---|
| The venue's own fields, as text | 17 series — the analyst ratings, what institutions and insiders hold, the shares on issue **and the float**, fund terms, the screening scores, margin, technical readings, the company's accounts and its calendar |
| Figures | 34 series — a bond's accrued interest and its analytics, what a future's margin is, what one contract delivers, volatility, the volume a contract opens and closes on, what an option is worth in and out of the money |
| Numbered tables | 10 series — among them **where a contract's volatility stands against its own past**: the rank, the percentile and the high and low, for the volatility the market implies and again for the one it has shown, over a quarter, a half year and a year |
| Paired runs | 2 series — the volatility the model holds against each point of a curve, and the weight it puts on each price a contract might reach |
| Rows of three | 6 series — a book dealt in size, the legs of a scanned strategy, and the packed quotes for a contract once dealing has closed |
| The option model at the close | Kept apart from the standing one, with every greek the venue states |
| A spread scan | Asked for and answered: the strategies the venue finds for an underlying, their legs, and what it makes of each |

One of those numbered series states a contract's price against a span of days
with the date of each point — **five years of history on a quote subscription,
with no historical-data request behind it**.

Each reader was checked against the venue's own record, and against a live
session where the account is served that series. Where the record and the wire
disagreed, the wire settled it.

---

## Session, second factor, connection — 39 fixes

| Was | Now |
|---|---|
| The logon was judged on the last message type in the answer | Judged on what it answered with |
| An acknowledgement sharing its type with account traffic was missed | Found wherever in the answer it sits |
| A fill riding in with the reconnect acknowledgement was dropped | Kept |
| 4 ways a login ended that the venue had not ended | Closed |
| 5 ways a connection could be held open or given up wrongly | Closed |
| Second factor ran on first login only | Runs on reconnect too |
| The code provider ran on the socket thread | Runs off it |
| A code was spent on a login already lost | Not spent |
| The code was submitted while the socket was busy | Submitted only while quiet, as soon as it resolves |
| The session token reached the log | Kept out |
| An unstamped reconnect left the logon time behind, so the next reconnect read this session's own logon as a competitor and gave the account up | The logon time moves on |
| A takeover could not be told from this session's stale logon | Told apart |
| A signal interrupting a read counted as silence, so the venue's routing answer went unread and the connection failed | A signal is retried, not counted |

---

## Protocol — 17 fixes

| Was | Now |
|---|---|
| Fields searched for rather than read where the format puts them | Read at stated positions |
| Length and signature read from the wrong offsets | Read where the header puts them |
| A message borrowed the next one's bits | Ends where its own length says |
| A compressed header split across reads failed to frame | Frames |
| A read taking two messages kept one | Keeps both |
| A pause inside a frame read as a failed frame | Does not |
| 9 ways a frame could say something it did not carry | Closed |

---

## Account, contracts, options — 42 fixes

| Was | Now |
|---|---|
| The account could be read before it was stated whole | Reads wait for the download; every download ends |
| A holding the new statement never names stayed open | Closed |
| P&L did not survive a trading reconnect | Renewed |
| Contract hours stated on the wrong clock | Stated on the clock they are named with |
| Market-rule ids not surfaced | Surfaced on contract details |
| A lookup bounded on answer length | Bounded on silence |
| `con_id=0` reported as a contract | Reported as "not found" |
| An option's model read from the wrong contract | Read for the contract asked about |
| An option on a future priced as one on a share | Priced as a future option |
| A non-number price yielded a volatility; an overflowed tree yielded a price | Both refused |
| Time to expiry taken from the local clock | Taken from the venue |

---

## What the gateway does that this did not — 15

The sweep above asks where this client is *more* than the gateway. This one asks the reverse:
what does a gateway do that a caller here cannot reach? Every row states what the counterpart
does, and every one was checked against the wire before it was written.

| Was | Now |
|---|---|
| A change to an order reached the venue only for the terms the replace names; a new display size, discretionary amount, all-or-none or algorithm parameter was dropped, while this client's own answer already read back the new value | Every replace goes behind the caller's own statement, and the latest statement stands |
| A site that was down was reported as credentials being refused, and stopped the retry ladder | The venue numbers its refusals; the two that mean "not now" are waited out |
| A host that stopped answering ended the session | Its two standbys are named after it and tried first, as the venue names them |
| An order naming a model portfolio was refused outright | The model rides the placement, the replacement and the withdrawal |
| Day-till-cancelled went out as good-till-date, carrying an expiry nobody named | Good-till-cancelled with the flag that stands it down at the day's end |
| A combination's legs stated open or close on the field that carries a leg's number | On the field that carries the intent; a short leg states its borrow in full |
| An account summary asked for by currency answered nothing | The per-currency set is named by a prefix, not a figure name |
| Delayed and frozen quotes were collapsed to one entry, so the trade half never arrived | Both halves, as every live subscription already asked for |
| A withdrawal omitted three companions its subscription had stated | States what it withdraws |
| Six bar and tick series the venue serves were refused before they were sent | Sent under the venue's own names |
| A news article delivered as a document, or as a refusal, reached the caller as article text | Delivered as what it is |
| What the venue says to the account holder was counted as wire nobody reads | Reaches the caller |
| A symbol search was sent exactly as typed | Trimmed and collapsed, refused only where the venue refuses it |

Thirty-nine further differences were found and left alone, each recorded with what closing it
would take. Most are one shape: a family of order types, or a request whose answer has nowhere
to go — a new type, a new channel, or a refactor across files rather than a smaller diff.

---

## Matching the gateway rather than improving on it — 13

This client stands in for the gateway, so a caller who switches must get the same
answers. In each of these it was being *more careful* than the thing it replaces.

| Was | Now |
|---|---|
| Calendar requests expired locally after 30s; a later answer was discarded | The venue's answer ends them; the gateway sets no timer |
| An execution reported the reconciled difference as its own quantity | Reports the quantity the venue stated; the reconciliation still uses the difference |
| An unmodelled risk level or algo flag was refused, then silently dropped | Travels as the caller wrote it; the venue answers for its own vocabulary |
| Adjusted history was refused on a contract stated by description | Sent; the venue resolves it |
| A historical-news request naming a time window was refused | Carries the window; the element names and their format are the ones this client's other queries already send to the same service |
| A caller asking the time got the stamp on the last message the venue happened to send — frozen on a quiet link, refused before any arrived | The venue states how far its clock sits from this one; the answer is this machine's clock moved by that, and a connected session is always told the time |
| An order stating an advisor group, method or percentage had them dropped, under a note claiming the protocol carries no field | All three ride the order and its replacement |
| A profit request naming another account was warned about, then answered with this account's figures under the caller's own number | Refused, and refused before it takes the one subscription slot |
| A market rule this session has not seen was refused under this client's own number and code | Refused under the ones the venue uses, on both surfaces |
| A withdrawal arriving while its subscription was still opening was refused, and the subscription went on | Recorded and answered; the registration reads it before publishing and takes back down what it opened |
| A session held as many quote subscriptions as this client's own table, ignoring the number the venue states on the logon | Reads the allowance and refuses past it under the venue's own code |
| Margin figures the venue had not stated carried a stand-in number | Stay empty, as the reference client leaves them |
| One account summary at a time, ended at its own first answer | Two at once, each lasting until it is cancelled |
| An ordinary exception from a Python callback was logged and dispatch continued | Leaves dispatch and closes the session, as the reference client's loop does |
| Values the reference decoder reads as absent arrived as sentinel numbers | Arrive as  |
| Repeated order-status reports were suppressed when the status was unchanged | All delivered; the venue sends them and the reference client hands them on |

Kept deliberately, with reasons: the SRP group check, the server-proof requirement
and the DH range check, since no caller can see them and dropping them would only
widen what a hostile peer can do; and the permanent-id derivation, because the
gateway's own parse throws on this venue's identifiers and zeroes the field.

---

## Known gaps

- Six series are read and have never been seen: the venue acknowledges the subscription, assigns it a tag and states nothing, which is how it answers a series an account is not entitled to. Measured the same on a paper and a live login for one account. The readers are written and tested; they fill in when the data flows. The same is true of the spread scan.
- Readers for bonds, municipals, warrants and perpetuals are written from the record's own shape and have not been exercised against one of those contracts.
- Some paper phases skip rather than pass. Each names what refused it — venue or account, not the client: an order type an exchange will not take for that security type, a lot size rule, a subscription the account does not hold.
- One process holds one session. A gateway process rents its session out over a local socket; this has no socket. That is issue #2, unaddressed here.

---

## Taking it

- **All of it:** fast-forward.
- **Part of it:** cut at `7cfa6e56` ("tests move beside the code they test"), which moves twelve source files into directories. Commits before it are written against your current layout and cherry-pick with ordinary conflicts; commits after it assume the new paths.

```
git log --reverse --oneline upstream/main..main | head -743
```

- **None of it:** close it. If a particular subsystem would be more useful on its own, say which and I will open it against your layout.


deepentropy:
Closing: this will not be merged. See #412.

